PR #10, opened 2021-08-15
Switch to upwind fluxes in gkylzero

JunoRavin:
In G2, we have a Lua side option to choose between the default penalty fluxes or upwind (or recovery, though that is very experimental). I propose we actually deprecate penalty fluxes

The reason to deprecate penalty fluxes are as follows:
1. The upwind fluxes are far, far more stable than the penalty fluxes. There are simulations which do not run at any feasible resolution with penalty fluxes that run stably at reasonably coarse resolutions with upwind fluxes.
2. Eliminates the need for the maxs_by_cell data structure in CPU and GPU implementation, and thus eliminates a reduction operation from the time-stepping. The only reduction which needs to be performed now is the scalar reduction to obtain the stable time-step.

The documented down sides:
1. Upwind fluxes are more expensive by at least a factor of 2, perhaps slightly more in higher dimensions (this is obviously not a reasonable downside when there are lots of simulations which flat. out. do. not. work with penalty fluxes).
2. The upwind fluxes kernels are prohibitively large for p>1 in higher dimensionality.
3. We currently do not understand whether there is an optimal set of nodes to evaluate the upwind direction at for the Serendipity basis for p >= 2. For tensor bases (Serendipity p=1, or Tensor p>1) there is a natural choice of the complete Gauss-Legendre basis and this choice seems to work remarkably well (see above note about how much more stable upwind fluxes are), but the choice of nodes requires experimentation.

Because the upwind flux kernels are prohibitively large for p>1 in 5D and 6D and our current lack of understanding of node location choices for Serendipity p>=2, THIS BRANCH ELIMINATES THE P=2 SERENDIPITY KERNELS.

These kernels can be added back in (take a few hours to generate with Maxima).

This commit breaks the ctest_hyper_dg because the Vlasov-Maxwell kernels are different now with the upwind fluxes.

The new implementation is valgrind clean and works on GPUs! In fact the GPU performance is virtually unaffected by upwind fluxes (while the CPU implementation is a bit slower) such that the GPU performance increase looks better now (was a factor of 25 before for weibel_4d, now it's a factor of 50).


ammarhakim:
Can you fix the unit test? We can merge the branch into main then. Simply print the answer and copy/paste them into the unit test so it can compare with the new correct results.

As to the larger kernels: we need to move to an hybrid approach in the 5D/6D p=2 case, in which we do the interpolation to quad nodes with mat-vec multiply. This will eliminate huge amount of the code. The performance might not be that much worse. All we would need to do is store the interp matrix and then reuse it in dgemv BLAS kernels. We would need to think on how to do this on the GPUs, though. Perhaps it is not too hard and we can use cuBLAS.

One other option is the following: we generate interpolation kernels that interpolate to quad nodes on faces. Then we can call this from all the places we need: we need them in the LBO upwind kernels, the Vlasov upwind kernels and the GK upwind kernels. I think this will significantly reduce the overall size of the kernels as they will now be in a different function. Then, we can also experiment with the node layouts so they get automatically applied to all the various surface updates we wish to use them in.

JunoRavin:
A new kernel which interpolates to quad nodes on faces is a good idea. It's common infrastructure to Vlasov (including different flavors of Vlasov like Vlasov-Maxwell and Vlasov-Poisson), GK, and LBOs (GK and Vlasov). Right now the entirety of the slow down between upwind and penalty is within this interpolation, so the question is: can we do better than the massive unrolled interpolation (for certain polyOrders and dimensionalities). 

The answer is almost certainly yes. For lower dimensions/polynomial orders the unrolled loop is beneficial. Higher maybe not so much. 

It is interesting to me that the GPU seems to have paid less of a penalty for the upwind fluxes than the CPU side, with the performance of the GPU weibel 4d problem being minimally affected (and thus the relative performance difference between CPU and GPU looks even better!)

manauref:
"THIS BRANCH ELIMINATES THE P=2 SERENDIPITY KERNELS"

do you mean for 5D and 6D?
For lower dim, although we don't understand it, they conserved energy in every test I have tried (both in VM and GK).

JunoRavin:
It occurs to me how confusing this comment was. I meant only for 5D. I have kept the p=2 lower dimensional kernels. 

Since gkylzero and the upwind fluxes work out of the box on GPUs, there are some stress tests I would like to do for highly nonlinear Vlasov-Maxwell simulations with the node placements currently in the code. I've talked with @nmandell and I find it highly probable if we can diagnose why certain node configurations work (and when the node configurations fail) I think we have a paper we could write since I know of no other DG code that does upwinding this way and we preserve the computational advantages of our orthonormal, modal basis.

PR #20, opened 2021-12-01
vm-lbo merge

kltrbrds:
Self-species Vlasov LBO collisions.
- Needs some refactoring of the collisions RHS terms from the Vlasov terms for cleanliness.
- LBO surface kernels are currently significantly more expensive computationally than the Vlasov ones, this needs to be examined in depth.


apps/vm_species.c
```diff
@@ -100,21 +109,77 @@ double
 vm_species_rhs(gkyl_vlasov_app *app, struct vm_species *species,
```
ammarhakim, line 109: We should refactor this function to move the LBO terms to their own file. This way we will avoid these nested conditionals.

ammarhakim (approved):

PR #30, opened 2022-02-17
superlu_ops: LU decomposition method, and check if decomp has been done in _solve

manauref:


ammarhakim:
There is a memory leak which I can't seem to figure out. It is possible SLU is not being used properly. Valgrind complains at line 112 of superlu_ops.c. However, no allocation is occurring there. This needs to be looked into more carefully. It is possible we are not using SLU properly and do not understand what it is doing. I want the code to be valgrind clean. It is critical as otherwise we will be asking for serious trouble.

PR #31, opened 2022-03-16
GK Hack week branch merge

JunoRavin:
Merging branch from the GK hack week which adds new equation objects. Process of merging gkylzero into G2 has also begun, with small changes to things like the grid indices (starting at 1 instead of 0 now)


ammarhakim:
This branch has lots of valuable work that needs to be merged into the main-line. We will rebranch to fix GK issues. 

The Linux build seems to be failing, for reasons I do not understand. I am accepting the PR anyway as we can then fix once the merge is complete. It is possible there is something funcky about the way github actions are setup on the Linux container. We will find out.

PR #33, opened 2022-03-19
Geom_3d

liangwang0734:
# Consolidating `calc_geom_3d`

### Important facts
- The tangent vector definitions are not consistent with those used in `calc_geom_2d`.
  - Presently, in `calc_geom_2d`, for the bottom edge, `norm`, `tau2`, tau2` point to `+y`, `+x`, `-z`.
  - In this PR,  `calc_geom_3d` is designed so  that for the lower-y face has `norm`, `tau2`, `tau2` point to `+y`, `+z`, `+x`. This way, the three lower faces in x, y, and z can be handled in the same way but just with a shuffle.
  -  I'm not sure if this is consistent with `wave_prop` in dimensional splitting (needs to check the rotation direction, etc.). This could be easily controlled, though, by shuffling the order of vertices used to define these faces.
- The hexahedron volume calculation only handles "good" cells. It divides the hexahedron into five tetrahedrons.
- The cell face quad norm and area calculation uses the [varignon](https://math.stackexchange.com/a/2919201) projection. But other options are possible if we encounter performance/accuracy issues. Tests of present implementation is consistent with the area and norm computed by vtk internal algorithms for non-planar quads.

#### Division of the hexahedron
In the following image, the upper left panel is the original hexahedron, the rest are constituting tetrahedrons.
drawing

### Unit tests
I still don't have a nontrivial unit test for `calc_geom_3d`, but its components are tested in `ctest_wave_geom_helpers.c`. Also, a trivial test of a cube is included in `ctest_wave_geom.c`.

### DIFF
--- a/regression/rt_landau_damping.c
+++ b/regression/rt_landau_damping.c
@@ -17,7 +17,7 @@ struct langmuir_ctx {
 
 static inline double sq(double x) { return x*x; }
 
-inline double
+static inline double
 maxwellian(double n, double v, double u, double vth)
 {
   double v2 = (v - u)*(v - u);
--- a/regression/rt_neut_lbo_sod_shock.c
+++ b/regression/rt_neut_lbo_sod_shock.c
@@ -15,7 +15,7 @@ struct free_stream_ctx {
 
 static inline double sq(double x) { return x*x; }
 
-inline double
+static inline double
 maxwellian(double n, double v, double u, double vth)
 {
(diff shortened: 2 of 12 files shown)


liangwang0734:
I still don't have a nontrivial unit test for `calc_geom_3d`, but its components are tested in `ctest_geom_helpers`.

Also, the impact should be very localized and harmless.

Therefore, I changed this PR ready to review from draft.

PR #52, opened 2022-05-30
Diffusion

pcagas:
# Adding DG diffusion to `vlasov_fluid_species`

Diffusion is now working in up to 3D and `p=2`. Currently, only diagonal terms (D_xx, D_yy, and D_zz) are supported. Similarly, to the logic in `dg_advection`, N-dimensional diffusion coefficient D can be specified as a spatially variable function and projected on a continuous basis function set.

The equation is solved using 1-cell recovery of `q`, which is put into surface kernels (due to requirements on the stencil); volume kernels are empty.


pcagas:
Changing to draft to fix diffusion on GPUs

pcagas:
DG diffusion now works on GPU. All the tests are passing.

PR #58, opened 2022-07-21
Implement hybrid basis for kinetic models

manauref:
This implements p=1 hybrid basis in kinetic models (this goes with PR of hybridBasisWp2V -> main in gkylcas, and g0-merge-gkhybrid -> g0-merge in gkyl/g2). Hybrid bases are:
- Hybrid: p=1 in conf-space and p=2 in v-space.
- GkHybrid: p=2 in vpar, p=1 in other directions.
The purpose is to do away with the complexity of dealing with p=1 in EMGK and LBO.

The approach is the following:
- Generate/implement basic methods for the Hybrid and GkHybrid basis types, like basis eval, quad_to_modal and upwinding methods.
- The kinetic Apps (e.g. vlasov) force p=1 to use hybrid basis. In this case there's a basis mix, where configuration space uses Serendipity and phase space uses Hybrid (or GkHybrid). Some updaters (at least on the g2 side) need to be made aware of this compatibility.
- Most updaters remain unchanged, although some (like MaxwellianOnBasis in g2 and obviously kinetic equation objects) need new p=1 kernels. I only generated new kernels for VM, VP, GK, VmLBO and GkLBO.

I tested p=1 collisionless VM and GK and got similar results as the p=2 tests (and similar as before all these changes). I also tested collisions-only VmLBO and GkLBO and confirmed that they conserve M0, M1, M2 with p=1 and p=2.

There are several IMPORTANT caveats that will be resolved soon (but are independent of this PR, I simply discovered them while working here):
1. I did not modify the CFL frequency that kernels return, which is not suitable now that v-space/vpar is p=2 even though the input file will say p=1. To compensate for this I think the user has to request a cflFrac of 0.6, but this is conservative since it is not p=2 in every dimension. In the future we should remove the (2p+1) factor in the CFL calculation outside of the kernels. That is left for a separate project/branch.
2. I made some changes to cross-collisions in order to get the code to compile, but I believe it is not actually working. Will work with @kltrbrds on that separately.
3. Upon further testing I found that GkLBO does not conserve M2 when B=B(x) is used. This is already a problem in gk-0, so it was not introduced by the changes in this branch.
4. There's other work unrelated to hybrid basis in this branch (e.g. conf*phase bin_op multiplication) that should make its way into gk-0/main.

p.s. tagging several developers so they are aware of/approve the new basis loading approach.


manauref:
Point 3 above (errors in GkLBO) were due to g2-side things (calculation of 1/B and 1/B^2), and has been fixed in g2's https://github.com/ammarhakim/gkyl/commit/3bd050c999e2086b81eee8ce3e23246c49614ef4 commit.

@kltrbrds made changes to gkylcas that should fix cross collisions, but they need to be incorporated into this branch.

manauref:
commit https://github.com/ammarhakim/gkylzero/commit/7e1e10c2d8b6e762e79d8988273f2fe742c02663 makes cross collisions consistent with gk-0 and main branches, and they all produce the same results now. This goes with https://github.com/ammarhakim/gkylcas/pull/6#:~:text=%E2%80%A6-,3b0fbb2,-Merge%20state in gkylcas.

 (I was wrong before and cross collisions were not broken at the time of submitting this PR, they just followed a slightly different approach and used older maxima).

PR #59, opened 2022-07-22
Boundary fluxes

kltrbrds:
- Updater that runs the dg surface updater inside the ghost cells.
- Extended the moment boundary corrections updater used for LBO to be able to do moment calculations in the boundaries of configuration space.
- Added a source term which takes species boundary flux and uses it to scale the specified source profile.


manauref:
@kltrbrds it'd be good if you can show us how this is going to plug into g2

ammarhakim:
I do not think we need to fully understand how this plugs into G2 before we do the merge. However, someone who knows the Poisson stuff should look into the conflicts before we accept the PR.

manauref:
I don't mean we should hold up the PR because of the g2-merge, but at our
next meeting it'd be good to get a vision of how this is going to plug into
g2. I'd hope that is true of all work in g0 moving forward.

kltrbrds:
The Poisson conflicts are due to merging in work from the gk-0 branch to ensure consistency since Noah modified some of the LBO code (such as mom_calc_bcorr, which this branch also modifies). I'm not sure what the state of the other changes made by that merge are, or if its a good idea to have them in main.

PR #61, opened 2022-07-24
bc_basic updater to apply BCs in kinetic species (Vlasov/GK)

manauref:
Previous the BCs, primarily absorb and wall (now reflect), were in a series of functions in each equation object. That led to lots of duplicated code, and g0-g2 complications because an equation object is needed, when I think BC application is a solely independent operation.

This PR introduces a bc_basic updater to apply absorb and reflect BCs. It works in Vlasov, according to visual examination of sheath and neut_lbo_wall reg tests (on both CPU and GPU). We could also make it apply copy BCs easily, but I didn't go that far.

One thing I'd appreciate a comment on:
- Could this updater also be used to apply BCs in fields and/or moment models? if not, maybe I should call it bc_kinspecies_basic, or something like it.


JunoRavin:
You can remove the release method for bcs from dg_gyrokinetic as well, but other than that I think this is fantastic y'all. Your most recent commit suggests there's still up with the bc_reflect unit test? The github actions pass but that might because the bc unit test isn't hooked in there yet. 

_Could this updater also be used to apply BCs in fields and/or moment models? if not, maybe I should call it bc_kinspecies_basic, or something like it._

To answer this question, we could add BCs for things like Maxwell's, but I do not think we should. Reflecting wall BCs for Maxwell's (perfect electric conductor) is its own special function, same thing for Euler, isothermal Euler, 10 moment. All four of these equation systems have very specific definitions of a "wall." (that was actually why I put the BCs in the equation object originally because wall BCs is a very specific thing for Maxwell's). We will need some way to hook in Maxwell's wall BCs into G2 but I do not think we want to do it through this updater.

Since things like Maxwell's require specific Maxwell's things, we might want the updater we wrap (something like dg_updater_maxwell) to include not just an advance method but also an advance BC method. This structure will also allow us to add other specific Maxwell's equations boundary conditions (symmetry BCs, perfectly matched layers) without complicating the logic in this updater with a bunch of Maxwell's only conditionals.

ammarhakim:
May I ask why ‘wall’ was renamed to reflect? Reflect is not really accurate. Walls may do more than reflect, and so this generic name was chosen.

JunoRavin:
@ammarhakim Mana and I discussed this name change. Part of the confusion is that multiple different equation objects all had a "wall" BC so the argument was that calling the BC type "wall" across different equation objects was confusing. Plus, as you say wall BCs mean multiple things and the "absorb" bc is its own type of wall. 

While reflect is also not a perfect name, at least for the specific types of equation objects this bc_basic_updater will be used for, it is more descriptive (since it's the BC that reflects the distribution function).

ammarhakim:
The same name wall was by design: you did not need to look up or remember what each objects perfect wall Bc was called. It’s ok to change names but changes like this impact people, and things will get more complex as we go along (with name changes etc). It’s best to allow some discussion amongst core dev people so we are aligned.

I am not worried about this specific instance. Just a caution as we move forward, specially with CSSI expanded team at RPI doing major work, plus UNH using G0 in Open GGCM.

manauref:
@JunoRavin, when you say BCs for Maxwell's, Euler and 10 moment are their
own special functions, what kind of functions are they? are they simple
(basic) functions of the ghost cell in terms of their immediate neighbor
skin cell? Because if they are then you could use bc_basic, but would
simply have to define such a function. But if you think it makes more sense
to use a different updater or the equation functions, I'm fine with that
too (and I hope you know how you are going to wrap those BCs in g2 because
I don't).

JunoRavin:
My current thought process: the utility of bc_basic is that there are shared BC functions across vlasov (all flavors) and gyrokinetic (all flavors). Now we don't need to define a bc_absorb and bc_reflect for all these different equation objects.

It is true that the "wall" BC for Maxwell's, Euler, Isothermal Euler, and 10 moment are all similar in structure to our typical BCs (take a function in, evaluate the input function in the skin cell, copy to the ghost cell) but these functions are specific to those equation objects. Maxwell's wall looks very different from Euler which looks different from 10 moment which all look different from Vlasov/Gyrokinetic (because Maxwell's, Euler, etc. don't have to do the flip_copy). 

One can imagine having an enum which has further inputs like BC_MAXWELL_PEC or BC_EULER_NO_SLIP, but I'm not sure that's the right approach because you're going to need both functions specific for finite volume vs. DG and you're going to need enums which distinguish between all the different fluid equations. 

If a BC is specific to that equation object and that equation object *only* then the right approach in my mind is to store a pointer to that BC function in the dg_updater (so we have a dg_updater_maxwell that contains both an advance and a BC advance method) and not clutter bc_basic with a dozen different enums for BC types that are specific to individual equation objects.

ammarhakim:
Before merging this branch, please change the line:

`
enum gkyl_bc_basic_type { BC_ABSORB, BC_REFLECT };
`
to:
`
enum gkyl_bc_basic_type { GKYL_BC_ABSORB, GKYL_BC_REFLECT };
`

Please remember to prepend enums with GKYL prefix so there is no danger of messing up with someone else's defined values.

Also, do not worry about the moments app at present. Moments are very different and each eqn object needs a different function, specific to that eqn object. Further, geometry needs to be handled too, differently than is in GK/VM. So no need to worry about that at present.

PR #77, opened 2022-08-24
MHD source

liangwang0734:
- Hooked a generic MHD source updater into the moment app
- Dedner's GLM source, simple version
  - ch is set to max speed from wave solvers
  - first way of using the source: specify glm_alpha, see Mignone & Tzeferacos, JCP (2010) 229, 2117, Equation (27)
  - second way of using the source: specify a constant glm_cp
- Powell's eight-wave scheme
- Orsag-Tang test updated to newer Athena version
  - results do not look great with either eight-wave or glm
 
TODOs
- more consistent calculation of divB (for eight-wave and glm) and B.grad(psi) (for glm)
  - right now, simple central difference is used for both divB and grad(psi), assuming uniform Cartesian grid
  - for nonuniform Cartesian grid
    - to compute divB one may use face-center B_normal due to linear interpolation, and compute integral of div(B) use Green's theorem (sum Area * Bn/Volume)
    - to compute B.grad(psi), one may change it into div(B*psi)-(divB)*psi, and compute div(B*psi) and divB using the surface integrals
    - for GLM/8-wave, since B_normal is indeed changed by the Riemann solver, it might be better to use some proper B_normal value after limiting; but this is not straightforward to compute in the wave-propagation method (in the flux difference method, one may use the value in the obtained numerical flux due to limited reconstruction)

### DIFF
--- a/apps/moment.c
+++ b/apps/moment.c
@@ -25,7 +25,6 @@
 #include <gkyl_wave_geom.h>
 #include <gkyl_wave_prop.h>
 #include <gkyl_wv_apply_bc.h>
-#include <gkyl_wv_euler.h>
 #include <gkyl_wv_maxwell.h>
 #include <gkyl_wv_ten_moment.h>
 #include <gkyl_wv_mhd.h>
@@ -1306,6 +1305,9 @@ gkyl_moment_app_stat_write(const gkyl_moment_app* app)
   time_t t = time(NULL);
   struct tm curr_tm = *localtime(&t);
 
+  // compute total number of cells updated in simulation
+  long tot_cells_up = app->local.volume*app->num_species*app->ndim*app->stat.nup;
+
   // append to existing file so we have a history of different runs
   FILE *fp = 0;
   with_file (fp, fileNm, "a") {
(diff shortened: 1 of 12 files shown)


ammarhakim:
Our design goal is never to exposed internals of equation objects in public-facing APIs. This allows a clean use of the various equation objects via well-defined APIs. Hence, please do not move the struct to the header. 

Further, please remember that all public-facing structs and functions MUST be prefixed with the gkyl_ prefix. Private with only local scope need not be prefixed like that. 

Once you remove the struct we can review the code and then accept the PR if it looks good.

ammarhakim:
This code is not ready for merge yet. First, you should not store pointers to eqn objects in this manner, but need to acquire() a pointer. Our eqn objects are reference counted and so storing pointers by simple assignments is very dangerous and can lead to hard-to-find memory leaks.

Second, we can't restrict the code to uniform Cartesian. The wave_prop solver works on any mapped grid and so putting a limitation on the GLM source to only work on uniform Cart grid is not correct. Basically, immediately someone will try a flow in another geometry (like me! I want to do theta-pinch in R-theta geometry, or the spinning mirror). So we need to reconsider carefully how to compute the needed terms properly, as done for rest of the FV solver.

What I suggest is to close this PR without merging at present. Let us talk on Tuesday after my geometry tutorial and we will disucss how to proceed.

ammarhakim:
Closing this PR without merging. Overall, this is good work and in the correct direction. However, we need to discuss some important (and not trivial to fix like geometry) issues before we can merge this.

liangwang0734:
> This code is not ready for merge yet. First, you should not store pointers to eqn objects in this manner, but need to acquire() a pointer. Our eqn objects are reference counted and so storing pointers by simple assignments is very dangerous and can lead to hard-to-find memory leaks.
This will be fixed.
> 
> Second, we can't restrict the code to uniform Cartesian. The wave_prop solver works on any mapped grid and so putting a limitation on the GLM source to only work on uniform Cart grid is not correct. Basically, immediately someone will try a flow in another geometry (like me! I want to do theta-pinch in R-theta geometry, or the spinning mirror). So we need to reconsider carefully how to compute the needed terms properly, as done for rest of the FV solver.
Nonuniform Cartesian is easy. But general geometry needs more thinking.
Nevertheless, overall the results do not look great in Orsag-Tang. It's not clear what I am missing.
> 
> What I suggest is to close this PR without merging at present. Let us talk on Tuesday after my geometry tutorial and we will disucss how to proceed.

ammarhakim:
Let us talk in-person on Tuesday to see where we are and how to resolve. Maybe two of us looking at this together will help.

PR #81, opened 2022-09-10
wv_euler: hllc

liangwang0734:
Tested with ctest_wv_euler and sodshock with manual hook into wave_prop (but this PR does not include the hook).

In a further PR, we may allow a user to specify the domain-invariant scheme (e.g., Lax, HLLC/HLLD). Note that the number of waves must be changed accordingly (automatically). But in this PR, I'd prefer not to make changes with broader impacts yet.

### DIFF
--- a/zero/wv_euler.c
+++ b/zero/wv_euler.c
@@ -195,6 +195,89 @@ qfluct_lax(const struct gkyl_wv_eqn *eqn,
   }
 }
 
+
+// HLLC
+static double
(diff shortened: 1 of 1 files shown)


ammarhakim:
Very good work, Liang! This resolves a long-standing issue  that we have only been able to use Roe with wave. Now other, perhaps more robust RPs, can be also used. Can you please resubmit this PR to merge into the lfix-4 branch instead? I will hook into the full system and then it will become available when we merge lfix-4 into main

PR #86, opened 2022-09-12
VM fluid source

kltrbrds:
- Added a source for the VM fluid species.
- Source relies on a constant user-defined profile function, no current implementation of any dependence on boundary fluxes.

### DIFF
--- a/apps/gkyl_vlasov.h
+++ b/apps/gkyl_vlasov.h
@@ -48,6 +48,15 @@ struct gkyl_vlasov_source {
   void (*profile)(double t, const double *xn, double *aout, void *ctx);
 };
 
+// Parameters for fluid species source
+struct gkyl_vlasov_fluid_source {
+  enum gkyl_source_id source_id; // type of source
+
+  void *ctx; // context for source function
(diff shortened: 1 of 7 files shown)


JunoRavin:
@kltrbrds Why within the app do these two blocks of code look different?

`  // compute source term
  // done here as the RHS update for all species should be complete before
  // bflux calculation of the source species
    if (app->species[i].source_id) {
      vm_species_source_rhs(app, &app->species[i], &app->species[i].src, fin, fout);
    }
  }
  for (int i=0; i<app->num_fluid_species; ++i) {
    if (app->fluid_species[i].source_id) {
      vm_fluid_species_source_rhs(app, &app->fluid_species[i], &app->fluid_species[i].src, fluidin[i], fluidout[i]);
    }
  }`

is there some reason to expect that you have the same source for every kinetic species, but not every fluid species?

kltrbrds:
The kinetic source gives the entire species table as it can depend on the boundary flux of a different species. The fluid source is only specified by a constant function at the moment, so there's no reason it would need to access any other species.

PR #91, opened 2022-09-29
separate mkdeps files for stellar amd and intel

akashukla:
created separate mkdeps files for stellar amd and intel


manauref:
@akashukla can you make some changes please:
- the new machine files should be modelled after the existing stellar machine file in /machines, and should go in /machines, not in /install-deps
- it'd be better to ```git mv``` the stellar mkdeps file that's there to rename it, that way github retains the history of that file.

ammarhakim:
Also please don’t accept your own PR. This problem would be caught if someone looked at it.

PR #93, opened 2022-09-30
LBO checks

manauref:
Under certain circumstances the LBO cross primitive moments, u and vtSq, can acquire unreasonable results, like

u > v_max
vtSq < 0
vtSq > v_max^2

Add if-statements in the LBO equation objects so that in these cases we turn the LBO off. We had this in g2 and it may improve stability. 

In g2 we did two more things:
- We kept track of how often the LBO gets turned off.
- We also turned the LBO off if M2<0.
at some point we should implement these in g0 as well, but for now I want to check if these are sufficient to get rid of some numerical instabilities.

### DIFF
--- a/zero/dg_lbo_gyrokinetic_diff.c
+++ b/zero/dg_lbo_gyrokinetic_diff.c
@@ -41,11 +41,11 @@ gkyl_lbo_gyrokinetic_diff_set_auxfields(const struct gkyl_dg_eqn *eqn, struct gk
 
 struct gkyl_dg_eqn*
 gkyl_dg_lbo_gyrokinetic_diff_new(const struct gkyl_basis* cbasis, const struct gkyl_basis* pbasis, 
-  const struct gkyl_range* conf_range, double mass, bool use_gpu)
+  const struct gkyl_range* conf_range, const struct gkyl_rect_grid *pgrid, double mass, bool use_gpu)
 {
 #ifdef GKYL_HAVE_CUDA
   if (use_gpu)
-    return gkyl_dg_lbo_gyrokinetic_diff_cu_dev_new(cbasis, pbasis, conf_range, mass);
+    return gkyl_dg_lbo_gyrokinetic_diff_cu_dev_new(cbasis, pbasis, conf_range, pgrid, mass);
 #endif
   struct dg_lbo_gyrokinetic_diff* lbo_gyrokinetic_diff = gkyl_malloc(sizeof(struct dg_lbo_gyrokinetic_diff));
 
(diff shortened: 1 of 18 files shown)


JunoRavin:
I do not want to support this robustness check. I don't think it does anything for Vlasov and I think it clutters the code. If either of these three conditions are met, the simulation should abort, not take time steps. If you want to add this robustness check to GK, that's fine, but I am rejecting this change to Vlasov.

manauref:
@JunoRavin 

changing
```
  if (dir >= lbo_vlasov_diff->cdim) {
```
to
```
  bool noPrimMomCross = checkPrimMomCross(lbo_vlasov_diff, nuSum_p, nuUSum_p, nuVtSqSum_p);
  if ((dir >= lbo_vlasov_diff->cdim) && (noPrimMomCross)) {
```
hardly qualifies as cluttering.

Negative thermal speeds can happen, and you can still get useful results in those cases because vtSq<0 may only happen in a few cells, so it's not necessary to terminate the whole simulation. That's like saying we should throw away every Vlasov simulation just because it has a negative f somewhere (which many do).

I'll reopen and we can wait for @ammarhakim to weigh in

JunoRavin:
No, negative distribution function is not the same as negative VtSq. Negative VtSq implies imaginary temperatures and the distribution function is not actually representable. Negative distribution function is fine so long as certain moments stay positive definite because we can always represent the distribution function with a positive definite basis in which the solution is positive everywhere.

That is definitely cluttering as it's a check which is called in every phase space cell, every forward Euler call, ever time step for a quantity that only varies in configuration space and should never, in any production Vlasov calculation, return false. 

If we want to check this and then abort out of the simulation loop, I am fine with that, but I don't want this inside the update method for Vlasov.

manauref:
> 



> Negative distribution function is fine so long as certain moments stay positive definite because we can always represent the distribution function with a positive definite basis in which the solution is positive everywhere.

The same is true of vtSq, but I think that the if the cell average f is <0, there's no representation that will give you f>0 everywhere (correct me if I'm wrong). So, checking vtSq[0]>0 is just as sensible as checking f[0]>0. But by your argument we should throw away a simulation with f[0] somewhere, which is too stringent.

JunoRavin:
Again, that is wrong because we can safely take time steps when f[0] < 0 because nothing in the solver depends on f being positive definitely, but we cannot necessarily take time steps if nuVthSq[0] < 0 because then the diffusion is the wrong sign and then there is no stable time step. 

**A proposed compromise:** instead of using dg_mul_op_range to get nu*vthSqSum and nu*uSum, we write an updater which loops over configuration space and checks the three conditions you've stated. If any are true, nu, nu*vthSq and nu*u are set to 0 in that cell, otherwise the updater works like weak multiplication. 

This moves the check to where it should be in configuration space and avoids further conditional checking in the inner loop of the LBO updater. It also makes it easier to count how often this happens since how often this happens is just a function of configuration space. And any simulation we run should not have this happen very often, so the extra cost of calling the LBO with nu, nu*VthSq, and nu*u all equal to 0 won't matter.

manauref:
I'd say we leave that as a potential improvement for later (which I believe
may actually be (negligibly) slower than what we are doing now because
it'll involve launching more kernels and looping over the conf-grid again).
At the moment the priority should be to make sure we can run simulations in
g0-g2 that we used to be able to run in g2 with both vlasov and
gyrokinetics (which is why I implemented this), and not to prioritize
(negligible) improvements to performance.

PR #95, opened 2022-10-03
Spitzer and pow(sqrt()) updaters

manauref:
Implement updaters that compute the Spitzer collision frequency and pow(sqrt(f),q) using Gaussian quadrature.

Spitzer_coll_freq currently only has the normNu option, since I think that's the most useful right now.

Unit tests pass on the CPU and GPU. However valgrind is giving errors on the ```qrange``` function in these updaters, but that is the exact same function as in proj_maxwellian_on_basis, so I don't know if it's a true error. Maybe reviewers can check in another (non-stellar) computer.

### DIFF
--- a/unit/README.md
+++ b/unit/README.md
@@ -7,4 +7,4 @@ To create/run a new unit test:
 
 1. Write and save your unit test file in this folder (with a `ctest_` prefix).
 2. In the main directory re-compile the code.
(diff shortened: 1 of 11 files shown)


JunoRavin:
I get zero errors when I run the unit test with the valgrind on my iMac (3.19), so I think this pull request is actually fine and I will merge now.

Issue #99, opened 2022-10-09
gkylzero.h is not generated by Makefile initially, and some compilations fail


manauref:
```make install```
generates gkylzero.h, but that is if it gets to the ```install``` part. I believe in a clean build with CUDA it will fail because several unit/*_cu.cu tests include gkylzero.h, but gkylzero.h has not been generated yet.

The makefile should have a step prior to install where it calls the ```minus/gengkylzeroh.sh``` script.

PR #100, opened 2022-10-10
Mods to Makefile to the gkylzero header is generated during build

ammarhakim:
This fixes Issue #99


ammarhakim:
This has a bug in it :(

PR #101, opened 2022-10-10
Automatically generating gkylzero header at build time

ammarhakim:
This fixed Issue #99


manauref:
Clean build on Stellar-amd seems to be fixed now.

PR #108, opened 2022-10-29
Change Makefile so /regression and /unit executables are not built by

manauref:
default. That's because currently those executables take way too much space and if you have other stuff in your $HOME, your diskspace can easily fill up and crash the build. Now one has to build these on demand with, for example,

make cuda-build/regression
make cuda-build/unit

or specific tests with

make cuda-build/regression/rt_weibel_4d
make cuda-build/unit/ctest_dg_bin_ops

Additionally, change ```-j``` in scripts that build dependencies to
```-j 32``` because Stellar and Perlmutter admins don't like that we hog
all the cores on the login node.

Adapt config and make files to allow passing location of cuda math
libraries, because on Perlmutter they are in some other nonstandard
location. Update Perlmutter machine files with this (also had to unload
darshan module for some reason, recommended by admins).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,19 +43,22 @@ endif
 KERN_INC_DIRS = $(shell find $(SRC_DIRS) -type d)
 KERN_INCLUDES = $(addprefix -I,$(KERN_INC_DIRS))
 
+# Location of CUDA math libraries
+CUDAMATH_LIBDIR = 
+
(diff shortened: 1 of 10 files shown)


ammarhakim:
This is probably ok, but it seems that target names should be just make unit and make regression. It’s ok to only build the lib by default. Make this change and we can merge the PR

PR #110, opened 2022-11-01
Fixing a problem with Maxwell eqn qfluct. This seems to be a serious issue!

ammarhakim:
This is an issue with many complicated problems in which the EM field has all terms active. I have to revert back to the pre-MP-XX qfluct till I figure out why the Maxima generated code is not working.

Maxwell MP-XX will NOT WORK at present. Need to figure this out slowly and properly.


ammarhakim:
I believe I know what the issue is. The problem is that rescaled speeds need to be used in qfluct. At present it is impossible to do this. Hence things are failing. A bit scary and need to check all qflucts to ensure we do not have this issue everywhere.

PR #121, opened 2022-11-16
Fix energy conservation in LBO for p>1

manauref:
Energy conservation in the LBO was not exact for p>1 because in creating g0 kernels a simplification was made which was only valid for p=1 (and possibly tensor basis). Essentially for the diffusion operator our Maxima had 

```
temp_diff_c : calcInnerProdList(varsP, 1, bP, diff(h_e, surfVar, 2)),
writeCExprs1(temp_diff, temp_diff_c),

diff_incr_c : calcInnerProdList(varsP, nuVtSqSum_e, bP, doExpand1(temp_diff,bP)),
writeCExprs1(diff_incr, diff_incr_c),
```

i.e. we first projected d^2(f)/dv^2, and later multiplied it by nu*v_t^2 in a second projection. That may only be valid if the phase basis is a tensor product of a position basis and a velocity basis. Instead we should've had (and now have), at least for p>1,

```
diff_incr_c : calcInnerProdList(varsP, nuVtSqSum_e, bP, diff(h_e, surfVar, 2)),
writeCExprs1(diff_incr, diff_incr_c),
```

This fixes M2 conservation. Note that previously the conservation might've looked almost correct, possibly fooling some users, with plots showing conservation up to 1e-10 or 1e-11 and increasing nonlinearly in time. But that was wrong in two ways:

1. In g2 we showed conservation should be good up to 1e-13 or 1e-14 (at least for 1x1v and 1x2v).
2. The errors should only accrue linearly, proportional to the number of time steps taken.

New LBO kernels are a little more expensive, but are correct. I think there may be ~20% improvements to be made by moving the x-dependent diffusion coefficient out of the velocity integral. I've experimented with this but may implement it fully in a separate PR.


JunoRavin:
@manauref The Gkeyll team needs some discretionary funds to buy devs alcohol/sweets of their choice when they track down bugs like this.

PR #122, opened 2022-11-18
Fluid collision, etc.

liangwang0734:
- time-centered updater for simple collisions as inter-species friction
  - the collision is plugged as an additional, separate step into the existing em_coupling method (to avoid more API codes and maybe it is simply better this way)
  - at least the momentum exchange can be merged into the em coupling; but that needs more careful analytic derivation and is done for now
  - presently, one needs to enable it in the main app with the `has_collision`  option and specify a valid table of `normalized collision frequencies `nu_base` ;
  - see [this example](https://gist.github.com/liangwang0734/9c410cc7e61b6003cf7ffb9c4b1f1cb1) and look for ` nu_base`
```   
    .has_collision = ctx.has_collision,
    // scaling factors for collision frequencies; nu_sr = nu_base_sr * rho_r
    // the matrix must be symmetrix
    .nu_base =
      {
        {0, ctx.nu_base_ei}, // ee, ei
        {ctx.nu_base_ei, 0}  // ie, ii
      },
```

- `user_source` option for each species; for now, it simply applies additional sources to number density and the scaled temperature `kB*T/(gas_gamma-1)` 
```  - the temperature is rescaled so that the update does not depend on `kB` and `gas_gamma`; 
  - the source is enabled for a species when it contains a valid `user_source_func`
  - see [this example](https://gist.github.com/liangwang0734/9c410cc7e61b6003cf7ffb9c4b1f1cb1) and look for ` user_source_func`
- other changes
  - Makefile, identify *.c files correctly by quoting `*.c`
  - moments app: do not update field if its evolve is false
  - `gkyl_mat_inc` method
  - in `moment_apply_bc` pass the skin cell's computational coordinates too; this needs to be changed to physical coordinates;
    - the main purpose is to allow user-defined boundary conditions


manauref:
@liangwang0734 can you please add a unit test for gkyl_mat_inc in unit/ctest_mat.c?

PR #127, opened 2022-12-12
DG Fluids + Parallel-kinetic-perpendicular-moment model

JunoRavin:
This branch has seen enormous development progress on a variety of fronts and can now be merged (modulo corresponding changes in G2 to facilitate continued functionality of G0 main with G2 g0_merge branch)

1. We have our first implementation of DG fluids using Lax fluxes at quadrature points before projection back onto our modal basis. 
2. In many places (such as moments, prim_lbo, prim_lbo_calc, and mom_calc) there has been a larger push to standardization of taking a use_gpu boolean in the new method, and then selecting the appropriate host or device method
3. As a result of all of the increased functionality, various methods that are wrapped by G2 have had to be extended such as dg_updater_vlasov. These changes mostly manifest as an increased number of auxiliary fields but another change is that we have split field_id into both a field_id and a model_id to facilitate easier specification of things like special relativistic or PKPM Vlasov (which still use Maxwell's equations unmodified) from genuine changes to the field equations (such as Vlasov-Poisson)
4. Many objects in this branch remove indirections that have previously proved meddlesome on GPUs. While the approach has not been totally standardized where possible (especially in volume kernels) I have removed one layer of indirection. The G2 side of things does not care about this *but if you're working in a branch of G0 be careful that merging main into your branch does not change this*. It's a tedious process and if the kernel selection gets messed up it will be painful to fix. 
5. Other small feature extensions in the G0 app as part of testing, such as applied currents and external electromagnetic fields. 

This is a big enough merge that anyone that wants to look this over can do so. We need to merge this branch before the holidays because the gk-nov-7-22 branch has had to make changes to the Vlasov equation object that need to be combined with this branch and it's better to start the review process now.


manauref:
Smaller merges/PRs are preferable, but OK.

I have a basic question before starting to review, does this pass the g2 regression tests? I think the majority of the GK and Vlasov reg tests run.

JunoRavin:
I am working on the corresponding G2 pull request so that the Vlasov regression tests run again. There have been a number of changes to the function signatures of things like dg_updater_vlasov, and there are few places throughout G2 that now need to take a boolean for use_gpu in their new method to simplify the initialization

manauref:
Sounds good. I'll look out for that.
One possible way to set this boolean is as done
in gkyl/Updater/GkSheathBc.lua.

pcagas:
@manauref, do you mean this?
`local useGPU = xsys.pickBool(tbl.useDevice, GKYL_USE_GPU or 0)`

manauref:
Yeah

JunoRavin:
Closing this pull request for now because of some recent breakthroughs in PKPM that require a slight refactoring. Commit coming imminently but now we solve for both the 0th and 1st Laguerre moments so that no equation is required for p_perp, and the dynamics of p_perp are kinetic (retain q_perp and the associated higher v_perp^2 v_parallel^k for all k). Preliminary simulations very promising though implementation is more complicated and there have been significant headaches in tracking down sign errors.

PR #129, opened 2023-02-02
Gk nov 7 2022

JunoRavin:
Hackathon branch from when @tnbernard visited PPPL. Includes a first pass at ionization, some build system changes, and a large refactoring of the Vlasov equation object to accommodate general geometry (which now works with neutrals!)

Corresponding G2 merge also needs to be completed. https://github.com/ammarhakim/gkyl/pull/113 will work though it won't have the option for general geometry. But that PR does have the necessary changes to the Vlasov updater to accommodate this G0 branch merge.


JunoRavin:
Strange Mac issues need to be debugged before merging

PR #136, opened 2023-03-16
These gauss quad utilities seem to require explicit casts for the build to work on Mac

JunoRavin:
G0 works on its own, but when I build G2 and it runs a test file when it looks for G0, if the explicit cast is not here it throws an error about assigning double * to void * (like when we have to do explicit casts on GPU).


ammarhakim:
I am closing this PR. We should not need these casts at all. I can help fix problems in G2 if needed.

PR #139, opened 2023-03-28
Prepend the current directory to EXEC_LIBS

liangwang0734:
Previously the relative path is linked and running the test from another place might fail because it can't find the lib.


ammarhakim:
So I had thought about this also and even made the change. For some reason I do not remember I had to undo the change. I think the problem is that on the Mac this will not work. I do not have a Mac to check things on. Maybe @JunoRavin  can pull your branch and check on the Mac. Shared linking is tricky on Mac as they use a different system.

liangwang0734:
The more robust solution can be found at https://stackoverflow.com/questions/18136918/how-to-get-current-relative-directory-of-your-makefile

PR #160, opened 2023-07-06
Geometry and various other things

ammarhakim:
At some point we need to get rid of Macs completely.


manauref:
One benefit of supporting Macs and PCs is that students, who make valuable contributions, are able to work on their own laptop before they are able to get a Linux-based computer from an institution or an account on a cluster (and editing code on a cluster is sometimes not possible, e.g. while flying).

ammarhakim:
One can get a Linux laptop. Mac is not the only laptop solution in town.

PR #164, opened 2023-07-08
Commenting out zeroing of range and rect_grid structs.

manauref:
It caused seg faults in g2. Please check g2 before making changes to core components in the future.


ammarhakim (commented):
These statements are perfectly legal C. There is no reason why it should cause any problems in G2. It does not cause any problems in G0 or G0 via LuaJIT. It is ok to comment these out, however, for now.

PR #165, opened 2023-07-08
Adding common header for Poisson eqn BCs so they are shared between

ammarhakim:
the various Poisson solvers. No enum or structs or functions should be redefined!


ammarhakim:
Just for record: the common header I introduced is probably not the best named one. For now this is fine and we will have to revisit this when we emerge from this hackathon. Clearly, we are all making very serious messes all over  the place!

PR #190, opened 2023-08-27
Improved robustness and functionality for the PKPM model

JunoRavin:
This branch was originally created back in April as part of ongoing testing to understand the behavior of the b_hat magnetic field unit vector within the PKPM model, and it spawned a multi-month careful investigation of the robustness of the PKPM model, which revealed a number of profound insights that have greatly improved the capabilities of PKPM:
1. The first version of PKPM used the recovery DG method in many places to create continuous expansions of certain quantities. This has been completely deprecated and there is now *zero* recovery anywhere in the PKPM model. Fundamentally, recovery just does not work very well for hyperbolic PDEs and generates sharp oscillations in quantities such as b_hat.
2. Further, by switching to an averaging procedure instead of recovery to handle interface fluxes (such as the advection by the flow velocity u) we can prove phase space incompressibility in the weak sense, which also increases the robustness of the solver. While we do not have point-wise phase space incompressibility like we do in Vlasov, this property is nevertheless important to enforce as much as possible. Importantly to properly insure phase space incompressibility weakly, we must use a split-form of the fluxes for the averaging piece (so the flux takes the form of {{u}}{{F}} where {{.}} is the averaging operator) just like in recent work in entropy-stable methods. We use the same flux for the parallel streaming as well {{b}}{{F}} with a penalization that allows us to reduce to an upwinded streaming term in the limit that b_hat is continuous.
3. There is a necessary consistency condition in the solution of the pressure force which appears in PKPM. Basically we must directly solve for div(p_parallel b_hat) from the streaming term so that two terms exactly cancel and eliminate the development of spurious flows (since the first moment of the kinetic equation must be identically 0 if it is zero at t=0). 

A number of other improvements have been made as part of this larger feature extension including a significant cleanup of the G0 app (in preparation for just using this for Vlasov simulations wrapped through Lua), a streamlined implementation of dg_diffusion for the constant diffusion case, more diagnostics for PKPM (integrated and reorganized arrays for the conserved quantities for use in 10 moment analysis and a set of outputs labeled "pkpm_vars" which included derived quantities necessary to reconstruct the lab-frame distribution function and to check robustness). 

A variety of non-trivial test cases which show the utility of PKPM now work, including low guide field reconnection, and a further reorganization of the "G0 input files" should commence to move certain production calculations out of the gkylzero repo and into a proper input file repo.


JunoRavin:
Oh one more additional comment on the feature extension in this PR:
An improvement to the PKPM robustness also involved a novel way of computing the interface values of certain quantities such as u and b_hat in the surface fluxes. Since u and b_hat are derived by weak division, and weak division does not produce a point-wise quantity, we separately compute quantities needed at the surfaces with weak operations *on the surface*. This structure produces a more accurate solution on coarse resolutions and has the added benefit of maintaining higher order even when our robustness checks are triggered. For example, we were reducing the order of b_hat to only using the cell averages everywhere in the volume if bb was negative at quadrature points (i.e. regions where one component of the field is close to 0). Now, because we compute, e.g., bx at x surfaces, we are guaranteed that we will at least have linear variation across the cell because, even is bxbx is close to 0, bxbx(xl) =/= bxbx(xr) in the DG expansion (so we can use the cell average on a surface instead of just the cell average in the volume).

ammarhakim:
This looks like a huge amount of work and so it will be hard to fully review properly. What I want @JunoRavin to do is to run valgrind on the PKPM tests (just 1 or 2 steps with -s1 or -s2) and make sure there are no memory leaks. Then we can go over the key parts of the code before accepting the PR.

PR #193, opened 2023-08-31
Elliptic integrals and other changes to facilitate mirror modeling with fluid solver

JunoRavin:
Branch includes a number of changes to facilitate mirror modeling with the multi-fluid solver, specifically CMFX, including:
1. Addition of elliptic integral functions in C, to allow for the initialization of A_phi of a current loop using Jackson 3rd ed. eq. (3.37)
2. An ability to slowly turn on the E field and applied currents which appear in the source terms of fluid equations via a linear ramp, so that all the external quantities can still be computed once (since projecting the external fields is 100x more expensive than a timestep of the fluid solver)
3. An improved Lax flux for isothermal Euler and 10 moment based on the revamped, two-wave, Lax flux in wv_euler


manauref:
@JunoRavin did you code the elliptic integrals yourself? If not, can you add comments in the code detailing what author/website they came from please?

PR #194, opened 2023-09-05
Modify dg_diffusion objects to support more (kinetic) options

manauref:
(this goes with gkylcas PR 24 https://github.com/ammarhakim/gkylcas/pull/24)

Create new dg_diffusion objects so that now we have
dg_diffusion_fluid
dg_diffusion_vlasov
dg_diffusion_gyrokinetic
dg_diffusion_gen

The first 3 support the operator \partial_i^(n/2) ( D_i(x) \partial_i^(n/2) f ) where i=1,2,3 and n=2,4,6. When user chooses a constant diffusion coefficient, different kernels are employed.

For vlasov and gyrokinetic the maxima, C and Lua code can be easily extended to support velocity space diffusion, but I have not added that option for now since I don't know of anyone trying to use it right now.

diffusion_const reg tests are not affected on CPU. Also ran a diffusion_const reg test on GPU and results looked the same.


JunoRavin:
Please regenerate and commit the tensor p=2 kernels for fluid diffusion. The PKPM model right now only works with tensor bases due to how some auxiliary variables have to be computed and so for p>1 I run with the tensor basis diffusion. But other than that I think this PR looks good. I redid the convergence studies for grad^4 and grad^6 with your new kernels and got what I expected
grad4-diffusion
grad6-diffusion

manauref:
I added tensor kernels to dg_diffusion_fluid @JunoRavin

PR #196, opened 2023-09-09
Fixed memory leaks in fem_poisson

ammarhakim:
There were some memory leaks and misallocated arrays in fem_poisson and superlu_ops. These are fixed. The fem_poisson unit tests are now fully valgrind clean! All is good with the world again.


ammarhakim:
Closing PR as one more issue needs resolution.

ammarhakim:
Seems that this PR is indeed ready to be merged. So I am reopening it. The issue Mana found is a problem with the dg bin ops not allowing different ranges for the two input arrays. This is a more elaborate problem and will need careful look to fix throughout the code. Can't do it in this branch

ammarhakim:
Reopening PR.

PR #198, opened 2023-09-11
2d cirular domain modeling aiming to mimic the intersection of LAPD

liangwang0734:
Using the Cornerific Tapered2 Mapping from [this paper](https://arxiv.org/pdf/1709.07875.pdf).

- This is 2d only as a starting point.
- The time step size seems to scale as 1/N^2 instead of 1/N.
  - Why?
  - Might not matter for 3d extension since dz is guaranteed to scale as 1/N.
- Have slightly strange matters near bad cells but overall OK.

The mapping looks like the following (with color lines to help indicating mapped lines):
indexCornerific Tapered2 Mapping

The Ex and Bz from a 128x128 run looks like the following
03-frame30-ExBz

### DIFF
--- a/regression/rt_5m_lapd_2d_circ.c
+++ b/regression/rt_5m_lapd_2d_circ.c
@@ -1,0 +1,560 @@
+#include <math.h>
+#include <stdio.h>
+
+#include <gkyl_alloc.h>
+#include <gkyl_const.h>
+#include <gkyl_moment.h>
+#include <gkyl_util.h>
+#include <gkyl_wv_euler.h>
+#include <gkyl_wv_iso_euler.h>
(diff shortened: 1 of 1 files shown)


manauref:
This is pretty cool Liang. You say there are strange things near the corner
cells. Can you show an example? those two plots above look smooth near the
corners

ammarhakim:
This looks very nice. I wonder if what is going is that the cells near the "corners" are getting so squished that their local area scales as 1/N^2, leading to a more stringent time-step limit that one would expect. I guess this is equivalent to saying that we now moved the small sized cell from the center to the edges but kept the time-step restriction still the same.

PR #202, opened 2023-09-13
FEM Poisson signature changed so need to adjust the regression test of FEM Poisson timing.

JunoRavin:


manauref:
@JunoRavin 
FYI to avoid future bugs
The following code
```
  struct gkyl_array *epsilon = mkarr1(use_gpu, basis.num_basis, localRange_ext.volume);
  gkyl_array_shiftc(epsilon, 1.0, 0);
```
is not equivalent to setting epsilon to 1. We need
```
  gkyl_array_shiftc(epsilon, 1.0*pow(sqrt(2.),basis.ndim), 0);
```
to take care of the normalization

manauref:
also not ideal that no objects are released in this reg test. But maybe it's ok since it's such a small program

PR #212, opened 2023-09-26
Find cell

jRoeltgen:
Only CPU version has been written/tested. Binary search to find cell indices in rect_grid.

### DIFF
--- a/unit/ctest_rect_grid.c
+++ b/unit/ctest_rect_grid.c
@@ -36,17 +36,37 @@ void test_grid_2d()
 }
 
 void test_find_cell(){
-  double lower[] = {1.0, 1.0}, upper[] = {2.5, 5.0};
-  int cells[] = {20, 20};
-  double point[] = {2.0, 2.0};
+  double lower[] = {0.0, 0.0, 0.0}, upper[] = {5.0, 5.0, 10.0};
+  double lower2[] = {0.0, 1.0, 3.2, 7.1}, upper2[] = {5.1, 5.0, 22.0, 8.2};
+  int cells[] = {5, 5, 10};
+  int cells2[] = {10, 11, 20, 3};
+  double point[] = {2.5, 0.1, 4.1};
+  double point2[] = {2.01, 3.1, 4.1, 8.05};
   bool pickLower = false;
-  const int *knownIdx[2] = {NULL, NULL};
+  int idx = 1;
+  const int *ptr = &idx;
+  const int *knownIdx[3] = {NULL, ptr, NULL};
+  const int *knownIdx2[4] = {NULL, NULL, NULL, NULL};
   int *cellIdx;
+  int correctIdx[3]={3,idx,5};
+  int correctIdx2[4]={4,6,1,3};//Got 6,9,2,6
   struct gkyl_rect_grid grid;
-  printf("Start test_find_cell\n");
-  gkyl_rect_grid_init(&grid, 2, lower, upper, cells);
+  struct gkyl_rect_grid grid2;
+  
+  printf("\nStart test_find_cell\n");
+  gkyl_rect_grid_init(&grid, 3, lower, upper, cells);
   cellIdx = gkyl_find_cell(&grid, point, pickLower, knownIdx);
-  TEST_CHECK(false);
+  for(int i=0;i<3;i++){
+    printf("i=%i, cellIdx[i]=%i\n",i,cellIdx[i]);
+    TEST_CHECK( cellIdx[i]==correctIdx[i]);
+  }
+
+  gkyl_rect_grid_init(&grid2, 4, lower2, upper2, cells2);
+  cellIdx = gkyl_find_cell(&grid2, point2, pickLower, knownIdx2);
+  for(int i=0;i<4;i++){
+    printf("i=%i, cellIdx[i]=%i\n",i,cellIdx[i]);
+    TEST_CHECK( cellIdx[i]==correctIdx2[i]);
+  }
   printf("End test_find_cell\n");
 }
 
--- a/unit/ctest_rect_grid.c
+++ b/unit/ctest_rect_grid.c
@@ -35,39 +35,137 @@ void test_grid_2d()
     }
 }
 
-void test_find_cell(){
-  double lower[] = {0.0, 0.0, 0.0}, upper[] = {5.0, 5.0, 10.0};
-  double lower2[] = {0.0, 1.0, 3.2, 7.1}, upper2[] = {5.1, 5.0, 22.0, 8.2};
-  int cells[] = {5, 5, 10};
-  int cells2[] = {10, 11, 20, 3};
-  double point[] = {2.5, 0.1, 4.1};
-  double point2[] = {2.01, 3.1, 4.1, 8.05};
+void test_find_cell_1x(){
+  double lower[] = {0.0}, upper[] = {5.0};
+  int cells[] = {5};
+  double point[] = {2.5};
   bool pickLower = false;
   int idx = 1;
   const int *ptr = &idx;
-  const int *knownIdx[3] = {NULL, ptr, NULL};
-  const int *knownIdx2[4] = {NULL, NULL, NULL, NULL};
-  int *cellIdx;
-  int correctIdx[3]={3,idx,5};
-  int correctIdx2[4]={4,6,1,3};//Got 6,9,2,6
+  const int *knownIdx[1] = {NULL};
+  const int *knownIdx2[1] = {ptr};
+  int cellIdx[]={0};
+  int correctIdx[1]={3};
   struct gkyl_rect_grid grid;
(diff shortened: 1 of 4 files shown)


zero/gkyl_rect_grid.h
```diff
@@ -29,6 +29,19 @@ struct gkyl_rect_grid {
 void gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
   const double *lower, const double *upper, const int *cells);
 
+/**
```
manauref, line 32: @jRoeltgen fix typos in these comments

zero/gkyl_rect_grid_priv.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
```
manauref, line 1: @jRoeltgen 
i don't think you need a private header just to declare the signature of private functions. Either just place those functions in the .c file before they get used, or put the functions themselves in the private header.

manauref: although I understand why one may want to do it this way, we just haven't AFAIK, so let's stick to the current model for now.

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
   }
 }
 
+GKYL_CU_DH
+void gkyl_rect_grid_find_cell(struct gkyl_rect_grid *grid, const double *point, bool pickLower, const int **knownIdx, int *cellIdx){
+  int nDim = grid->ndim;
+  int searchNum = 0;
+  int searchDim[GKYL_MAX_DIM], *idxOut;
```
manauref, line 31: I think you are not using idxOut. Delete.

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
   }
 }
 
+GKYL_CU_DH
+void gkyl_rect_grid_find_cell(struct gkyl_rect_grid *grid, const double *point, bool pickLower, const int **knownIdx, int *cellIdx){
```
manauref, line 28: If I understand correctly, you are declaring a bunch of variables at the beginning of your function. That's a common style in Fortran, and some C programmers. However I (or maybe even we) prefer that you declare variables as needed. So please move them to right before they are used. The benefit of this latter style is that you keep the definition closer to where it is used, and don't require the reader to memorize the definition to later understand or check its (correct) use

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
...
+  int searchDim[GKYL_MAX_DIM], *idxOut;
+  int *dimTrans[GKYL_MAX_DIM];
+  int plusminus[2] = {-1,1};
+  bool allLessEq = true;
+  int iStart[GKYL_MAX_DIM], iEnd[GKYL_MAX_DIM], iMid[GKYL_MAX_DIM], iNew[GKYL_MAX_DIM];
+  int *cells, lowHighCellIdx[2*GKYL_MAX_DIM];
+  double lowerInDir[nDim], upperInDir[nDim];
+  double low, high;
+  
+  for(int d=0; d<nDim; d++){
```
manauref, line 40: here and elsewhere, please use spacing like other gkeyll code.
In this case leave a space after **for**. Below leave a space after **if**, and before and after **else**.

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
...
+    iEnd[d] = cells[searchDim[d]];
+    iMid[d] = 0;
+    iNew[d] = 0;
+  }
+  /* Below we use a binary search. That is, if the i-th coordinate of the point in
+   * question is below(above) the ith-coordinate of the current mid-point, we search
+   * the lower(upper) half along that direction in the next iteration.
+   */
+  while (allLessEq){
+    for(int d=0; d<searchNum; d++){
```
manauref, line 67: Here and elsewhere:
- please remove the curly brackets in for loops that contain a single instruction.
- Leave at least 2 spaces between code and comment.
- Leave 1 space after // in your comment.

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
...
+    if(isInCell(grid, point, iMid, dimTrans, knownIdx)){
+      //Check if neighboring cells also contain this point.
+      for(int k=0; k<searchNum; k++){
+	iNew[k] = iMid[k];
+	lowHighCellIdx[searchDim[k]] = iMid[k];
+	lowHighCellIdx[nDim+searchDim[k]] = iMid[k];
+      }
+      for(int i=0; i<searchNum; i++){	
+	for(int j=0; j<2; j++){
+	  iNew[i] = fmax(fmin(iMid[i] + plusminus[j],cells[i]),0);
```
manauref, line 80: i think fmax and fmin are for floats/doubles (maybe I'm wrong).
There's a function like GKYL_MAX or something that you can use instead (see zero/gkyl_util.h).

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
...
+  return inCell;
+}
+
+GKYL_CU_DH
+void InDir(const struct gkyl_rect_grid *grid, int *iIn, int *dimTrans[1], const int **knownIdx, double lowerInDir[], double upperInDir[]){
+  int nDim, checkIdx;
+  const double *dx, *lower;
+  nDim = grid -> ndim;
+  double testarr[nDim];
+  setbuf(stdout, NULL);
```
manauref, line 147: what is this?

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
...
+	allLessEq = false;
+	break;
+      }
+    }
+  }
+
+}
+
+GKYL_CU_DH
+bool isInCell(const struct gkyl_rect_grid *grid, const double *pIn, int *iIn, int *dimTrans[1], const int **knownIdx){
```
manauref, line 119: please add some (very brief) comments describing this function and its inputs/outputs.

zero/rect_grid.c
```diff
@@ -22,6 +24,138 @@ gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
...
+    if(lowerInDir[d]-eps>pIn[d] || upperInDir[d]+eps<pIn[d]){
+      inCell = false;
+      break;
+    }
+  }
+  return inCell;
+}
+
+GKYL_CU_DH
+void InDir(const struct gkyl_rect_grid *grid, int *iIn, int *dimTrans[1], const int **knownIdx, double lowerInDir[], double upperInDir[]){
```
manauref, line 142: please add some (very brief) comments describing this function and its inputs/outputs.

zero/rect_grid.c
manauref, line 28: Biggest change I would recommend is to change knownIdx from a ** to a *. The latter will be easier to work with. We can assume that entries in knownIdx <0 are not known (like nil in the Lua code), and entries >=0 are known indices.
If you agree, please change throughout.

unit/ctest_rect_grid.c
```diff
@@ -35,6 +35,139 @@ void test_grid_2d()
     }
 }
 
+void test_find_cell_1x(){
```
manauref, line 38: please add some (very brief) comments to your testing functions to at least say which case you are testing in each bit of code. So future readers know the purpose of each bit.

manauref:
@jRoeltgen you technically haven't tested your method on GPUs. Although you may have run on/with a GPU, it executed the code on the CPU.

To test it on the GPU you'll have to add code to uni/ctest_rect_grid_cu_.cu

ammarhakim:
Please do not use camelCase for names. Please be consistent with the rest of the code and use underscores to name functions. Please make these changes before we consider the PR.

zero/rect_grid.c
ammarhakim: Do not use Fortran style :) I agree with Mana

zero/rect_grid.c
jRoeltgen: Should the curly brackets be neglected from the outer for loop in double for loops even when the inner for loop has multiple instructions (assuming the outer for loop only contains the inner for loop), such as the following?

for (int i=1; i<a; i++) {  // this curly bracket
   for (int j=1; j<b; j++) {
      x=i+j;
      .
      .
      .
   }
}

manauref: I think not, good catch. An exception to my first bullet above is when that single instruction is another loop or switch

PR #218, opened 2023-10-24
Fix bug in gk mom private header. Seems benign but it appears it is not.

manauref:
@akashukla can you check if this fixes the issue you were seeing and merge if it does?


akashukla (approved):
This does seem to fix the problem.

PR #220, opened 2023-10-26
Add -u to strip command to fix Mac build.

manauref:


ammarhakim:
The -u command does not exist on Linux. We can't merge this change.

PR #224, opened 2023-11-13
Adding 'static' to inline functions

pcagas:
Adding `static` to all `inline` functions in the wv_eqn header. This was causing some compilation issues; for example, it was impossible to build the regression tests on Linux without the optimization, i.e. with `-O0`


ammarhakim:
The problem is that when you turn off optimization the functions are not inlined. This will lead to duplicate symbols. In general one should never build without optimization. But this change is correct.

PR #226, opened 2023-11-20
change GKYLSOFT to PREFIX in machine files in a way that allows external specification

manauref:
so that g2 can specify these env variables


ammarhakim:
Is this syntax compatible with all the shells people use? Was it tested on Linux and our clusters?

PR #228, opened 2023-11-28
Change the location of some default LAPACK, SuperLU, MPI and LuaJIT p…

manauref:
This was needed for non-default install locations.

I believe we thought non-default install location was working for g0-g2 because it was looking for these libraries in $HOME/gkylsoft, and we all had an install there from previous work. But if that doesn't exist it in fact doesn't work. This PR fixes that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ CUDA_ARCH ?= 70
 CFLAGS ?= -O3 -g -ffast-math -fPIC -MMD -MP
 LDFLAGS = 
 PREFIX ?= ${HOME}/gkylsoft
-INSTALL_PREFIX ?= ${PREFIX}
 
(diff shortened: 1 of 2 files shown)


ammarhakim:
All non-default paths should be set via the various flags. This change is not needed as non-standard paths are already supported, I believe.

manauref:
@ammarhakim non-default install locations are quite common, especially for developers working on multiple things at the same time. We should make it easy to do that, and it is with this change, so that it all follows the PREFIX path. Also, users (even most developers) do not necessarily know what all the paths they should set are, so it's best to use PREFIX

ammarhakim:
Please do not mod the Makefile. It is ok to have the install prefix in it. Revert that change and reopen the PR.

PR #237, opened 2024-01-05
Find cell method

jRoeltgen:
Find cell binary search method in rect_grid.


zero/gkyl_rect_grid.h
```diff
@@ -29,6 +29,20 @@ struct gkyl_rect_grid {
 void gkyl_rect_grid_init(struct gkyl_rect_grid *grid, int ndim,
   const double *lower, const double *upper, const int *cells);
 
+/**
+ * Find cell indices of point
+ *
+ * @param grid Grid object
+ * @param point The point to find the cell indices at
+ * @param pick_lower
```
manauref, line 37: @jRoeltgen can you add a description for this argument please?

zero/gkyl_rect_grid_priv.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include <gkyl_rect_grid.h>
+
+/* Checks if given point is in given cell
```
manauref, line 5: @jRoeltgen please revise these comments, I think they are not right for in_dir

zero/gkyl_rect_grid.h
jRoeltgen: Added "If point on cell boundary, pick lower cell if true, and upper if false"

zero/gkyl_rect_grid_priv.h
jRoeltgen: Changed "Checks if given point is in given cell" to "Find upper and lower boundaries of given cell"

PR #257, opened 2024-01-27
Cleaning up memory errors and first pass at reactions implementation (which was needed to clean up memory errors)

JunoRavin:
This branch makes the GK G0 app (mostly) valgrind clean through a couple of not-insubstantial changes
1. There is now a separate projection object, modeled after some of the ideas in the Projection object on the G2 side. Essentially we instantiate a projection object with a particular ID (a function, a Maxwellian, or a bi-Maxwellian) and fetch the appropriate pointers to the necessary input functions to do this projection. This moves hundreds of lines of code out of initial conditions and sources into its own self-contained set of functions which can then cleanly handled the allocations and frees (it also extends our functionality to now also be able to source bi-Maxwellians, should we desire to do so)
2. The reactions struct is now filled and initialized based on the desired forms of ionization and recombination, with the latest versions of the adas_neutral branch pulled in. This is not yet fully functional as @tnbernard is still restructuring the ionization and recombination updaters to *only* return the reaction rates, letting the app itself handle how the final distribution function update is computed. The structure of these functions is also designed to make species parallelization as easy as possible, following the assumption that computing the configuration space fields is a sub-dominant cost and every single species can compute all the needed configuration space fields (such as the reaction rate). *Note: right now this struct is giving a warning because we made some enums generic instead of specific to ionization and recombination. @tnbernard is refactoring based on this but I don't want this branch to be unmerged before I am unavailable*

All of the input files have been redone to follow this new structure with the projection object. I believe I did this without error and I've ran a number of them and checked that things initialize appropriately. I could still have made mistakes and if anyone catches a simulation behaving strangely, check to make sure the electrons are initializing electron quantities/ions are initializing ion quantities and I did not accidentally type a name incorrectly.


manauref:
I'm gonna take a brief look at the superLU memory leak to see if I can push a fix here

zero/calc_derived_geo.c
manauref, line 1: @tnbernard @JunoRavin
can you update the public header for this with accurate descriptions of the functions and its arguments please?

manauref (changes requested):
I ran make check and i'm seeing some failures. Specifically:
```
build/unit/ctest_array_integrate
Test test_1x_cpu...                             /bin/sh: line 1:  9231 Segmentation fault: 11  build/unit/ctest_array_integrate -E
```

```
build/unit/ctest_correct_maxwellian_gyrokinetic
Test test_1x1v_p1...                            ctest_correct_maxwellian_gyrokinetic(9261,0x7ff853e4c640) malloc: Heap corruption detected, free list is damaged at 0x600001a940c0
*** Incorrect guard value: 0
ctest_correct_maxwellian_gyrokinetic(9261,0x7ff853e4c640) malloc: *** set a breakpoint in malloc_error_break to debug
/bin/sh: line 1:  9261 Abort trap: 6           build/unit/ctest_correct_maxwellian_gyrokinetic -E
```

```
Test dg_gyrokinetic...                          [ FAILED ]
  ctest_dg_gyrokinetic.c:77: Check gyrokinetic->conf_range.volume == 32768... failed
FAILED: 1 of 1 unit tests has failed.
build/unit/ctest_dg_gyrokinetic_kern_tm
Test test_gyrokinetic_3x2v_p1...                /bin/sh: line 1:  9271 Segmentation fault: 11  build/unit/ctest_dg_gyrokinetic_kern_tm -E
```

and these tests either hang or take forever to run on my Mac (and I cancelled them rather than continue to wait):
ctest_dg_rad_gyrokinetic
ctest_mom_gyrokinetic

Also, I noticed the following warnings during compilation:
Screenshot 2024-01-29 at 3 41 53 PM

JunoRavin:
@manauref I believe all of your comments have been addressed. Let me know if I missed anything

manauref:
@JunoRavin @tnbernard what about this test
Screenshot 2024-02-05 at 10 27 29 AM

manauref:
@JunoRavin @akashukla 
this is saying that there are conflicts in gk_field and gk_geometry_tok. Can you resolve them please? (there are merge problems in gk_geometry_mirror too but I've asked @Maxwell-Rosen to remove those changes from gk-g0-app since they are not ready, or I'll do it).

JunoRavin:
@manauref I will work on the conflicts and get back to you. The ADAS test fails unless you run the python script to get the ADAS data. @tnbernard may have ideas for how we want to handle this in general since it's not a default part of the build to fetch the ADAS data (yet)

manauref:
@JunoRavin the reverts I pushed to gk-g0-app resolved some of the conflicts. There's still one in apps/gk_field and we'll wait for @tnbernard on the seg fault in iz and recomb unit test.

@tnbernard one possible solution, if I understand what is going on correctly, is that the configure+Makefile files set a flag, like GKYL_HAVE_ADAS, so that you can wrap certain things in an #ifdef and we don't have problems with it when ADAS data is not downloaded/installed.

manauref (approved):

PR #262, opened 2024-02-04
Gk g0 app spitzernu hookin

akashukla:
Spitzer Nu has been hooked in using normNu. Regression tests for relaxing a bump on tail instability using both constant coll frequency and time dependent collision frequency have been added. Comparisons with g2 indicate it is working and is not affectng the constant collision frequency case. Please review to see if anything seems wrong.


apps/gkyl_gyrokinetic_priv.h
```diff
@@ -176,6 +176,8 @@ struct gk_lbo_collisions {
   struct gkyl_array *nu_sum_host, *prim_moms_host, *nu_prim_moms_host; // LBO primitive moments host-side for I/O
   bool normNu; // Boolean to determine if using Spitzer value
   struct gkyl_array *norm_nu; // Array for normalization factor computed from Spitzer updater n/sqrt(2 vt^2)^3
+  double *norm_nu_facs; // // List of collision freqs without factor of n_r/(v_ts^2+v_tr^2)^(3/2)
```
JunoRavin, line 179: Why is this a pointer? There seems to be an inconsistency here in how this normalization factor is defined between the public and private headers.

JunoRavin:
This is great work @akashukla but I think there should be a change to the user-facing side for right now to avoid confusion in the setup even if we change this later.

Please define two input norm_nu_facs. One is a double and call that self_nu_fac. The other is a double[GKYL_MAX_SPECIES] and call that cross_nu_fac. Then initialize cross_nu_fac in the cross_init and use that in the normalized nu computation in the cross_moms updater.

I don't want us to leave things like [i+1] accesses inside cross_moms. We may want to change the API for self vs. cross collisions, but I would like the APIs to stay separate for self collisions and cross collisions for now because we do this everywhere else as well.

PR #271, opened 2024-02-08
Modify dg_diffusion_gyrokinetic so the inner most derivative is on f, not J*f

manauref:
So with these changes the operator currently is
```
d(J*f)/dt_diff = d/dx( J*D d(J*f/J)/dx )
```

where J is the field-aligned coords Jacobian. The inner most division is done inside the diffusion kernels, multiplying by 1/J. This introduces more FLOPS than we'd like but saves us the trouble of having to allocate another f (this seems like the kind of thing shared memory on GPUs would be good for).

Tested with a STEP sim without drifts and there was no apparent change.

That said, my current understanding is that it should actually be
```
d(J*f)/dt_diff = d/dx( J*g^xx*D d(J*f/J)/dx )
```
but g^xx for STEP is ~80, so the time step tanks, by at least 1e4X in fact, so maybe something else was wrong. But this makes me think that earlier comparisons of SOLPS' D and Gkeyll's D where not quite right, and that we should actually compare the flux-surface averaged transport equations, because g^xx is different in these two codes.


ammarhakim:
Is this operator for the cross-field diffusion term?

PR #273, opened 2024-02-09
Mod BC and source input file interface, and implement GK_IWL option

manauref:
The overall intent here is to implement the BC options for 2x inner-wall-limited (IWL) simulations. This entailed:

- Adding the GK_IWL BC option for the species.
- Creating a new ES_IWL field id.
- Upgrading the BCs given in the input file from simple enums to a struct, so we can pass a function and a parameter for each BC too. This is used here to pass the location of the LCFS but other BCs could benefit from this too: twist-shift, maxwellian ghost, and others.
- For clopen sims we may want to add an aditional, grad-B source, so here we change the source interface in the input file so to allow the option to specify multiple sources.

d3d sim runs for 1 ms in 10 hours on an A100. There are a bunch of funky things about it though, so this needs to be explored more and understood. We know that modifications to the perp field solver may be necessary (that's more in depth work that may need to take place later). In the meantime i would suggest, in addition to more analysis, trying:
- Reducing the ref temp if it's too high for the SOL
- Doing convergence on velocity space limits
- Trying the FEM_PARPROJ_DIRICHLET BC option for the parallel smoother in the SOL
- Using a Neumann BC for phi in the core boundary
- Running this more than 1 ms (AFAICT it hasn't saturated, or maybe it never will)


tnbernard (commented):
Regarding the ref temp: in my other d3d tests, it is difficult to match the temperature profiles at the inner radial boundary to experimental values unless the reference temperatures are that high. Otherwise the higher temperatures are not realizable on the grid as well. I know this means the lower temperatures are not as well resolved, particularly Tperp. It depends which reason you want to prioritize until we add non-uniform vspace grid mapping.

manauref:
i really wouldn't try to match the inner radial boundary to experiment. Too much numerical artifice is happening there. You should do your analysis some distance away from that boundary anyway.

Your point may still remain, but I think what we need to do is both lower the ref temp (if the SOL calls for it) and increase v-limits so that the core is captured well.

PR #275, opened 2024-02-13
Gk g0 app mirror

Maxwell-Rosen:
This merge includes the non-uniform mesh into the mirror simulations.


ammarhakim:
We will re-open this PR after we have looked at this with more care. I am closing it so it does not get merged by mistake.

Issue #276, opened 2024-02-14
5-moment Rayleigh-Taylor instability is failing


JonathanGorard:
The rt_5m_rt (5-moment Rayleigh-Taylor instability) regression test is failing on all but one time step. The output appears qualitatively wrong, and there is an absurdly large amount of memory being allocated and deallocated (presumably because of all the failed time-steps).

I believe that I have eliminated a few of the obvious sources of error. Specifically, the problem persists even when one disables the application of accelerations, tweaks the boundary conditions, and/or lowers the CFL coefficient. It also persists when one sets .force_low_order_flux=true for each species, indicating that the issue is probably not (as @JunoRavin first suspected) simply some kind of automatic fallback from higher-order/Roe fluxes to first-order/Lax fluxes.

All of this suggests that something quite fundamental is wrong (or at the very least has changed) within the moment app implementation. I'm going to look into this at some point soon, but in the meantime, if anyone has any suggestions as to what this might be, please do comment and/or message me.

Issue #188, opened 2023-08-22
Remove "regression tests" that are not really regression tests!


ammarhakim:
We are using the regression test folder to store simulation C code. These are not really regression tests at all. A regression test should run in a min or two (max) and not require hours. Very soon we will want to run all tests automatically and so clearly this must be cleaned up.

I recommend to move all these files to the inp-files repo so they are built from a G0 installed. A Makefile would then compile the input C so sims can be done.

We should do this ASAP.

JunoRavin:
Since this issue is quite broad and many regression tests have been refactored and removed in the interim since this issue was file, I am closing it for now. When the automatic regression system is revived shortly now that https://github.com/ammarhakim/gkylzero/pull/594 is complete, we will remove any straggler high cost tests, or more clearly define what is a high cost test that is not run nightly, but perhaps weekly, to test production capabilities in a controlled fashion.

PR #278, opened 2024-02-15
Began refactoring of regression system (for the moment app)

JonathanGorard:
I'm working my way through the moment app regression tests and attempting to refactor them somewhat systematically, so I figured that I might as well request a code review before I go too much further. A few things to note:

- Despite being systematic, there is one test (namely the 5-moment LAPD 2d circular test) that I have not refactored, since it will clearly require significant work, and I didn't want to proceed with that before obtaining a sanity-check on the rest of the design.
- The 5-moment Rayleigh-Taylor instability test does not work (see issue #276), and therefore, although the code has been refactored, it is currently excluded from our automated regression-testing system, and will remain so until the underlying bug in the moment app has been tracked down.
- The runtime of several of these tests was too high (see issue #188) for them to act as effective regression tests; in all such cases (e.g. the 10-moment parallel firehose and Burch gradient-closure tests) I have simply dropped the final time of the simulation to get the runtime below one minute on my laptop.
- I think that I was able to track down reasonable sources for all of the test cases I've dealt with so far, although in many instances these sources may not be ideal or canonical, so feel free to correct me. However, I have absolutely no idea where the "clouda" test for the cold fluid equations came from, or even what it might have been adapted from. If anyone (looking at you, @ammarhakim) knows, please do send me a reference.


JunoRavin:
Couple of comments (some of these are pedantic so apologies in advance):

1. Can we change speed_light -> light_speed. I think I am just so used to doing light_speed (or lightSpeed) that speed_light looks wrong. I understand this is not quite in the convention of thing_modifier, but since we don't typically follow this convention for speeds/velocities (because the modifier is usually species and the thing is the particular velocity, like vt or vA) I figure this is an okay thing to ask
2. What would you think about changing vt_elc -> vte (and vt_ion -> vti). That way the definition of the thermal velocities looks more like the Alfvén velocities (vAe and vAi) and other quantities such as the inertial lengths (di and de)?
3. The citation for the Burch event simulation is this one by @jtenbarg: https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2019JA026731
4. The citation for the parallel firehose simulation is this one from the Pegasus code methods paper: https://www.sciencedirect.com/science/article/abs/pii/S0021999113007973
5. I would be inclined to not define pi by hand and just set pi = M_PI, but I do think this is reasonable to have pi in the struct.

JonathanGorard:
@JunoRavin
Thanks so much for the comments! Please don't apologize - (productive) pedantry is what this is all about ;)
(1): yes, I admit that I was essentially just blindly following the thing_modifier convention with "speed_light". But if you think that this deserves to be treated as a special case then I'm happy to revert to "light_speed".
(2): you make a good point about the consistency between Alfvén and thermal velocities. Ideally it'd be nice to use i and e as our general suffixes for species distinctions, but there are some conventions (such as mass_elc/charge_elc) which seem so deeply embedded that it would probably be silly to try to change them. I naively thought that thermal velocities might be another one of these, but you're right - it likely isn't.
(3) and (4): thank you! These are harder to track down than I anticipated, but I'll hopefully get better...
(5): yes, the interchangeably use of hard-coded pi, M_PI and GKYL_PI is a strange feature of the code. Setting pi = M_PI as part of the struct seems like a good compromise (and, based on what I understand of what @ammarhakim has said, may be more likely to play nicely with Lua).

JunoRavin:
With respect to the charge and mass convention, part of that comes from the fact that (although people are inconsistent sometimes), the charge and mass are always needed as specific inputs to the species structs, so the naming convention for charge and mass sprung out of "whatever name you give the species table/struct, that should be the name of the charge/mass"

Whereas quantities such as thermal velocity, Alfvén velocity, plasma frequency, cyclotron frequency, etc. are not direct inputs anywhere and are derived quantities utilized to set things like the velocity space extents, the end time of the simulation, the box size in some set of normalized units. 

Like I make this distinction because in a non-trivial number of simulations, we initialized a proton species called "ion" so the charge and mass were charge_ion and mass_ion, but then we ended up using a suffix p for things like vtp, dp, etc. (sometimes because there was a third species which was another ion species called something like Ar for Argon so we needed to distinguish vtp from vtAr and it was better to have vtp in that case than vt_ion). 

The alternative (which would hurt my brain but is perhaps better for longevity) is to follow this convention everywhere. So:

1. vt -> vt_elc/vt_ion
2. vA -> vA_elc/vA_ion
3. d -> d_elc/d_ion
4. rho -> rho_elc/rho_ion
5. omega_p -> omegap_elc/omegap_ion
6. omega_c -> omegac_elc/omegac_ion

and so on. My only hesitation on this is I don't want to do double underscores and we are so used to doing omega_p and omega_c that switching that would be a non-trivial cultural change. And I don't know of good abbreviations for "proton" in the case where we want to distinguish a proton from another ion species (whereas we have historically been quite comfortable saying the species named ion is always a proton/H+ and then other ions get more descriptive names like Li1 for Li+ or Li2 for Li++).

manauref:
2. I generally favor not differentiating variables by a single letter where
possible/not too inconvenient, in case someone types vti when they actually
meant to type vte. I think it's harder to make this mistake with _elc and
_ion rather than e and i.

JonathanGorard:
@manauref do you have thoughts on @JunoRavin's point above regarding the potential "double underscore problem" when using _elc and _ion?

JonathanGorard:
@JunoRavin even with the omegaP_elc, etc. workaround, this rapidly starts to become comical in the case of some other quantities. Do we really want to be writing `T_ion1_over_T_ion2`?

JunoRavin:
@manauref the gyrokinetic system is going to encounter similarly long names when defining the temperatures in different regions of the tokamak. What do y'all currently do for the ion temperature in the core vs. the edge. Do you do T_ion_core and T_ion_edge or something else?

manauref:
Some double under scoring OR mid-word capitalization is necessary,
especially as things become more complex. I'm fine with either, but I
prefer clarity and safety over convenience.

JonathanGorard:
I don't really have a dog in this fight, but I do feel like there is some clarifying advantage in the non-underscored single letter convention, beyond mere convenience. If you take something like `Ti_perp` vs `T_ion_perp` - in the first case, it is clear that "perpendicular" is referring to the ion temperature, whereas in the second, the referent is ambiguous (is it the perpendicular ion temperature, or the temperature of the perpendicular ion?)

manauref:
Yes, this is fine. I prefer full words, but I don't follow that rule
absolutely either.

manauref:
More precisely, I don't follow that preference absolutely. It's not a rule,
and if it is, I break it all the time.

JunoRavin (approved):

PR #291, opened 2024-03-03
Implement adiabatic elc option (only for 1x). Rename previous GK_FIELD_ADIABATIC to GK_FIELD_BOLTZMANN

manauref:
- Rename the GK_FIELD_ADIABATIC option to GK_FIELD_BOLTZMANN, since in 2 papers we've referred to that as the Boltzmann electron model.
- Implement a new GK_FIELD_ADIABATIC, only in 1x for now, in which the electrons are not evolved and they simply contribute to the charge density in the Poisson equation with n_e = n_e0*(1 + e*phi/T_e).
- Tested GK_FIELD_ADIABATIC (one CPU and GPU) with one of the ion acoustic Landau damping tests in the GkLBO paper and it qualitatively looks very similar, and quantitatively is only 12% off (but I didn't try to do resolution convergence).

This also fixes a bug for 1x simulations with geometry in that the polarization weight was not multiplied by jacobgeo.

### DIFF
--- a/apps/gk_field.c
+++ b/apps/gk_field.c
@@ -42,7 +42,7 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
   f->epsilon = 0;
   f->kSq = 0;  // not currently used by fem_perp_poisson
   double polarization_weight = 0.0; 
-  if (f->gkfield_id == GKYL_GK_FIELD_ADIABATIC) {
+  if (f->gkfield_id == GKYL_GK_FIELD_BOLTZMANN) {
     polarization_weight = 1.0; 
     f->ambi_pot = gkyl_ambi_bolt_potential_new(&app->grid, &app->confBasis, 
       f->info.electron_mass, f->info.electron_charge, f->info.electron_temp, app->use_gpu);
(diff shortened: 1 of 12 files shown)


Maxwell-Rosen (commented):
It all looks good to me. Thank you for correcting the bug and notation consistency issues

JunoRavin:
What are the canonical references for this distinction? I've always used Boltzmann electrons and "the electrons respond adiabatically" interchangeably. I just found some technical notes from Oak Ridge in which they say "The Boltzmann electron model in the limit of small perturbations gives the adiabatic electron model" which is manifestly what we're doing Taylor expanding the exponential in the Boltzmann model, but it's not terminology I've seen used this way in space and astrophysics because if the fluctuations in phi are large, the electrons aren't Boltzmann anyway. What's the fusion context in which you want to run Taylor expanding the exponential? Is this just an option gyrokinetic codes have always historically supported? What is being gained from Taylor expanding vs. just using the full Boltzmann response?

manauref:
I'm not following "canonical" references here, but the two Gkeyll GK
publications that have used this model (Shi et al 2015, Francisquez et al
2023). Historically in fusion gyrokinetics when you say adiabatic you mean
what this branch implements. It is linearized because the full exponential
response requires a nonlinear solver, which we don't have.

The now GK_FIELD_BOTLZMANN should really be the
GK_FIELD_BOLZTMANN_ELECTRON_WITH_AMBIPOLAR_SHEATH model, but that is too
long, so I just used GK_FIELD_BOLTZMANN. In fusion, if you call this the
adiabatic model, people are confused at first because what comes to mind is
what I mentioned above.

Issue #287, opened 2024-02-28
Package BGK operator assembly in GK app by using zero/gkyl_bgk_collisions


manauref:
currently the BGK operator
```
(nu*f_M)_sum - nu_sum * f
```
is put together, and its cflrate computed, with these 4 lines in app/gk_species_bgk (L174-181)
```
  // Accumulate RHS of BGK collisions
  gkyl_dg_mul_conf_phase_op_range(&app->confBasis, &app->basis, bgk->nu_sum_f, 
    bgk->nu_sum, fin, &app->local, &species->local);  
  gkyl_array_accumulate(rhs, 1.0, bgk->nu_fmax);
  gkyl_array_accumulate(rhs, -1.0, bgk->nu_sum_f);

  // Determine the maximum collision frequency in each cell for finding the stable time-step
  gkyl_dg_calc_average_range(app->confBasis, 0, bgk->max_nu, 0, bgk->nu_sum, app->local);
```
Rather than having 4 lines, 4 kernel launches, 4 places for potential mistakes and possible doubts about whether cflrate is computed correctly (although it seems to me that it is for the nu=const case), it'd be better to simply use the `gkyl_bgk_collisions` object in `zero/`. That way we have 1 line, 1 kernel launch, only 1 place to make a mistake (in the app) and the cflfreq is also computed within it like it is for other objects. Further more, if we do that, these lines in apps/gk_species (L399-410)
```
  // If we are doing BGK collisions, find the maximum collision frequency and accumulate it
  // onto the CFL. Collision frequency is only a configuration space quantity, so do the reduce
  // here instead of utilizing species->cflrate, which is a phase space array. 
  if (species->collision_id == GKYL_BGK_COLLISIONS) {
    gkyl_array_reduce_range(species->omega_cfl, species->bgk.max_nu, GKYL_MAX, &app->local);
    double omega_cfl_bgk_ho[1];
    if (app->use_gpu)
      gkyl_cu_memcpy(omega_cfl_bgk_ho, species->omega_cfl, sizeof(double), GKYL_CU_MEMCPY_D2H);
    else
      omega_cfl_bgk_ho[0] = species->omega_cfl[0]; 
    omega_cfl_ho[0] += omega_cfl_bgk_ho[0];
  }
```
can be removed, and we'd be treating the BGK cflrate with the same code/interface we already treat the LBO cflrate.

JunoRavin:
[...]

PR #293, opened 2024-03-04
gk-g0-app-correct

dingyunl:
Changes are:

apps/gk_species_bgk.c: modified .max_iter (from 50 to 30) and .eps_err (from 1.0e-14 to 1.0e-7) in the input struct for correct_maxwellian apps/gk_species_bgk.c
zero/correct_maxwellian_gyrokinetic.c: use the input moments to project the maxwellian if it doesn’t converge.
regression/rt_gk_bgk_3x2v_p1.c: decreased the .cfl_frac (to 0.1) for numerical stability and extended the simulation time.


apps/gk_species_projection.c
manauref, line 1: there's some less than ideal repetition of code here (in rescaling to match m0)

apps/gk_species.c
manauref, line 1: @JunoRavin @dingyunl 
it's not necessary to treat the BGK cfl any differently than the LBO's, see issue #287

PR #298, opened 2024-03-11
Touchup to MJ Density Moment 

johnson452:
Changed:
- Fixed array operations (over gkyl_array_clear/accumulate to include "_range")
- Changed to an equivalent but more accurate way of computing "n", the fluid-stationary density, for the Maxwell-Juttner distribution.


JunoRavin:
What's the reason on changing the clear methods to clear_range? If you're just setting the value to 0.0 it's fine to use clear (and actually faster on GPUs).

PR #300, opened 2024-03-12
Parallel reads and restarts

ammarhakim:
This completes the major work of enabling parallel reads and restarts. Along the way meta-data was also written to the  moments output, enabling pgkyl to use this info in plotting etc. The restarts are only hooked into the moment App, but this is easy to copy into GK and VM.

I have had to restore the Euler 2D Riemann problem header as that has the restart hooks. I will remove this once I hook the restart example into an existing regression test. Need to keep this header for now to see how this is used.


manauref:
@ammarhakim i believe you need to make the corresponding changes to nccl_comm too, otherwise multi-gpu sims might break

ammarhakim:
Need to fix the nccl IFC before the PR can be accepted.

PR #302, opened 2024-03-12
More moment app regression refactoring

JonathanGorard:
This is mostly a natural continuation of the work done as part of the previous pull request #278 . Now all 5-moment, 10-moment and Euler tests have been refactored (I think!), leaving only a couple of SR, Maxwell and MHD tests remaining on the moment app front.

However, there are a couple of other notable changes, most particularly the inclusion of a consistent scheme for writing frames to file, enforcement of various brace/indentation/etc. conventions, as well as several modifications to the way that the app communicator is set up, in an effort to achieve greater consistency with what is now being done for the GK regression system, as part of the gk-regression-refactor branch.


regression/rt_iso_euler_sodshock_dg_comparison.c
```diff
@@ -1,128 +1,309 @@
+// Sod-type shock tube test, with a small jump in state variables for comparison against Discontinuous Galerkin methods, for the isothermal Euler equations.
```
JunoRavin, line 1: You can delete this regression test. I forgot we have it. I had some breakthroughs in November for DG fluids that allow us to do the regular sod shock using a characteristic based limiter and I don't see a reason to support two isothermal Euler sod shocks with different left and right states.

JonathanGorard: Roger roger.

JunoRavin:
I don't remember why a lot of the moment regression tests were written this way, but there's a further standard I want to implement that requires a small tedious further refactor:

If there is a field object (gkyl_moment_field), can we define that struct right after the plasma species? Instead of defining the field struct inside the gkyl_moment {} input struct as .field = {}, define a gkyl_moment_field with those exact inputs and then just set .field = field in the gkyl_moment struct.

The reason is that for the other parts of the code we're working with, we will likely move towards a system in which the input field struct has some flexibility. This flexibility may just be an enum stored in the field object (that is the way things currently work) but I prefer the idea that when a field struct can generically refer to different field objects, its declaration be structured to be paired with the plasma species near the top of main like you've so nicely done for the elc and ion initializations (this will help organize input files that "look" like multi-species simulations, but maybe one species is adiabatic and the species evolves implicit through the field equation).

JunoRavin (approved):

PR #317, opened 2024-03-14
Gk g0 app neut

tnbernard:
These changes allow for low Te ionizations. Previously if Eiz/Te < 3/2, the reaction rate was set to zero. Now a finite value is returned and the temperature for the resulting electrons is given by Te = Eiz/6, used in vtSq_iz. (Based on https://link.springer.com/content/pdf/10.1007/978-3-642-71935-6_2.pdf). This model requires that vtSq_iz for the resulting Maxwellian is calculated in the loop over the grid rather than array ops as before. 

This was tested on CPUs and GPUs, including multi-gpus. Unit tests pass. Regression tests run with reasonable results, though high velocity-space resolution is necessary to avoid positivity issues. 

Changes to set path to ADAS data as a CFLAG were also included in these commits.


manauref:
@tnbernard what does "high velocity resolution" mean exactly? can you give some examples, relative to what was used in the past?

tnbernard:
Velocity space resolution for some previous reg tests was (Nvpar, Nmu) = (8,4) and caused them to fail. (16, 8) is more appropriate.

zero/dg_iz.c
```diff
@@ -214,24 +211,29 @@ void gkyl_dg_iz_coll(const struct gkyl_dg_iz *up, const struct gkyl_array *moms_
...
     else {
       double *iz_dat_d = gkyl_array_fetch(up->ioniz_data, gkyl_range_idx(&up->adas_rng, (int[2]) {t_idx,m0_idx}));
       double adas_eval = up->adas_basis.eval_expand(cell_vals_2d, iz_dat_d);
       coef_iz_d[0] = pow(10.0,adas_eval)/cell_av_fac;
+
+      // calculate vtSq_iz at each cell
+      if (up->E/temp_elc_av >= 3./2.) {
+      	// calculate vtSq_iz using current model
+      	// vtSq_iz = vtSq_elc/2.0 - Eiz/(3*me)
+      	array_set1(nc, vtSq_iz_d, 0.5, vtSq_elc_d);
```
manauref, line 226: @tnbernard i would just use `vtSq_iz->ncomp` instead of `nc` here, since this is it's only use and it doesn't cost anything (and you can write the GPU code similarly).

zero/dg_iz_cu.cu
```diff
@@ -15,16 +15,17 @@ extern "C" {
...
 __global__ static void
 gkyl_iz_react_rate_cu_ker(const struct gkyl_dg_iz *up, const struct gkyl_range conf_rng, const struct gkyl_range adas_rng,
   const struct gkyl_basis *adas_basis, const struct gkyl_dg_prim_vars_type *calc_prim_vars_elc_vtSq,
   const struct gkyl_dg_prim_vars_type *calc_prim_vars_donor_gk, const struct gkyl_array* moms_elc,
-  const struct gkyl_array* moms_donor, struct gkyl_array* vtSq_elc, struct gkyl_array* prim_vars_donor,
+  const struct gkyl_array* moms_donor, struct gkyl_array* vtSq_elc, struct gkyl_array* vtSq_iz, struct gkyl_array* prim_vars_donor,
   struct gkyl_array* coef_iz, struct gkyl_array* ioniz_data, int num_basis, enum gkyl_react_self_type type_self,
   double mass_elc, double elem_charge, double E, double maxLogTe, double minLogTe, double dlogTe,
-  double maxLogM0, double minLogM0, double dlogM0, int resTe, int resM0)
+  double maxLogM0, double minLogM0, double dlogM0, int resTe, int resM0, long nc)
```
manauref, line 28: @tnbernard best not to pass nc, you can just get it from `vtSq_iz` inside the kernel (that way your CPU and GPU code will look more similar)

manauref:
@tnbernard another request: have you checked that this works if your input file is not in gkylzero/? I suspect you have to make the corresponding changes to Makefile.sample

tnbernard:
Closing this request to update ionization model to be consistent with https://link.springer.com/content/pdf/10.1007/978-3-642-71935-6_2.pdf. Will be implementing two separate Maxwellians for primary and secondary electrons.

PR #323, opened 2024-03-21
Merge model of low-energy ionizations

tnbernard:
Adding model to ionization to allow for low energy ionizations. When Te/2 > E_iz/3, the phase-space component of the collision term is: 2f_M,iz(n_e, upar_d, vtSq_iz) - f_elc, where vtSq_iz = vtSq_elc/2 - E_iz/3. This gives an energy loss rate proportional to -E_iz when the energy moment is calculated from this term. 

Previously, when Te/2 < E_iz/3, we set the reaction rate to 0 and no ionizations occur in that cell. Now in this case, the phase-space component of the collision term is: 2f_elc - f_M,iz(n_e, upar_d, vtSq_iz), where vtSq_iz = vtSq_elc + 2E_iz/3, which gives the same energy loss rate as higher energy case. 

Since the collision term is calculated at the app level and these logical statements are applied at each cell, two configuration-space fields "fac_felc" and "fac_fmax" are created to multiply f_elc and f_M,iz. For the high Te case, fac_felc = -1 and fac_fmax = 2. Otherwise, fac_felc = 2 and fac_fmax = -1. Then using bin_ops at the app level, the following collision term is accumlated for electrons: 

C_iz = m0_donor * coeff_iz * (fac_felc * f_elc - fac_fmax * f_M,iz).

When the low energy case is turned off, regression tests agree to the previous model within machine precision. The unit test ctest_coll_iz.c pass on CPUs and GPUs and is Valgrind clean. The regression test rt_gk_li_react_3x2v_p1.c runs on CPUs and multi-GPUs and is also Valgrind clean.


JunoRavin (approved):

JunoRavin:
Just adding a comment that I might have missed this change, but I had to copy the *.npy files containing the ADAS data to the installation location (for me $HOME/gkylsoft/gkylzero/share/adas) even though I was just running gkylzero from the repo instead of running gkylzero via the shared library in the installation directory.

tnbernard:
Closing again to add recommendations based on Greg's feedback.

PR #331, opened 2024-03-28
Lbo bug fix

jRoeltgen:
Two fixes for the LBO. First, the cross moments weren't being accounted for when writing out the moments (nu_sum and nu_prim_moms). Second the tpar_min was missing a factor of dvpar.
Plotted here is nu_sum for the "square" species at the final time and the "bump" species at the final and initial times for before any changes, only the change in the write function, and both fixes from the rt_gk_lbo_relax_varnu_1x2v_p1 regression test. The change in tpar_min has no effect on nu as seen by nu_sum being the same for both "only_write_changed" and "vtsq_min_fixed".
The "square" species at t=0 still has nu_sum=0. This looks like it is some subtlety with the write that I haven't figured out as it is nonzero even if only one step is taken.
bump_at_t=0_nusum
bump_at_t=100_nusum
squared_at_t=100_nusum

### DIFF
--- a/apps/gk_species_lbo.c
+++ b/apps/gk_species_lbo.c
@@ -31,7 +31,7 @@ gk_species_lbo_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s, stru
   if (s->info.collisions.normNu) {
     lbo->normNu = true;
     lbo->self_nu_fac = s->info.collisions.self_nu_fac;
-    double tpar_min = (s->info.mass/6.0)*s->grid.dx[cdim];
+    double tpar_min = (s->info.mass/6.0)*s->grid.dx[cdim]*s->grid.dx[cdim];
     double tperp_min = (s->info.collisions.bmag_mid/3.0)*s->grid.dx[cdim+1];
     lbo->vtsq_min = (tpar_min + 2*tperp_min)/(3*s->info.mass);
     double nuFrac = s->info.collisions.nuFrac ? s->info.collisions.nuFrac : 1.0;
--- a/apps/gyrokinetic.c
+++ b/apps/gyrokinetic.c
@@ -782,7 +782,7 @@ gkyl_gyrokinetic_app_write_coll_mom(gkyl_gyrokinetic_app* app, int sidx, double 
   const struct gkyl_array *fin[app->num_species];
   gk_species_lbo_moms(app, gk_s, &gk_s->lbo, gk_s->f);
   if (gk_s->lbo.num_cross_collisions)
-    gk_species_lbo_moms(app, gk_s, &gk_s->lbo, gk_s->f);
+    gk_species_lbo_cross_moms(app, gk_s, &gk_s->lbo, gk_s->f);
 
   // copy data from device to host before writing it out
   if (app->use_gpu) {


apps/gyrokinetic.c
```diff
@@ -782,7 +782,7 @@ gkyl_gyrokinetic_app_write_coll_mom(gkyl_gyrokinetic_app* app, int sidx, double
   const struct gkyl_array *fin[app->num_species];
   gk_species_lbo_moms(app, gk_s, &gk_s->lbo, gk_s->f);
   if (gk_s->lbo.num_cross_collisions)
-    gk_species_lbo_moms(app, gk_s, &gk_s->lbo, gk_s->f);
+    gk_species_lbo_cross_moms(app, gk_s, &gk_s->lbo, gk_s->f);
```
manauref, line 785: @JonathanGorard while you are right correct this, I think we should comment this and the writing of nu_sum and nu_prim_moms out. Those quantities are only typically used for development and testing and we almost never look at them in practice, except for specific rare diagnostics. So I think we should comment out everything in `gkyl_gyrokinetic_app_write_coll_mom` having to do with nu_sum and nu_prim_moms, and we can always uncomment them when we need these detailed diagnostics.

apps/gk_species_lbo.c
```diff
@@ -31,7 +31,7 @@ gk_species_lbo_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s, stru
   if (s->info.collisions.normNu) {
     lbo->normNu = true;
     lbo->self_nu_fac = s->info.collisions.self_nu_fac;
-    double tpar_min = (s->info.mass/6.0)*s->grid.dx[cdim];
+    double tpar_min = (s->info.mass/6.0)*s->grid.dx[cdim]*s->grid.dx[cdim];
```
manauref, line 34: @jRoeltgen I now realize tperp_min needs to be fixed too for the 1x1v case. If vdim=1, line 35 will lead to a memory error. Can you place it in a ternary like
```
tperp_min = vdim>1? (s->info.collisions.bmag_mid/3.0)*s->grid.dx[cdim+1] : tpar_min;
```
? thanks

JonathanGorard:
[image0.gif]

On Mar 27, 2024, at 9:40 PM, Manaure Francisquez ***@***.***> wrote:

﻿

@manauref commented on this pull request.

________________________________

In apps/gyrokinetic.c<https://github.com/ammarhakim/gkylzero/pull/331#discussion_r1542210811>:

> @@ -782,7 +782,7 @@ gkyl_gyrokinetic_app_write_coll_mom(gkyl_gyrokinetic_app* app, int sidx, double
   const struct gkyl_array *fin[app->num_species];
   gk_species_lbo_moms(app, gk_s, &gk_s->lbo, gk_s->f);
   if (gk_s->lbo.num_cross_collisions)
-    gk_species_lbo_moms(app, gk_s, &gk_s->lbo, gk_s->f);
+    gk_species_lbo_cross_moms(app, gk_s, &gk_s->lbo, gk_s->f);


@JonathanGorard<https://github.com/JonathanGorard> while you are right correct this, I think we should comment this and the writing of nu_sum and nu_prim_moms out. Those quantities are only typically used for development and testing and we almost never look at them in practice, except for specific rare diagnostics. So I think we should comment out everything in gkyl_gyrokinetic_app_write_coll_mom having to do with nu_sum and nu_prim_moms, and we can always uncomment them when we need these detailed diagnostics.

apps/gyrokinetic.c
jRoeltgen: So, only prim_moms would be typically output?

apps/gk_species_lbo.c
jRoeltgen: What is policy on vdim=1, then? I was under the impression it was being phased out in gk.

apps/gyrokinetic.c
manauref: yeah, let's only write out prim_moms, since we actually use those to plot upar and T

jRoeltgen: I moved things a bit so that it would be clearer what needs to be done to write out nu_sum and nu_prim_moms.

PR #334, opened 2024-03-31
Ability to specify zero flux BCs on one boundary only and design changes in hyper_dg

manauref:
Previously when we specified zero flux BCs, they had to be used for both boundaries along a direction. This change allows us to specify it on only one side. Made the change in hyper_dg, but only the GK app can use it now.

Also make a couple of style changes in hyper_dg:
- replace hdg with up (like many other updaters).
- Make the _cu methods private and call them if use_gpu is true.

### DIFF
--- a/unit/ctest_dg_rad_gyrokinetic.c
+++ b/unit/ctest_dg_rad_gyrokinetic.c
@@ -673,12 +673,12 @@ void test_1x2v_p2_gpu() { test_1x(2, true, 30.0); }
 TEST_LIST = {
   { "test_1x2v_p1_30eV", test_1x2v_p1_30eV },
   { "test_1x2v_p1_5000eV", test_1x2v_p1_5000eV },
-  { "test_1x2v_p2", test_1x2v_p2 },
+//  { "test_1x2v_p2", test_1x2v_p2 },
   { "test_2x2v_p1", test_2x2v_p1 },
 
 #ifdef GKYL_HAVE_CUDA
   { "test_1x2v_p1_gpu", test_1x2v_p1_gpu },
-  { "test_1x2v_p2_gpu", test_1x2v_p2_gpu },
+//  { "test_1x2v_p2_gpu", test_1x2v_p2_gpu },
 
 #endif
   { NULL, NULL },
--- a/apps/gk_species.c
+++ b/apps/gk_species.c
@@ -111,10 +111,11 @@ gk_species_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app, struct gk_
     s->info.charge, s->info.mass, s->gkmodel_id, app->gk_geom, app->use_gpu);
 
   // by default, we do not have zero-flux boundary conditions in any direction
-  bool is_zero_flux[GKYL_MAX_DIM] = {false};
+  bool is_zero_flux[2*GKYL_MAX_DIM] = {false};
 
-  // determine which directions are not periodic, if any directions are zero-flux, need to set is_zero_flux
-  // keep a copy of num_periodic_dir and periodic_dirs in species so we can
+  // Determine which directions are not periodic. If any BCs are zero-flux,
+  // need to set it in is_zero_flux.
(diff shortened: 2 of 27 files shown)


unit/ctest_dg_rad_gyrokinetic.c
```diff
@@ -673,12 +673,12 @@ void test_1x2v_p2_gpu() { test_1x(2, true, 30.0); }
 TEST_LIST = {
   { "test_1x2v_p1_30eV", test_1x2v_p1_30eV },
   { "test_1x2v_p1_5000eV", test_1x2v_p1_5000eV },
-  { "test_1x2v_p2", test_1x2v_p2 },
+//  { "test_1x2v_p2", test_1x2v_p2 },
```
JunoRavin, line 676: Is there a reason you commented this test out?

JunoRavin:
@manauref can you coordinate with @JonathanGorard on your changes to regression tests? I believe your changes here are going to make @JonathanGorard's more difficult because you have made both style and functional changes to certain regression tests (specifically the 3x2v sheath and cylindrical LAPD). 

There is a style change you've made reducing the space between the -> (so ptr->object instead of ptr -> object). But I don't think we should include such a style change here in this PR when if we want this we should be commenting on @JonathanGorard's regression-refactor branch to request this change (and I was planning on making the same request for this style change so I am in agreement that we should request this change). 

My preference for minimizing pain would be to take the three regression tests you've modified and instead copy @JonathanGorard's current version of them from his branch into here, make the desired functional changes (such as being able to specify the zero flux BC on one side), and then we merge this PR and allow @JonathanGorard to merge the latest gk-g0-app into his branch. 

Especially with how critical the testing is and how much work @JonathanGorard is doing to help us with automated testing, I don't think we should introduce merge conflicts this way requesting particular style and organizational changes.

unit/ctest_dg_rad_gyrokinetic.c
manauref: p2 doesn't work

manauref:
The changes to the lapd_cyl test (adding MPI/NCCL) are absolutely
needed. @JonathanGorard can you let me know if you've done that in your
branch already?

Regarding " -> " vs. "->": I made a comment in an earlier commit that I
think it's unreasonable to go around changing all of these from "->" to "
-> ". I didn't hear opposition to that comment, unless I forgot, so I
changed it back in this one test.

JunoRavin:
The reason I pointed this out is you can see in @JonathanGorard's branch the LAPD cylindrical test was already updated with the NCCL and MPI functionality in his branch: https://github.com/ammarhakim/gkylzero/blob/gk-regression-refactor/regression/rt_gk_lapd_cyl_3x2v_p1.c

But it looked like you didn't copy-paste @JonathanGorard's work so small ordering changes/white space differences are going to create a headache in the merge.

manauref:
If the only differences are in spacing, we can just force @JonathanGorard's
version over this one when his branch is ready to merge. For not this
branch is ready, so we should just push it as is.

JunoRavin:
That's assuming you didn't change anything else besides the new boundary conditions. Can you not just copy the version from @JonathanGorard's branch and modify the BCs as desired so we don't have to double-check that there are no other differences when @JonathanGorard's branch is ready to merge?

ammarhakim:
I agree that we should not make any major spacing changes in this PR merge. I also agree that I do not like extra spaces around -> or other operators. I do not think this helps readability, and in fact, may make it worse. But I am open either way, and we should aim to merge this PR soon.

The more important thing is to get @JonathanGorard to complete his branch and merge. We also need to come up with a style guide. I think @JonathanGorard's code is mostly in compatible style but we need a bit of changes and document them going forward.

JonathanGorard:
Sounds good to me. Happy to proceed without extra spacing around -> operators in future.
Will submit a PR for the gk-regression-refactor branch in the coming day or two, once I've had a chance to test a few things.

PR #335, opened 2024-04-01
Gk g0 app restartfix

akashukla:
Some small changes to make gk restarts work:
1. Add read function to nccl comm object which points to mpi comm read function
2. If on gpu, read to host arrays and copy to gpu
3. Project source in the from file function since sources are all time independent right now.
4. Make from_file and from_frame functions visible to user so we can use them in input files
5. Add restart functionality to 2x2v sheath test

Tested restart on one cpu, one gpu, and 4 gpus


regression/rt_gk_sheath_2x2v_p1.c
```diff
@@ -137,7 +137,7 @@ create_ctx(void)
   double Lv_ion = 8.0 * vti; // Domain size (ion velocity space: parallel velocity direction).
   double Lmu_ion = (3.0 / 2.0) * 0.5 * mass_ion * (4.0 * vti) * (4.0 * vti) / (2.0 * B0); // Domain size (ion velocity space: magnetic moment direction).
   double t_end = 6.0e-6; // Final simulation time.
-  int num_frames = 1; // Number of output frames.
+  int num_frames = 2; // Number of output frames.
```
manauref, line 140: @akashukla did you mean to leave a 2 here?

manauref:
@akashukla sorry to be a pain but I think you should do a complete job here and make these changes to all regregression tests, especially since people are continuously copying tests to create their own input files and we don't want errors to creep in due to various people doing this only when needed.

Also, can you post a plot of a test here please? e.g. run gk_sheath once without a restart, and run it again to the same time with a restart in the middle, and plot (on top of each other for the two runs) the final phi and field_energy. You might want to check this for gk_step also.

akashukla:
image

akashukla:
Restarting from certain frames gives different results

PR #336, opened 2024-04-02
Gk g0 app radiation2

jRoeltgen:
Add radiation to gk_neutrals. Since both gk_species and gk_neut_species can have GKYL_MAX_SPECIES, (most) radiation array maxes are set to 2*GKYL_MAX_SPECIES, since there is little difference in the structure of neutral and ion radiation.
Add a protection to keep the radiation from driving the temperature negative, specifically to not evaluate the radiation when the temperature is below some specified temperature set to 1eV. 1eV is somewhat arbitrary, but few charge states have noticeable radiation below it. This is necessary because when the decrease in emissivity is not resolved due to the smallest resolvable temperature being similar to or greater than the temperature at which the emissivity drops off, the calculated emissivity doesn't decrease and so causes significant radiation at very small temperatures (~1eV) where it shouldn't.


apps/gyrokinetic.c
```diff
@@ -873,27 +878,90 @@ gkyl_gyrokinetic_app_write_rad_emissivity(gkyl_gyrokinetic_app* app, int sidx, d
...
+    if (s->rad.is_neut_species[i]) {
+      int sz_emissivity = gkyl_calc_strlen(fmt_emissivity, app->name, s->info.name, app->neut_species[s->rad.collide_with_idx[i]].info.name, frame);
+      char fileNm_emissivity[sz_emissivity+1]; // ensures no buffer overflow
+      snprintf(fileNm_emissivity, sizeof fileNm_emissivity, fmt_emissivity, app->name, s->info.name, app->neut_species[s->rad.collide_with_idx[i]].info.name, frame);
+      gkyl_comm_array_write(s->comm, &app->grid, &app->local, mt, s->rad.emissivity_host[i], fileNm_emissivity);
+    } else {
+      int sz_emissivity = gkyl_calc_strlen(fmt_emissivity, app->name, s->info.name, app->species[s->rad.collide_with_idx[i]].info.name, frame);
+      char fileNm_emissivity[sz_emissivity+1]; // ensures no buffer overflow
+      snprintf(fileNm_emissivity, sizeof fileNm_emissivity, fmt_emissivity, app->name, s->info.name, app->species[s->rad.collide_with_idx[i]].info.name, frame);
+      gkyl_comm_array_write(s->comm, &app->grid, &app->local, mt, s->rad.emissivity_host[i], fileNm_emissivity);
```
akashukla, line 905: @JunoRavin Should this comm_array_write use the species communicator or app communicator? Are they different?

I had issues reading the files with pgkyl when the species communicator was used here (multi-gpu simulation). I switched this to app->comm and the files seemed to be written correctly.

JunoRavin: If an output is on the configuration space grid, it should be app->comm. If an output is on the phase space grid, it should be species->comm.

apps/gk_species_lbo.c
```diff
@@ -34,6 +34,9 @@ gk_species_lbo_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s, stru
     double tpar_min = (s->info.mass/6.0)*s->grid.dx[cdim]*s->grid.dx[cdim];
     double tperp_min = vdim>1 ? (s->info.collisions.bmag_mid/3.0)*s->grid.dx[cdim+1] : tpar_min;
     lbo->vtsq_min = (tpar_min + 2*tperp_min)/(3*s->info.mass);
+    printf("test tpar min=%e\n",(s->info.mass/6.0)*s->grid.dx[cdim]*s->grid.dx[cdim]/s->info.charge);
```
manauref, line 37: @akashukla @jRoeltgen please remove print statements

apps/gk_species_radiation.c
```diff
@@ -91,6 +99,31 @@ gk_species_radiation_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s
   rad->nvnu = mkarr(app->use_gpu, app->basis.num_basis, s->local_ext.volume);
   rad->nvsqnu_surf = mkarr(app->use_gpu, surf_mu_basis.num_basis, s->local_ext.volume);
   rad->nvsqnu = mkarr(app->use_gpu, app->basis.num_basis, s->local_ext.volume);
+  rad->vtsq = mkarr(app->use_gpu, app->confBasis.num_basis, app->local_ext.volume);
+
+  // Needed arrays for calculating temperature
+  rad->boundary_corrections = mkarr(app->use_gpu, 2*app->confBasis.num_basis, app->local_ext.volume);
+  rad->prim_moms = mkarr(app->use_gpu, 2*app->confBasis.num_basis, app->local_ext.volume);
+
+  rad->vtsq_min = 1.0;
```
manauref, line 108: @akashukla @jRoeltgen please
- place this higher up so it doesn't get lost deep in the code, given that this is an important hardcoded parameter
- Are you sure you wanna set it to 1? 1 what? does it have the correct units?
- Please add a comment of what this is, what it's for, and why you picked this value.

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -152,20 +152,22 @@ struct gk_species_moment {
 struct gk_rad_drag {  
   int num_cross_collisions; // number of species we cross-collide with
   struct gk_species *collide_with[GKYL_MAX_SPECIES]; // pointers to cross-species we collide with
-  int collide_with_idx[GKYL_MAX_SPECIES]; // index of species we collide with
+  struct gk_neut_species *collide_with_neut[GKYL_MAX_SPECIES]; // pointers to neutral cross-species we collide with
+  int collide_with_idx[2*GKYL_MAX_SPECIES]; // index of species we collide with
```
manauref, line 156: @jRoeltgen @akashukla 
why is MAX_SPECIES doubled in various places? If it's not high enough we should just change MAX_SPECIES in gkyl_util.h

apps/gyrokinetic.c
```diff
@@ -840,10 +841,14 @@ gkyl_gyrokinetic_app_write_rad_drag(gkyl_gyrokinetic_app* app, int sidx, double
   snprintf(fileNm_nvsqnu, sizeof fileNm_nvsqnu, fmt_nvsqnu, app->name, gk_s->info.name, frame);
 
   // Compute radiation drag coefficients
+    const struct gkyl_array *fin_neut[app->num_neut_species];
```
manauref, line 844: @jRoeltgen @akashukla fix indent

apps/gyrokinetic.c
```diff
@@ -873,27 +878,90 @@ gkyl_gyrokinetic_app_write_rad_emissivity(gkyl_gyrokinetic_app* app, int sidx, d
...
+      char fileNm_emissivity[sz_emissivity+1]; // ensures no buffer overflow
+      snprintf(fileNm_emissivity, sizeof fileNm_emissivity, fmt_emissivity, app->name, s->info.name, app->species[s->rad.collide_with_idx[i]].info.name, frame);
+      gkyl_comm_array_write(app->comm, &app->grid, &app->local, mt, s->rad.emissivity_host[i], fileNm_emissivity);
+    }  
   }
 
   gyrokinetic_array_meta_release(mt);   
 }
 
+void
```
manauref, line 912: @jRoeltgen @akashukla I think these write functions should go in the gk_species_radiation app. Otherwise we are continuously cluttering the gyrokinetic app with module-specific stuff. I know why you put them here, because there was precedent for it, but I think it was a precedent we should not follow anymore as the system becomes more complex.

apps/gyrokinetic.c
```diff
@@ -873,27 +878,90 @@ gkyl_gyrokinetic_app_write_rad_emissivity(gkyl_gyrokinetic_app* app, int sidx, d
...
+  if (app->use_gpu) {
+    gkyl_cu_memcpy(avals_global, gk_s->rad.red_integ_diag_global, sizeof(double[2+vdim]), GKYL_CU_MEMCPY_D2H);
+  }
+  else {
+    memcpy(avals_global, gk_s->rad.red_integ_diag_global, sizeof(double[2+vdim]));
+  }
+  gkyl_dynvec_append(gk_s->rad.integ_diag, tm, avals_global);
+
+  // Write from rank 0
+  int rank;
```
manauref, line 943: @jRoeltgen @akashukla totally sure, but I think this should be `rank[1]`.

apps/gyrokinetic.c
```diff
@@ -873,27 +878,90 @@ gkyl_gyrokinetic_app_write_rad_emissivity(gkyl_gyrokinetic_app* app, int sidx, d
...
+  gkyl_comm_get_rank(app->comm, &rank);
+  if (rank == 0) {
+    // write out integrated diagnostic moments
+    const char *fmt = "%s-%s_radiation_%s.gkyl";
+    int sz = gkyl_calc_strlen(fmt, app->name, gk_s->info.name,
+      "integrated_moms");
+    char fileNm[sz+1]; // ensures no buffer overflow
+    snprintf(fileNm, sizeof fileNm, fmt, app->name, gk_s->info.name,
```
manauref, line 951: @jRoeltgen @akashukla 
some or part of these string manipulations could happen only once at t=0, so you don't have to do them in time. It'll be better to do that and store it if you move these write functions to gk_species_radiation.

regression/rt_gk_rad_1x2v_p1.c
```diff
@@ -391,7 +391,11 @@ main(int argc, char **argv)
       .num_cross_collisions = 1,
       .collide_with = { "ion" },
     },
-
+    .source = {
```
manauref, line 394: @jRoeltgen @akashukla 
why is this needed?

zero/gkyl_dg_rad_gyrokinetic_drag.h
```diff
@@ -12,6 +12,8 @@ struct gkyl_dg_rad_gyrokinetic_auxfields {
   const struct gkyl_array *nvnu;
   const struct gkyl_array *nvsqnu_surf;
   const struct gkyl_array *nvsqnu;
+  const struct gkyl_array *vtsq;
+  double vtsq_min;
```
manauref, line 16: @jRoeltgen @akashukla 
I suggest you add comments describing what each of these aux fields is, especially since this is a public header so it's supposed to be very clear.

manauref (commented):
@jRoeltgen @akashukla also, can you:
- Rename gkyl_read_radiation to gkyl_radiation_read (modifier always comes after)
- add comments in gkyl_radiation_read to describe each content of `rad_fit_parameters`
please?

apps/gyrokinetic.c
JunoRavin: @manauref I don't think write functions should go in species specific C files at the moment because things like gk_species_radiation are not user facing. In fact, if you wanted to do this it would lead to some annoying convolution of a user-facing function being defined in gkyl_gyrokinetic.h but only being defined in functions which just include gkyl_gyrokinetic_priv.h (which you can legally do because gkyl_gyrokinetic_priv.h includes gkyl_gyrokinetic.h, but does not make the code any easier to parse). 

If we want to move code around, we should add a gkyl_gyrokinetic_write.h and a gyrokinetic_write.c which defines all the user-facing write and read calls. @ammarhakim what do you think about this?

manauref: I just think we need to modularize the whole gyrokinetic app more, and make it so that gyrokinetic calls some write method in the modules, rather than each module dumping its write into gyrokinetic. Same thing that's happening here happened with GkSpecies in g2, leading to a very large GkSpecies, until later I modularized it and moved writes to their corresponding modules. Not only is it cleaner, but it's safer for people to develop new modules.

I see your point though. So I can think of two approaches:
1. Create a gyrokinetic_write as you say, with all the user-facing writes.
2. Remove independent write functions from the input file, simply have a single user-facing gyrokinetic_write function, which calls the (private) writes of the modules. Which of the write calls get called would be controlled by flags in the input file (e.g. we already have .write_source in the source table, we could have similar things in reactions, or tables of diagnostics just like the species diagnostics).

I favor 2

apps/gyrokinetic.c
JunoRavin: So long as you pass rank by reference to mpi_comm_rank, it's fine. You do not need to make this explicitly rank[1] anywhere (and we do not do this in any of the other dynvec write methods). 

The only reason to make it rank[1] is if you want to use the implicit pass by reference in this case (which I don't think we should be doing anyway).

apps/gyrokinetic.c
JunoRavin: I am fine with 2 as well and agree that it would move certain write methods to more natural places. Some care would need to be taken in certain cases (like when we're looping over the diagnostic moment struct that is a struct of structs) but that seems achievable and could even move certain types of moment writes to a specific place (gk_species_moment) that could then be called for other objects (like if radiation or reactions wants to write out one of their moments they computed via gk_species_moment). 

We may want to not overly modularize it. We may want something like gkyl_gyrokinetic_write_species (which checks for a given species what every specific species thing one wants to write out is) and differentiate between distribution function and configuration space quantities. But we can certainly move things around. 

Tagging @ammarhakim and @JonathanGorard in this discussion since this will affect them.

ammarhakim:
I would like to go over this next week. I am not worried about size of the file. That’s irrelevant. More important is to not expose private functions in user facing api. Sent from my iPhoneOn Apr 4, 2024, at 1:59 PM, James Juno ***@***.***> wrote:﻿
@JunoRavin commented on this pull request.



In apps/gyrokinetic.c:
>    }
 
   gyrokinetic_array_meta_release(mt);   
 }
 
+void

I am fine with 2 as well and agree that it would move certain write methods to more natural places. Some care would need to be taken in certain cases (like when we're looping over the diagnostic moment struct that is a struct of structs) but that seems achievable and could even move certain types of moment writes to a specific place (gk_species_moment) that could then be called for other objects (like if radiation or reactions wants to write out one of their moments they computed via gk_species_moment).
We may want to not overly modularize it. We may want something like gkyl_gyrokinetic_write_species (which checks for a given species what every specific species thing one wants to write out is) and differentiate between distribution function and configuration space quantities. But we can certainly move things around.
Tagging @ammarhakim and @JonathanGorard in this discussion since this will affect them.

apps/gkyl_gyrokinetic_priv.h
jRoeltgen: Since both gk_species and gk_neut_species can have GKYL_MAX_SPECIES, (most) radiation array maxes are set to 2*GKYL_MAX_SPECIES, since there is little difference in the structure of neutral and ion radiation. We may want to change the variable name or add a GKYL_MAX_NEUT_SPECIES or something, since currently GKYL_MAX_SPECIES is not actually the max number of species.

regression/rt_gk_rad_1x2v_p1.c
jRoeltgen: The reason for the empty source in the regression test is there was a valgrind error without it. Though I guess this doesn't actually fix the problem, just avoid it. @JunoRavin

JunoRavin: Oh I bet this is an issue that the source struct isn't properly initialized. I'll check.

apps/gyrokinetic.c
jRoeltgen: Other integrated moment functions calculate the string every time step. We're following the same procedure here.

ammarhakim:
Calculating the string repeatedly is fineSent from my iPhoneOn Apr 4, 2024, at 5:06 PM, jRoeltgen ***@***.***> wrote:﻿
@jRoeltgen commented on this pull request.



In apps/gyrokinetic.c:
> +    const char *fmt = "%s-%s_radiation_%s.gkyl";
+    int sz = gkyl_calc_strlen(fmt, app->name, gk_s->info.name,
+      "integrated_moms");
+    char fileNm[sz+1]; // ensures no buffer overflow
+    snprintf(fileNm, sizeof fileNm, fmt, app->name, gk_s->info.name,

Other integrated moment functions calculate the string every time step. We're following the same procedure here.

PR #338, opened 2024-04-02
GK automatic regression system

JonathanGorard:
Principally, these commits introduce both serial and parallel versions of the GK automatic regression system (a GPU version is still in progress). They also implement various aesthetic changes suggested by @manauref for the refactored GK regression tests overall (e.g. removing spaces in between "->" operators, renaming rho_si -> rho_s, etc.), and integrate @manauref's changes to the boundary conditions in the LAPD test in cylindrical coordinates.

None of this functionality is strictly "complete" at the present stage (c.f. the lack of automated GPU testing), but I'm eager to get these changes merged in gk-g0-app as soon as possible, so as to reduce the risk of conflicts moving forward.


JunoRavin:
@JonathanGorard This is fantastic! I have one specific request and then want to rope in @manauref for a more discussion of a second thought.

The specific request: can you please add the gkyl_gyrokinetic_app_write_field_energy to the write_data method? Right now the field energy is being calculated but it's not being written out as far as I can tell. Since the write_field_energy uses dynvectors, it will just do an append to the file (the dynvec_awrite method) if the file already exists.

And if you're adding this, can you also add: 

gkyl_gyrokinetic_app_calc_integrated_mom 

right after you call the calc_field_energy method, and correspondingly add

gkyl_gyrokinetic_app_write_integrated_mom 

after you call the write_field_energy method? It would be awesome if your CI system can also be comparing the integrated diagnostics as well through the output files. If this gives you any trouble we don't have to add this right away (if for whatever reason your CI system right now doesn't like our dynvec data) I just wanted to mention this now as something on our radar to add. 

The second thing which is more of a discussion and it would be useful to get @manauref's perspective: what are people's opinion on vpar_max and mu_max instead of Lvpar and Lmu? The reason I mention this is because while what has been done is imminently reasonable, especially for the vparallel extents I have never thought about velocity space extents in terms of a length (so -0.5*Lvpar to 0.5*Lvpar), I have always thought about them as a distance from 0 (so -vpar_max to vpar_max). Part of this is because sometimes velocity space extents are asymmetric or centered around a bulk velocity, so I'm always defining the velocity space extents _with respect to something_, instead of an absolute distance/length. 

Other than that I think this is ready to go, and we could even punt on both of these comments to get this merged sooner rather than later.

ammarhakim:
I have several comments and suggestions, but the details can wait till later. In general, this is a good start.  Where are the new executables being put? I do not see them anywhere. How do I build them? Can you please add instructions to the README page in a new section?

Also, I would the C code to become more generic and work on any set of regression tests, not just GK. For this, we can read the config info from a toml file. There is an excellent pure-C toml single-file parser. That config can tell us what tests to run and other config info like tolerances etc. We can imagine that this will then be extensible to VM and moments App too.

I can outline the refactors needed to get there next week. But I think once these comments are addressed we can merge the PR.

JonathanGorard:
Alright, starting with @ammarhakim's comments: would you prefer for me to add instructions for how to compile the auto regression code to the README, or for me to modify the makefile so that it compiles automatically? Or both?

Also, I agree that there may potentially be merit to parsing both test and config data from a TOML file rather than hardcoding it in C. However, I don't think that extensibility to the VM and moments apps should be the reason for doing it. From my perspective, although there are certain code similarities, the way that we handle automatic regression testing truly should be different in these cases (certainly between GK and moments, though there may be a case to be made that GK and VM should ultimately be treated the same). For instance, the metrics that we output from the simulations are different, and although those could in principle all be placed into a TOML file, the way that we perform _comparisons_ on those metrics, and interpret the results, is also different (e.g. if the species RHS time is higher on one test run then we should treat that as a regression, but not if it's lower; however, if the number of RK-2 failures is higher _or_ lower on one test run then we should treat that as a regression). Moreover, the approach to parallelization is fundamentally different (e.g. parallelizing only in Z vs. parallelizing over all coordinate directions), and a couple of other things. Those differences are already reflected in the differences between the GK regression code in the branch, and the moment app regression code in regression-refactor. If we attempted to build all of those distinctions into a unified TOML format to be parsed by a single regression system framework, I personally believe that we would be at risk of over-engineering this. I'm open to be contradicted - this is simply my perspective.

JonathanGorard:
As for @JunoRavin's comments: yes, absolutely, I will incorporate the appropriate gkyl_gyrokinetic_app_write_field_energy, gkyl_gyrokinetic_app_calc_integrated_mom and gkyl_gyrokinetic_app_write_integrated_mom calls to the various write_data methods. And yes, I'll be sure to add the integrated diagnostics output(s) to the list of things that should be automatically compared - I considered doing this by default, but as a general rule I wanted to be somewhat conservative with what exactly we stored and compared with each run, since those comparisons are quite costly (simply in terms of computation time), sometimes being comparable or even exceeding the computation time of the test itself. But the integrated diagnostic moments have quite a small footprint compared to the overall output file, so I don't see this being a problem.

Yes, the GK system did use vpar_max and mu_max initially, and I only modified that for the sake of consistency of our conventions across both configuration and velocity space (i.e. everywhere else in the code, we seem to use Lx, Lr, Ltheta, etc. for denoting configuration space extents). But I don't have any strong opinion on this, and it sounds as though you feel that velocity space is sufficiently different that it does make sense to have a genuinely different convention for it. If @manauref (and possibly @ammarhakim) agree, then I'm quite happy to revert to that.

ammarhakim:
We need to add a make target to build the execs. I am not sure how to build or use at present. Some instructions would be helpful. Sent from my iPhoneOn Apr 3, 2024, at 10:11 AM, JonathanGorard ***@***.***> wrote:﻿
As for @JunoRavin's comments: yes, absolutely, I will incorporate the appropriate gkyl_gyrokinetic_app_write_field_energy, gkyl_gyrokinetic_app_calc_integrated_mom and gkyl_gyrokinetic_app_write_integrated_mom calls to the various write_data methods. And yes, I'll be sure to add the integrated diagnostics output(s) to the list of things that should be automatically compared - I considered doing this by default, but as a general rule I wanted to be somewhat conservative with what exactly we stored and compared with each run, since those comparisons are quite costly (simply in terms of computation time), sometimes being comparable or even exceeding the computation time of the test itself. But the integrated diagnostic moments have quite a small footprint compared to the overall output file, so I don't see this being a problem.
Yes, the GK system did use vpar_max and mu_max initially, and I only modified that for the sake of consistency of our conventions across both configuration and velocity space (i.e. everywhere else in the code, we seem to use Lx, Lr, Ltheta, etc. for denoting configuration space extents). But I don't have any strong opinion on this, and it sounds as though you feel that velocity space is sufficiently different that it does make sense to have a genuinely different convention for it. If @manauref (and possibly @ammarhakim) agree, then I'm quite happy to revert to that.

JonathanGorard:
@ammarhakim and @JunoRavin hopefully those changes should address your immediate concerns/suggestions (though please let me know if not).

manauref:
catching up on this conversation. There are 2 minor changes I'd definitely like to make:
1. Yes, I think we should change Lvpar and Lmu in favor or vpar_max and mu_max. The latter is more intuitive, and as @JunoRavin there may be some niche (in GK at least) sims where we use domains not centered at vpar=0.
2. There are some comments in the reg tests I think should be changed (e.g. the description of the Coulomb logarithm). I can do this.

There's one change I'd like to suggest, not strictly require:
Presently we have variables like Nx, Ny, Nz and NX, NY, NZ. Later someone has to put some subset of these in the ccells array depending on dimensionality, and we also have to hardcode cdim and vdim in the creation of the app. I think it's possible to automate these things, so that one doesn't have to make multiple of these changes when changing dimensionality. For example, I propose that instead of Nx,Ny,Nz we simply have Nx[ctx.cdim] and NX[ctx.cdim], and similarly for Nvpar and Nmu. This way we can make things in the app creation more dimensionally independent, e.g. instead of 
```
  int NX = APP_ARGS_CHOOSE(app_args.xcells[0], ctx.Nx);
  int NY = APP_ARGS_CHOOSE(app_args.xcells[1], ctx.Ny);
  int NZ = APP_ARGS_CHOOSE(app_args.xcells[2], ctx.Nz);
  int NV = APP_ARGS_CHOOSE(app_args.vcells[0], ctx.Nv);
  int NMU = APP_ARGS_CHOOSE(app_args.vcells[1], ctx.Nmu);
```
we would have
```
  int NX[cdim], NV[vdim];
  for (int d=0; d<cdim; ++d)
    NX[d] = APP_ARGS_CHOOSE(app_args.xcells[d], ctx.Nx[d]);
  for (int d=0; d<vdim; ++d)
    NV[d] = APP_ARGS_CHOOSE(app_args.vcells[d], ctx.Nv[d]);
```
This would would for all dimensions, and would remove a number of hardcoded things in the app creation. Again, the intention is to change the number of things that someone needs to change when creating a new input file/reg test.

manauref:
Another thing I'd like to add, since we are trying to standardize and set templates here, is a step in/around the while loop that terminates the simulation when the time step is too small (say 1e4 or 1e5 smaller than the initial time step). I can give an example in a bit to see if you approve. This is useful/important in production sims, otherwise people burn computer time unnecessarily.

ammarhakim:
I think the execs should be built with "make regression". There should be no need to type the long make commands. This should be easy to do in the Makefile.

I think it is best to replace our json output with toml output. We need not append to existing stat file and just overwrite them. However, this not needed for PR merge.

JonathanGorard:
@ammarhakim the current makefile supports running "make ci" which builds all automatic regression executables (I don't consider it a good idea to have the CI system built with "make regression", since running the CI executables builds each regression test individually anyway, and I'd prefer not to duplicate effort on the compiler's end unnecessarily). I realize that I did not add that syntax to the README, but I can easily do so. Is that sufficient?

@manauref since both you and @JunoRavin support using different conventions for configuration and velocity space, I shall make that change today. Also, I like the idea of auto-terminating the simulation if the time-step remains too low for too long. I'll make the appropriate modifications to the GK simulations today, with a note that we should apply that change to the regression system more generally. Thanks!

PR #339, opened 2024-04-03
Corner syncs and mpi_comm refactor

ammarhakim:
Completed the corner syncs. This required a major refactor of sync code (periodic and non-periodic). I have not made the corresponding changes for nccl comm but these also need to be made.

### DIFF
--- a/regression/rt_10m_lhdi_grad_closure.c
+++ b/regression/rt_10m_lhdi_grad_closure.c
@@ -529,7 +529,6 @@ main(int argc, char **argv)
     step += 1;
   }
 
-  gkyl_moment_app_write(app, t_curr, 1000);
   write_data(&io_trig, app, t_curr);
   gkyl_moment_app_stat_write(app);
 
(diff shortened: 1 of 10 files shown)


regression/rt_10m_lhdi_grad_closure.c
```diff
@@ -526,6 +529,7 @@ main(int argc, char **argv)
     step += 1;
   }
 
+  gkyl_moment_app_write(app, t_curr, 1000);
```
JunoRavin, line 532: I think this line should be deleted. You're calling write_data immediately after this.

zero/gkyl_range.h
```diff
@@ -428,6 +434,19 @@ bool gkyl_range_is_on_lower_edge(int dir, const struct gkyl_range *range,
 bool gkyl_range_is_on_upper_edge(int dir, const struct gkyl_range *range,
   const struct gkyl_range *parent);
 
+/**
+ * Check if @a targ range shares an edge with the @a base range. The
```
JunoRavin, line 438: Is there a reason this comment does @a for the definitive article before targ and base?

ammarhakim: @a is the syntax for doxygen to say the thing that follows it is a parameter

PR #342, opened 2024-04-08
More improvements/standardization of GK reg tests

manauref:
- Correct the way we were checking for small dt
- Add small dt check to reg tests that didn't have it.
- Modify write_data so it allows being forced to write (when dt crashed).
- Add standard write_data to reg tests that didn't have it.
- Create a separate function to calculate integrated diagnostics (field_energy and int moms for now).
- Use a separate trigger to calculate int diags at a different rate (one can set the number of calcs to INT_MAX to calculate in every time step, see comment in code).


JonathanGorard:
Generally I think this looks very good! The only immediate question I think I have is whether we really want to call the trigger for calculating integrated diagnostics `io_trig_int_diag`, given that there isn't (or at least shouldn't be...) any "IO" happening here? Could something like `calc_trig_int_diag` be more descriptive?

manauref:
sure, we can do that

Issue #265, opened 2024-02-05
Implement T_parallel and T_perp into prim_moms


Maxwell-Rosen:
I'm working on branch gk-g0-app and it would be tremendously helpful if the diagnostics for T_parallel and T_perp were included into the prim_moms output.

manauref:
@JunoRavin you might be better suited to address this, because I have no clue how diagnostics work in the g0 app

JunoRavin:
This functionality is now implemented as of https://github.com/ammarhakim/gkylzero/pull/350

One can request either the "MaxwellianMoments" which return (n, u_par, T/m)
or one can request the "BiMaxwellianMoments" which return (n, u_par, T_par/m, T_perp/m)

in the same way one requests "M0, M1, M2,..."

PR #350, opened 2024-04-16
Refactor the Maxwellian correction routine in gyrokinetics based on successful refactoring in Vlasov

JunoRavin:
This branch takes the work of the Vlasov-Maxwell refactor of the Maxwellian correction routine (the LTE correction routine for Vlasov since Vlasov is agnostic to non-relativistic vs. relativistic) and hooks the results in to the gyrokinetic app. The big picture changes are as follows:

1. The correction algorithm converges in the cell averages of (n, u_par, T/m) and **does not** try to converge in all the moments (i.e., does not try to converge point-wise). This is a change both in iterating on the primitive moments and only using the cell averages. 
2. We utilize a relative error instead of an absolute error, which allows us to reduce finite precision errors when running with SI units. Note that for the u_par convergence we check if fabs(u_par) < 1.0 (i.e., if u_par is O(1)) because we can safely use absolute errors in that case and avoid dividing by 0.0 (since some cells will have very small u_par while other cells may have large flows, like near the sheath).
3. The option to utilize the last iteration if one wants to utilize a small fixed number of iterations (like 2-5 iterations) so that even if you "fail to converge" you may still get some benefit from the correction routine. 

The details of these big picture changes are as follows:

1. The creation of a new updater: gyrokinetic_maxwellian_moments, which computes (n, u_par, T/m) with the option to return n *with or without* the Jacobian factor, depending on how one is projecting the Maxwellian (either just using the velocity space Jacobian, or the total Jacobian).
2. This new updater also addresses issue https://github.com/ammarhakim/gkylzero/issues/265 by giving an option to compute the bi-Maxwellian moments (n, u_par, T_par/m, T_perp/m). 
3. The deletion of the old correction routine and the addition of the updater gyrokinetic_maxwellian_correct, which has been hooked in to both the app and unit test
4. We can write out the correction status as an optional diagnostic, which for GK returns a dynvector of size 5 (the number of iterations, the status (whether or not it converged, 0 means converged), the maximum error in n, the maximum error in u_par, and the maximum error in T/m)
5. A small change to proj_gkmaxwellian_prim_moms to take a *single* array of (n, u_par, T/m) (@manauref want to draw your attention to this change in the hopes that conflicts with the non-uniform velocity-space work are minimized). 
6. Clean up of the app to utilize this new infrastructure, including a big of streamlining of the logic and the option to "correct_all_moms" in both the projection routine and the BGK operator
7. Hooking in the Vlasov correction routine and projection objects to the neutral species. 
8. A small change to gk_species_react (@tnbernard @akashukla) to accommodate the fact that gkmaxwellian_prim_moms expects a single array now (we might want to slightly refactor this change since it's a little convoluted). 

We have a new regression test: gk_periodic_bgk_sod_shock_p1, which does a 1x sod shock set up in the gyrokinetic app with no evolution of the potential, so similar to a 1x3v Vlasov sod shock. We can converge the primitive moments in 5-8 iterations to 1.0e-14 error and thus obtain perfect energy conservation (if you do not correct the moments, the error in the conservation is 1-10 percent). 

If you want to correct initial conditions or sources, you can now do so with the parameter "correct_all_moms" in the projection object. Give it a shot! So far my experience is universally positive. 

Next steps:
1. This PR comments out the cross-species collisions in BGK because the BGK cross species collisional moments should now be the *primitive moments* u_{rs} T_{rs}. We need to generate new kernels and deploy these new kernels.


ammarhakim:
My goodness. A War and Peace length PR comment!

apps/gk_species_projection.c
```diff
@@ -44,53 +35,37 @@ gk_species_projection_init(struct gkyl_gyrokinetic_app *app, struct gk_species *
...
+      .use_last_converged = false, // do not use the unconverged moments if the scheme fails to converge
+      .mass = s->info.mass,
+      .use_gpu = app->use_gpu,
+      .max_iter = 50,
+      .eps = 1e-10,
     };
-    proj->corr_max_lab = gkyl_correct_maxwellian_gyrokinetic_new(&inp);  
-    proj->proj_max_lab = gkyl_proj_maxwellian_on_basis_new(&s->grid, &app->confBasis, &app->basis, 
-      app->basis.poly_order+1, app->use_gpu);
+    proj->corr_max = gkyl_gyrokinetic_maxwellian_correct_inew( &inp_corr );
```
manauref, line 54: @JunoRavin I don't think you released this object

apps/gk_species_react.c
```diff
@@ -166,9 +168,10 @@ gk_species_react_rhs(gkyl_gyrokinetic_app *app, const struct gk_species *s,
 
     if (react->react_id[i] == GKYL_REACT_IZ) {
       if (react->type_self[i] == GKYL_SELF_ELC) {
-        gkyl_array_set_offset(react->prim_vars[i], 1.0, react->vt_sq_iz[i], 1*app->confBasis.num_basis);
+        gkyl_array_set_offset(react->prim_vars_proj_inp[i], 1.0, react->moms_elc[i].marr, 0*app->confBasis.num_basis);
+        gkyl_array_set_offset(react->prim_vars_proj_inp[i], 1.0, react->vt_sq_iz[i], 2*app->confBasis.num_basis);
```
manauref, line 172: @JunoRavin can you double check that `prim_vars_proj_inp` is the right size and that it's used correctly here?
Am i correct in interpreting this as:
- L171 sets `prim_vars_proj_inp = {M0,M1,M2}`
- L172 rewrites the last entry, so that `prim_vars_proj_inp = {M0,M1,vt_Sq_iz}`?

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -339,25 +365,32 @@ struct gk_proj {
...
-      struct gkyl_array *prim_moms_dev; // device-side prim_moms for GPU simulations
-      struct gkyl_array *dens_dev; // device-side density for GPU simulations
-      struct gkyl_dg_bin_op_mem *mem; // memory needed in correcting density
+      bool correct_all_moms; // boolean if we are correcting all the moments
 
       struct gkyl_proj_on_basis *proj_dens; // projection operator for density
       struct gkyl_proj_on_basis *proj_upar; // projection operator for upar
-      struct gkyl_proj_on_basis *proj_udrift; // projection operator for upar
       
       union {
```
manauref, line 377: @JunoRavin @ammarhakim these multi-nested structs and unions are hard to get used to, though I could. Is this the best way to do it?

apps/gyrokinetic.c
```diff
@@ -1132,10 +1132,12 @@ gkyl_gyrokinetic_app_write_mom(gkyl_gyrokinetic_app* app, double tm, int frame)
       snprintf(fileNm, sizeof fileNm, fmt, app->name, app->species[i].info.name,
         app->species[i].info.diag_moments[m], frame);
 
-      // Rescale moment by inverse of Jacobian
-      gkyl_dg_div_op_range(app->species[i].moms[m].mem_geo, app->confBasis, 
-        0, app->species[i].moms[m].marr, 0, app->species[i].moms[m].marr, 0, 
-        app->gk_geom->jacobgeo, &app->local);      
+      // Rescale moment by inverse of Jacobian if not already re-scaled 
+      if (!app->species[i].moms[m].is_bi_maxwellian_moms && !app->species[i].moms[m].is_maxwellian_moms) {
```
manauref, line 1136: @JunoRavin same here, I would rename `bi_maxwellian` to `bimaxwellian`.

Also, should we just make the regular moments always be rescaled by 1/jacobgeo? They are used for LBO prim moms and for reactions. For LBO I think it doesn't matter whether we use J*moms or just moms. For reactions, since we use them to compute cross sections and Maxwellians, can't we just do all that with moms (instead of J*moms)?

zero/gkyl_gyrokinetic_maxwellian_correct.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
```
manauref, line 1: Ahh, I see you renamed it do put the gyrokinetic as a prefix. Why? Maybe we should do that for all objects?

zero/gkyl_gyrokinetic_maxwellian_correct.h
```diff
@@ -0,0 +1,93 @@
...
+struct gkyl_gyrokinetic_maxwellian_correct_status 
+gkyl_gyrokinetic_maxwellian_correct_all_moments(gkyl_gyrokinetic_maxwellian_correct *up,
+  struct gkyl_array *f_max, const struct gkyl_array *moms_target, 
+  const struct gkyl_range *phase_local, const struct gkyl_range *conf_local);
+
+/**
+ * Host-side wrapper for computing the absolute value of the 
+ * difference in cell averages between the target moments and iterative moments.
+ */
+void gkyl_gyrokinetic_maxwellian_correct_all_moments_abs_diff_cu(const struct gkyl_range *conf_range,
```
manauref, line 84: @JunoRavin can you put this declaration in the private header? also, please describe its arguments.

zero/gyrokinetic_maxwellian_correct.c
```diff
@@ -0,0 +1,298 @@
...
+        // Iterate over the input configuration-space range to find the maximum error
+        gkyl_range_iter_init(&biter, conf_range);
+        while (gkyl_range_iter_next(&biter)){
+          long midx = gkyl_range_idx(conf_range, biter.idx);
+          const double *moms_local = gkyl_array_cfetch(up->moms_iter, midx);
+          const double *moms_target_local = gkyl_array_cfetch(moms_target, midx);
+          // Check the error in the absolute value of the cell average
+          // Note: for density and temperature, this error is a relative error compared to the target moment value
+          // so that we can converge to the correct target moments in SI units and minimize finite precision issues.
+          up->error[0] = fmax(fabs(moms_local[0*nc] - moms_target_local[0*nc])/moms_target_local[0*nc],fabs(up->error[0]));
```
manauref, line 169: I'm not sure when we should use `fmax` or `GKYL_MAX2`. Let me know if you know

zero/gyrokinetic_maxwellian_correct.c
```diff
@@ -0,0 +1,298 @@
...
+          // However, u_par may be ~ 0 and if it is, we need to use absolute error. We can converge safely using
+          // absolute error if u_par ~ O(1). Otherwise, we use relative error for u_par. 
+          if (fabs(moms_target_local[1*nc]) < 1.0) {
+            up->error[1] = fmax(fabs(moms_local[1*nc] - moms_target_local[1*nc]),fabs(up->error[1]));
+          }
+          else {
+            up->error[1] = fmax(fabs(moms_local[1*nc] - moms_target_local[1*nc])/moms_target_local[1*nc],fabs(up->error[1]));
+          }
+          // Check if density and temperature are positive, if they aren't we will break out of the iteration
+          ispositive_flte = (moms_local[0*nc]>0.0) && ispositive_flte;
```
manauref, line 180: @JunoRavin what does `_flte` stand for or mean?

zero/gyrokinetic_maxwellian_correct.c
```diff
@@ -0,0 +1,298 @@
...
+
+void 
+gkyl_gyrokinetic_maxwellian_correct_all_moments_abs_diff_cu(const struct gkyl_range *conf_range, 
+  int nc, const struct gkyl_array *moms_target, const struct gkyl_array *moms_iter, 
+  struct gkyl_array *moms_abs_diff)
+{
+  assert(false);
+}
+
+#endif
```
manauref, line 298: @JunoRavin if you move the declaration from the public to the private header I mentioned above, you can just delete this. We should do this in future code and maybe go back and change this in old code because it's cleaner.

zero/gyrokinetic_maxwellian_moments.c
```diff
@@ -0,0 +1,233 @@
...
+  else {
+    up->vdim = 3;
+  }
+  up->gk_geom = gkyl_gk_geometry_acquire(inp->gk_geom);
+  up->divide_jacobgeo = inp->divide_jacobgeo;
+
+  long conf_range_ncells = inp->conf_range->volume;
+  long conf_range_ext_ncells = inp->conf_range_ext->volume;
+
+  if (inp->use_gpu) {
```
manauref, line 35: @JunoRavin no strong preference on this one, but should we tell `gyrokinetic_maxwellian_moments` what kind of moments we are going to request, so that we only allocate the temporary arrays we need, instead of all of these?
Maybe not since it'll be kind a pain to go through and change it now, and we are not starving for memory AFAICT

zero/gyrokinetic_maxwellian_moments.c
```diff
@@ -0,0 +1,233 @@
...
+    // Bin op memory for needed weak divisions
+    up->mem = gkyl_dg_bin_op_mem_new(conf_range_ncells, up->num_conf_basis);
+    if (up->vdim == 3) {
+      // Additional moments if computing Bi-Maxwellian moments (Tpar, Tperp)
+      up->p_par = gkyl_array_new(GKYL_DOUBLE, up->num_conf_basis, conf_range_ext_ncells);
+      up->t_par = gkyl_array_new(GKYL_DOUBLE, up->num_conf_basis, conf_range_ext_ncells);
+      up->p_perp = gkyl_array_new(GKYL_DOUBLE, up->num_conf_basis, conf_range_ext_ncells);
+      up->t_perp = gkyl_array_new(GKYL_DOUBLE, up->num_conf_basis, conf_range_ext_ncells);
+    }
+  }
```
manauref, line 68: @JunoRavin  L35-L68: can you create a local mkarr function like the one in the apps, and consolidate these two blocks into a single one please?

manauref (changes requested):
Thanks for making these changes. I'm overall approving, and I think this will greatly improve simulations with Maxwellian projections/BGK (cost TBD).
Aside from the relatively minor things I brought up in individual comments I have 1 request, 1 suggestion, and 3 questions:
1. Request: did you run the relevant unit and reg tests and make sure they work and that they are valgrind free? i'm thinking the unit tests modified here + bgk,step,react reg tests. Especially since I think I spotted at least one memory leak.
2. Suggestion: that we implement the cross-species BGK moments here so we can wrap the BGK stuff up.
3. Question 1: In your PR message you said this addresses issue #265. Does allow me to request prim_moms with Tpar/Tperp from the input file? how?
4. Question 2: Can you clarify, maybe by editing the PR message, what the default setting is now for BGK collision and reaction operators, with regards to corrections? are they do it? with how many iterations, etc.
5. Question 3: Is it clear to you if the maxwellian_correction routine could be easily written in a way that also works for bimaxwellians?

JunoRavin:
> Thanks for making these changes. I'm overall approving, and I think this will greatly improve simulations with Maxwellian projections/BGK (cost TBD). Aside from the relatively minor things I brought up in individual comments I have 1 request, 1 suggestion, and 3 questions:
> 
> 1. Request: did you run the relevant unit and reg tests and make sure they work and that they are valgrind free? i'm thinking the unit tests modified here + bgk,step,react reg tests. Especially since I think I spotted at least one memory leak.
> 2. Suggestion: that we implement the cross-species BGK moments here so we can wrap the BGK stuff up.
> 3. Question 1: In your PR message you said this addresses issue [Implement T_parallel and T_perp into prim_moms #265](https://github.com/ammarhakim/gkylzero/issues/265). Does allow me to request prim_moms with Tpar/Tperp from the input file? how?
> 4. Question 2: Can you clarify, maybe by editing the PR message, what the default setting is now for BGK collision and reaction operators, with regards to corrections? are they do it? with how many iterations, etc.
> 5. Question 3: Is it clear to you if the maxwellian_correction routine could be easily written in a way that also works for bimaxwellians?

1. Good catch on things like the memory leak and the allreduce_host. Yes this PR has not been extensively vetted yet by the regression system because I'm traveling but I wanted to file it and get your comments now before the next week and a half is busy with writing
2. We can push ahead with the cross-species BGK refactor, but since these changes were self-contained and @dingyunl is still working on implementing the refactored BGK cross moments, I wanted to file the PR and get people's eyes on this now. There is other clean up of this work I'd like to do but did not do to minimize the overall number of changes for this refactor (big example: in Vlasov the density correction is just part of the projection routine as opposed to a separate method in the correction updater, but I avoided doing this because I didn't want to make a bunch of changes to the project Maxwellian routines and create conflicts with the non-uniform velocity space work. Another example: the reaction routines could switch to utilizing the Vlasov LTE projection object which does give density correction for free and hooks in more seamlessly to the correction routine).
3. There are two new diagnostic moment options: "MaxwellianMoments" and "BiMaxwellianMoments" which return (n, u_par, T/m) and (n, u_par, T_par/m, T_perp/m). These moments are returned this way because of how they are utilized by the correction updater and projection updaters. We could add options to unpack some of them but my chief ambition was to finally be able to write out T_par and T_perp and get rid of the need to set write_collisions=true for writing out the primitive moments. **If you have a preferred name change please let me know. I am happy to change names.**
4. **No solver or updater turns on corrections by default**. You must set the "correct_all_moms=true" flag. 
5. Yes this is as close to trivial as I am willing to call things trivial. It literally could be an option to the correction updater as is (set the number of corrected components to either 3 or 4 depending on whether you're doing a Maxwellian or Bimaxwellian). 

With respect to names. I made the deliberate choice to rename things as gyrokinetic_* so that it was more clear that certain objects went together (gyrokinetic_maxwellian_moments and gyrokinetic_maxwellian_correct). This naming convention I have adopted for Vlasov (see all the vlasov_lte functions) and if anything I think in the refactor of proj_maxwellian, since we should not need the Vlasov version anymore, we can repurpose this object as gyrokinetic_maxwellian_proj_on_basis.

dingyunl:
Thanks Jimmy for revising the maxwellian correction routine. One thing I notice is that in this new routine, the density is also corrected by iteration instead of rescaling, which saves some lines of code but seems to increase the dimension of the iteration. Do we want to keep the rescaling for density correction?

apps/gk_species_react.c
JunoRavin: Good catch. I should only be using m0_elc in that first call so that the final array is:

{M0, 0, vt_sq_iz}

to be consistent with what we were doing before

PR #352, opened 2024-04-19
Canonical Poisson Bracket

johnson452:
Adds the Canonical PB infrastructure for 1x1v, 2x2v (p1 + p2), and 3x3v (p1) to G0. GPU enabled.

Important Notes:
- The infrastructure requires a hamiltonian written in canonical-coordinates. 
- There is no check to ensure the user has handed a hamiltonian in canonical coordinates.
- Results are returned in the computational/canonical coordinates (q,p). 
- This PR also modified zero/gkyl_dflt.h to fix issues building with GPU on power9 systems like traverse.

Future work will require:
- Addition of a metric function pointer (similar to hamil). (Currently this is not required without BGK collisions/specific initial conditions with a maximum entropy state.)
- Extending the LTE moments to properly compute equilibria based on the metric.

Two regression tests (for 1x1v and 2x2v) have been added as examples.
- The 1x1v test runs a neutral bgk shock test with free-streaming particles (rt_neut_can_pb_bgk_sod_shock_1x1v_p2.c)
- The 2x2v test runs a small neutral free-streaming test on the surface of a sphere. (rt_can_pb_free_streaming_surf_sphere.c)


JunoRavin:
Also please fix the comments in zero/gkyl_dg_canonical_pb.h to reflect the fact that you're doing a Canonical Poisson Bracket update, not relativistic Vlasov.

PR #353, opened 2024-04-22
Gk g0 app norm nu refactor

akashukla:
Now bmag at the center of the domain is calculated internally so it does not need to be passed to the field or the collisions in the input file. As part of this, a host array broadcast function has been added to the nccl comunicator. nuFrac is now used properly by normNu (previously the  user had to factor it into their calculation of the normNu factors).


regression/rt_gk_lbo_relax_varnu_1x2v_p1.c
```diff
@@ -378,13 +377,14 @@ main(int argc, char **argv)
...
       .collision_id = GKYL_LBO_COLLISIONS,
       .normNu = true,
-      .bmag_mid = ctx.B0,
-      .self_nu_fac = ctx.self_nu_fac,
-      .cross_nu_fac = ctx.cross_nu_fac_tophat,
+      .n_ref = ctx.n0,
+      .T_ref = ctx.vt*ctx.vt*ctx.mass,
+      .hbar = 1.0,
+      .eps0 = 1.0,
+      .eV = 1.0,
```
manauref, line 384: @akashukla Since reg tests serve as an example to others, please comment on each of these lines describing their meaning, and why or why not they need to be specified.

regression/rt_gk_mdpx_cart_3x2v_p1.c
```diff
@@ -434,11 +434,8 @@ main(int argc, char **argv)
     .collisions =  {
       .collision_id = GKYL_LBO_COLLISIONS,
       .normNu = true,
-      .self_nu_fac = norm_nu_func(ctx.nuFrac, ctx.n0, ctx.n0, ctx.massElc, ctx.massElc, ctx.chargeElc, ctx.chargeElc, ctx.Te, ctx.Te),
-      .cross_nu_fac = {
-        norm_nu_func(ctx.nuFrac, ctx.n0, ctx.n0, ctx.massElc, ctx.massAr, ctx.chargeElc, ctx.chargeIon, ctx.Te, ctx.TAr)
-      },
-      .bmag_mid = ctx.B0,
+      .n_ref = ctx.n0,
+      .T_ref = ctx.Te,
```
manauref, line 438: @akashukla likewise, since these are examples for others, please comment on these lines and explain what n_ref and T_ref are.

regression/rt_gk_step_out_2x2v_p1.c
```diff
@@ -611,12 +608,9 @@ main(int argc, char **argv)
...
       .normNu = true,
-      .self_nu_fac = norm_nu_func(ctx.nuFrac, ctx.n0, ctx.n0, ctx.massIon, ctx.massIon, ctx.chargeIon, ctx.chargeIon, ctx.Ti, ctx.Ti),
-      .cross_nu_fac = {
-        norm_nu_func(ctx.nuFrac, ctx.n0, ctx.n0, ctx.massIon, ctx.massElc, ctx.chargeIon, ctx.chargeElc, ctx.Ti, ctx.Te), 
-        norm_nu_func(ctx.nuFrac, ctx.n0, ctx.n0Ar, ctx.massIon, ctx.massAr, ctx.chargeIon, ctx.chargeIon, ctx.Ti, ctx.TAr)
-      },
-      .bmag_mid = 2.51,
+      .nuFrac = ctx.nuFrac,
+      .n_ref = ctx.n0,
+      .T_ref = ctx.Ti,
```
manauref, line 613: same here

zero/gk_geometry.c
```diff
@@ -108,6 +108,30 @@ void gkyl_gk_geometry_augment_local(const struct gkyl_range *inrange,
...
+  int cdim = up->grid.ndim;
+  int idx_mid[cdim];
+  double xc[cdim];
+  for(int i = 0; i <cdim; i++) {
+    if (up->grid.cells[i]%2 == 0) {
+      idx_mid[i] = up->grid.cells[i]/2;
+      xc[i] = 1.0;
+    }
+    else {
+      idx_mid[i] = up->grid.cells[i]/2+1 ;
```
manauref, line 121: @akashukla better (recall integer division):
```
idx_mid[i] = up->grid.cells[i]/2;
xc[i] = up->grid.cells[i]%2 == 0? 1.0 : 0.0;
```

zero/gk_geometry_tok.c
```diff
@@ -129,6 +130,7 @@ gkyl_gk_geometry_tok_new(struct gkyl_gk_geometry_inp *geometry_inp)
   up->ref_count = gkyl_ref_count_init(gkyl_gk_geometry_free);
   up->on_dev = up; // CPU eqn obj points to itself
 
+  gkyl_tok_geo_release(geo);
```
manauref, line 133: @akashukla is this the bug that @Maxwell-Rosen found in the mirror branch? is there a similar fix for mirror_geo?

zero/gkyl_gk_geometry.h
```diff
@@ -39,6 +39,7 @@ struct gk_geometry {
   struct gkyl_array* gyyj;
   struct gkyl_array* gxzj;
   struct gkyl_array* eps2; // eps2 = Jg^33 - J/g_33
+  struct gkyl_array* bmag_mid;
```
manauref, line 42: @akashukla Can you add comments explaining all the entries in this struct please?

zero/spitzer_coll_freq.c
```diff
@@ -9,6 +9,37 @@
 #include <gkyl_spitzer_coll_freq_priv.h>
 #include <gkyl_range.h>
 
+// Calculate the plasma frequency
+double plasma_frequency(double n, double m, double eps0, double hbar, double eV)
```
manauref, line 13: @akashukla why is hbar an argument here?

zero/spitzer_coll_freq.c
```diff
@@ -9,6 +9,37 @@
...
+double coulomb_log(double ns, double nr, double ms, double mr, double Ts, double Tr, double qs, double qr, double bmag_mid, double eps0, double hbar, double eV)
+{
+
+  double vts = sqrt(Ts/ms);
+  double vtr = sqrt(Tr/mr);
+  double wps = plasma_frequency(ns,ms, eps0, hbar, eV);
+  double wpr = plasma_frequency(nr,mr, eps0, hbar, eV);
+  double wcs = qs*bmag_mid/ms;
+  double wcr = qr*bmag_mid/mr;
+  double inner1 = (wps*wps + wcs*wcs)/(Ts/ms + 3*Ts/ms) + (wpr*wpr + wcr*wcr)/(Tr/mr + 3*Ts/ms);
```
manauref, line 28: @akashukla why is this called inner again? I think it has a physical reason. Please comment in code for clarity/interpretability

zero/tok_geo.c
```diff
@@ -597,6 +597,11 @@ gkyl_tok_geo_get_stat(const gkyl_tok_geo *geo)
 void
 gkyl_tok_geo_release(gkyl_tok_geo *geo)
 {
+  gkyl_array_release(geo->psiRZ);
+  gkyl_array_release(geo->psibyrRZ);
+  gkyl_array_release(geo->psibyr2RZ);
+  gkyl_array_release(geo->fpoldg);
+  gkyl_array_release(geo->qdg);
```
manauref, line 604: @akashukla likewise, are these bugs @Maxwell-Rosen found in the mirror branch? are there similar bugs in mirror_geo?

apps/gk_field.c
```diff
@@ -55,10 +55,19 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
       f->sheath_vals[2*j+1] = mkarr(app->use_gpu, 2*app->confBasis.num_basis, app->local_ext.volume);
     }
   } else {
+    struct gkyl_array* bmag_mid_host = app->gk_geom->bmag_mid;
+    if (app->use_gpu) {
+      bmag_mid_host = mkarr(false, 1, 1);
+      gkyl_array_copy(bmag_mid_host, app->gk_geom->bmag_mid);
+    }
+    double *bmag_mid_ptr = gkyl_array_fetch(bmag_mid_host, 0);
+    double bmag_fac = f->info.bmag_fac ? f->info.bmag_fac : bmag_mid_ptr[0];
```
manauref, line 64: @akashukla I think we should change `bmag_fac` in the app (and input files if it's still in use) to bmag_ref. `_fac` is super vague. Or since we have `polarization_density` and this bmag_fac is specifically for the polarization, maybe it should be `polarization_bmag`

apps/gk_field.c
```diff
@@ -55,10 +55,19 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
...
+    if (app->use_gpu)
+      gkyl_array_release(bmag_mid_host);
```
manauref, line 66: @akashukla a cleaner way to do array allocation/release for the case where we are supporting GPUs too (having learned about gkyl_array_acquire) is 
```
struct gkyl_array* bmag_mid_host = app->use_gpu? mkarr(false, 1, 1) : gkyl_array_acquire(app->gk_geom->bmag_mid);
gkyl_array_copy(bmag_mid_host, app->gk_geom->bmag_mid);
...
gkyl_array_release(bmag_mid_host);
```
it uses fewer lines and fewer if-statements.
FYI @tnbernard @dingyunl @Maxwell-Rosen

apps/gk_species_lbo.c
```diff
@@ -30,15 +31,25 @@ gk_species_lbo_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s, stru
...
+    double eV = s->info.collisions.eV ? s->info.collisions.eV: GKYL_ELEMENTARY_CHARGE;
+    struct gkyl_array* bmag_mid_host = app->gk_geom->bmag_mid;
+    if (app->use_gpu) {
+      bmag_mid_host = mkarr(false, 1, 1);
+      gkyl_array_copy(bmag_mid_host, app->gk_geom->bmag_mid);
+    }
+    double *bmag_mid_ptr = gkyl_array_fetch(bmag_mid_host, 0);
+    double bmag_mid = s->info.collisions.bmag_mid ? s->info.collisions.bmag_mid : bmag_mid_ptr[0];
+    if (app->use_gpu)
+      gkyl_array_release(bmag_mid_host);
```
manauref, line 46: @akashukla see earlier comment about more compact way to handle allocate/deallocate bmag_mid_host.

apps/gkyl_gyrokinetic.h
```diff
@@ -49,11 +49,11 @@ struct gkyl_gyrokinetic_collisions {
 
   // inputs for Spitzer collisionality
   bool normNu; // Set to true if you want to rescale collision frequency
-  double self_nu_fac; // Self collision frequency without factor of n_r/(v_ts^2+v_tr^2)^(3/2)
-  double cross_nu_fac[GKYL_MAX_SPECIES]; // Cross collision freqs without factor of n_r/(v_ts^2+v_tr^2)^(3/2)
+  double n_ref;
```
manauref, line 52: @akashukla please add comments explaining these inputs

apps/gyrokinetic.c
```diff
@@ -299,6 +299,16 @@ gkyl_gyrokinetic_app_new(struct gkyl_gk *gk)
...
   gkyl_gk_geometry_release(gk_geom_3d); // release temporary 3d geometry
 
+  gkyl_gk_geometry_bmag_mid(app->gk_geom); // set bmag mid
+  int comm_sz;
+  int bcast_rank;
+  gkyl_comm_get_size(app->comm, &comm_sz);
+  if (comm_sz == 1)
+    bcast_rank = 0;
+  else
+    bcast_rank = comm_sz%2 == 0 ? comm_sz/2 - 1 : comm_sz/2;
```
manauref, line 309: @akashukla again, leverage integer division. You don't need this if-statement.

apps/gyrokinetic.c
```diff
@@ -299,6 +299,16 @@ gkyl_gyrokinetic_app_new(struct gkyl_gk *gk)
 
   gkyl_gk_geometry_release(gk_geom_3d); // release temporary 3d geometry
 
+  gkyl_gk_geometry_bmag_mid(app->gk_geom); // set bmag mid
+  int comm_sz;
+  int bcast_rank;
```
manauref, line 304: @akashukla declare on the same line please

regression/rt_gk_asdex_3x2v_p1.c
```diff
@@ -467,7 +467,6 @@ main(int argc, char **argv)
 
   // field
   struct gkyl_gyrokinetic_field field = {
-    .bmag_fac = ctx.B0,
```
manauref, line 470: @akashukla there are some 1x mirror input files with this bmag_fac. Delete them please. I believe that model doesn't use it, and we don't want to set a bad example.

manauref (commented):
Very good. I only made minor comments/request, and have just one general question:
Did you check that normNu or the collision frequency comes out correct for at least 3 species? i.e. nu_sr, nu_st, nu_rs, nu_rt, nu_ts, nu_tr

zero/gk_geometry.c
akashukla: I don't think the first line works the way I want. If up->grid.cells[i] = 4, I want idx_mid[i] = 2. If up->grid.cells[i] = 3, I want 2. But 3/2 = 1 in c right?

akashukla: I'll do this:

zero/spitzer_coll_freq.c
akashukla: I don't know if there is a physical meaning. To me they were just temporary variable names for the terms that appear inside the logarithm.

regression/rt_gk_asdex_3x2v_p1.c
akashukla: I think the 1d field model does use these. If I remove them the factor will default to bmag_mid calculated internally. I don't know the mirror tests, so up to you.

akashukla:
> Very good. I only made minor comments/request, and have just one general question: Did you check that normNu or the collision frequency comes out correct for at least 3 species? i.e. nu_sr, nu_st, nu_rs, nu_rt, nu_ts, nu_tr

For the step regression test with 3 species, I checked normNu for all self and cross collisions against normNu from gk-g0-app. Mostly they are identical but some differ by about 10% which is from including the cyclotron frequency in the calculation of normNu. (I checked earlier that this does indeed change the coulomb logarithm some).

regression/rt_gk_asdex_3x2v_p1.c
manauref: ah, the boltzmann electron model doesn't use it, so we can delete it in those. But the kinetic electron does, and that's just the polarization_bmag i suggested

zero/gk_geometry.c
manauref: you can do what I suggested, and evaluate at -1 instead of 1 in the even case, right?

manauref:
great

zero/gk_geometry.c
akashukla: I don't think so, because the issue is with the odd case not the even. It is only one extra line now.

PR #356, opened 2024-05-01
Change meaning of GKYL_SPECIES_FIXED_FUNC BC option

manauref:
Mostly symbolic PR (see commit message) but also to ask:
- Do you approve?
- Want me to rename this BC in moment/vlasov/pkpm apps too? it's just a `sed`


ammarhakim:
You have not changed this in the Lua driver code and so the build with Lua will break. If this is a cosmetic change I recommend waiting till that branch is merged.

Issue #39, opened 2022-04-25
1x3v esshock_lbo regression test incorrect behavior


JunoRavin:
The 1x3v esshock regression test does not shock, and higher velocity resolution leads to a time-step crash. Other 1x3v tests behave appropriately. Need to determine if error is in initial conditions or in the 1x3v kernels.

JunoRavin:
Was a problem with the ICs. Fixed in commit [92bbb6694bc5664fc79fa33bd96720008b64aa39 ](https://github.com/ammarhakim/gkylzero/commit/92bbb6694bc5664fc79fa33bd96720008b64aa39)

PR #359, opened 2024-05-07
Nonuniform velocity map (for gyrokinetics)

manauref:
(this goes with [gkylcas PR #39](https://github.com/ammarhakim/gkylcas/pull/39))

Implement support for nonuniform velocity (in gyrokinetics).
- Allow users to specify a mapc2p for each species, specifying a univariate c2p map for each velocity coordinate.
- Create a gkyl_velocity_map object which discretizes the velocity c2p map (vmap), and stores other information necessary to use it.
- Pass this mapc2p to various objects used in gyrokinetics, e.g. projectors, moment calculators, collisionless and collisional terms.

Some practical details:
1. When the user doesn't provide a species mapc2p, gkyl_velocity_map creates an identity map under the hood.
2. This replaces all previous GK kernels and functionality, there is not separate chunk of code for uniform GK anymore. The only exception is LBO kernels, which do have separate uniform (nomap) kernels because the nonuniform recovery for the diffusion terms produces very large, and presumably costly, kernels.
3. In the future, when writing code, we should not use the grid objects to obtain velocity coordinates. We should us the velocity_map object.
4. We have a total Jacobian quantity, `jacobtot`, in the code. The current implementation doesn't actually include the factor of jacobvel in jacobtot. Partly because `jacobtot` is created/assigned by the geometry object, and this work is independent of the geometry object. We could reconsider this, perhaps by making `jacobtot` a species object, specially since in some formulations the phase-space jacobian is B^*_\parallel/m, which is species dependent.
5. The distribution written out is in computational coordinates and contains a factor of jacobvel in it. To plot it one must use something like
```
pgkyl --c2p-vel <sim name>-elc_mapc2p_vel.gkyl <sim name>-elc_0.gkyl -t jf <sim name>-elc_jacobvel.gkyl -t jac ev -t distf 'jf jac /' activ -t distf ...
```

Other work that happened in this branch in the process (that I can remember, there may be more):
- We rewrote the LBO back to the 2 IBP formulation.
- Introduced `GKYL_MAX_VDIM` in gkyl_util.h.
- Allowed Vlasov to specify a zero-flux BC on one boundary but not the other.
- Did further cleaning and standardization of gk regression tests (including removing some mirror tests because we don't actually need so many).
- Plugged in all the gk reg tests to ci/gk_regression, but I can't get it to run all the tests at once due to catastrophic abortions. I don't know if this is a problem in ci/gk_regression, or a problem with memory errors in the GK solver. One suggestion: listing tests and their outputs in ci/gk_regression is a big pain and very prone to human error; we should instead gather all the gk_ tests in regression/ automatically.
- Reinterpret the meaning of a FIXED_FUNC BC in GK. Instead of using the initial skin cell, now the user must provide the function (in the form of a projection input struct) which gets projected into the skin cell at t=0.
- Introduce a factor of 2 in M3perp, so that M3 = M3par+M3perp. This way it is consistent with M2perp.

Tests and sims ran:
- There are several collisional, collisionless and sheath sims with nonuniform velocity in regression/ now. I ran these on CPU and GPU.
- A Boltzmann electron mirror sim using vparmax=20v_t, reproducing a the result in our 2023 PoP.
- Numerous LAPD turbulence simulations on GPU.

Curiously, I've seen an identical LAPD simulation survive in this branch that crashes in gk-g0-app.

Some things that remain to be done (we could add it to this PR before closing it):
- Done in [commit 3f719585](https://github.com/ammarhakim/gkylzero/commit/3f7195857d0682e780da1a1fa3784e2f65e53828): Infrastructure to allow nonuniform v when the user uses the `GKYL_PROJ_FUNC` option. The thing is that currently it give the user computational coordinates, but I think we need it to give physical (velocity) coordinates for ease of use.
- Done in [commit 5f73fec](https://github.com/ammarhakim/gkylzero/pull/359/commits/5f73feccebaa0b02500c391eca68d2c6cc119920): Add nonuniform v to radiation operator.
- Currently the diffusion surface term does recovery in v-space, in a way, rather than in computational space, like Greg Hammett suggested. This seems to work, but I think we should experiment with doing recovery in comp space to see what happens. [UPDATE:] Done; I tried doing recovery in computational space, and tested it with the bump-in-tail test, but it produced a different steady state, with a large difference at vpar=0, and momentum and energy where not conserved.

It would be good if @tnbernard @akashukla @Maxwell-Rosen @dingyunl run their sims in this branch with nonunfiform v so we can stress test this further.


ammarhakim:
Let’s wait on this till I return from my trip. As this is so critical we need to make sure the design is something we agree on. I will try and take a look this week, but it will be difficult.Sent from my iPhoneOn May 7, 2024, at 7:29 PM, Manaure Francisquez ***@***.***> wrote:﻿
@manauref requested your review on: #359 Nonuniform velocity map (for gyrokinetics).

PR #363, opened 2024-05-15
Nonuniform grids for mirror simulations

Maxwell-Rosen:
The branch is valgrind clean and passes basic testing to show that the geometries are identical for the uniform and non-uniform cases. Initial conditions are projected onto the non-uniform geometry as well. The issues related to parallelism are resolved by the allgather in gyrokinetic.c. Style has been reviewed by @manauref, but additional style review is needed prior to merging.


apps/gyrokinetic.c
```diff
@@ -13,6 +13,8 @@
 
 #include <mpack.h>
 
+#include <gkyl_mirror_geo.h>
```
JunoRavin, line 16: This should be the gyrokinetic_priv header

apps/gyrokinetic.c
```diff
@@ -279,6 +284,52 @@ gkyl_gyrokinetic_app_new(struct gkyl_gk *gk)
...
+        gkyl_gk_geometry_release(gk_geom_3d); // release temporary 3d geometry
+        gkyl_gk_geometry_release(app->gk_geom); // release 3d geometry
+        gkyl_array_release(c2fa_app->c2fa);
+        gkyl_array_release(c2fa_app->c2fa_deflate);
+        gk_geom_3d = gkyl_gk_geometry_mirror_new(&geometry_inp);
+        if(app->cdim < 3) {
+          gkyl_gk_geometry_c2fa_deflate(gk_geom_3d, &geometry_inp);
+        }
+        gkyl_array_release(bmag_global);
+        // I don't think I'm releasing the uniform deflated geometry correctly
```
JunoRavin, line 328: This code sequence seems somewhat convoluted. We have multiple deflations of the geometry, quantities being released then reallocated and there is still a final deflation call following. 

At minimum this section of the code needs more comments for why it needs to do things in this order. I understand that there is some complication here because we need to get the magnetic field before we can know what the non-uniform mapping is, so in effect we have to compute the geometry multiple times (once uniform and once non-uniform) but if anything it feels like some of this sequence should not be app-side, some of the sequence should be within the mirror geometry initialization (in the same way that we have a very particular call sequence for initializing geometry on GPUs where things are initialized host-side, then copied over to the device, in effect creating a situation where geometry on GPUs is always from file).

Maxwell-Rosen:
I ran a test to compare the uniform and non-uniform grids. It seems that there is a positivity issue with the temperature when reducing the cells by half. The initial study with uniform grids used 288 cells, so here we consider 280 cells as the benchmark. The non-uniform grid with the same total number of cells compares well for the entire length of the simulation, 100 microseconds. Reducing the number of cells by a factor of two compares well at early times, 10 microseconds, but the temperatures go negative around 50 microseconds. The codes used to run this are on my private repo, gkylmax, under commit 8819a9a157a73a2f108e5a09fb1049cd071f9937. 

moments_0
moments_1
moments_2
moments_3
moments_4
moments_5
moments_6
moments_7
moments_8
moments_9
moments_10
moments_11
moments_12
moments_13
moments_14
moments_15
moments_16
moments_17
moments_18
moments_19
moments_20

Maxwell-Rosen:
Plots of the moments for the mirror geometry simulations, repeated identically to the WHAM simulations but with a different geometry and a slightly different implementation of the same non-uniform grid algorithm

moments_0
moments_1
moments_2
moments_3
moments_4
moments_5
moments_6
moments_7
moments_8
moments_9
moments_10
moments_11
moments_12
moments_13
moments_14
moments_15
moments_16
moments_17
moments_18
moments_19
moments_20

Maxwell-Rosen:
These plots scan resolution, holding the minimum cell size fixed. Decreasing the total cell count decreases the number of cells in the center and the expander.


moments_0
moments_2
moments_4
moments_6
moments_8
moments_10

ammarhakim:
Make it even more coarse: 50 cells. Seems that phi is pretty well
established with a coarse grid.

Maxwell-Rosen:
I tried and it went wildly unstable with 100 cells within 10 microseconds. I only am reporting the sims before the temperature goes negative. There are some serious issues when going to lower cell counts. Perhaps using the positivity fix will address the issues with the temperature. It is strange that the throat regions are the locations of these instabilities when they are also the regions we are packing more cells into

PR #364, opened 2024-05-16
Canonical Poisson Bracket Equilibrium

johnson452:
Adds the capability to have equilibrium initial projection for Canonical-PB. This is used for both initial conditions and BGK operations.

Added:
1. det_h function needed in the input script for normalizing f_lte. This is initially projected on basis in vm_species.c and used in proj_lte
2. New Kernels have been generated to computed the canonical-pb pressure moment using the existing M2ij moment calculator. This is needed because the temperature moment: Jv*d*T = int( h^{ij}(w_i - u_i)*(w_j - u_j)*f*dw), depends on the inverse metric tensor. 
3. Added a new unit test which verifies the moments are projected correctly on a curved surface via the canonical-pb framework. ([unit/ctest_canonical_pb_equilibrium.c](https://github.com/ammarhakim/gkylzero/compare/can_pb_eq?expand=1#diff-83cb8fa3c888c16d8f12c23654d674ddf9f429b55235cccc0a6e5eec3fde42f5))

Notes:
a. For canonical-pb, the lte-moments are assumed to be [Jvn, u_i, T]. Places in the code that need n, like projecting f_lte requires det_h to isolate n.
b. Works for GPU.
c. Math-scratch notes available here: [Canonical_Poisson_Bracket-3.pdf](https://github.com/ammarhakim/gkylzero/files/15342260/Canonical_Poisson_Bracket-3.pdf)


zero/gkyl_dg_calc_canonical_pb_vars_priv.h
```diff
@@ -149,4 +161,18 @@ choose_canonical_pb_alpha_surf_v_kern(int dir, int cdim, int poly_order)
     return ser_canonical_pb_alpha_surfvz_kernels[cdim-1].kernels[poly_order];
   else
     return NULL;
+}
+
+GKYL_CU_D
+static canonical_pb_pressure_t
+choose_canonical_pb_pressure_kern(int dir, int cdim, int poly_order)
```
JunoRavin, line 168: This kernel selection is wrong. There is only one kernel selected so you should not be passing dir, and you're just calling the same return method three times. We should fix this. You only need cdim to select the correct pressure kernel.

zero/vlasov_lte_moments.c
```diff
@@ -212,11 +249,21 @@ gkyl_vlasov_lte_moments_release(gkyl_vlasov_lte_moments *lte_moms)
...
     gkyl_array_release(lte_moms->GammaV_inv);
-    gkyl_array_release(lte_moms->GammaV2);    
+    gkyl_array_release(lte_moms->GammaV2);  
+    gkyl_dg_updater_moment_release(lte_moms->Pcalc);  
+  }
+  else if (lte_moms->model_id == GKYL_MODEL_CANONICAL_PB) {
+    gkyl_array_release(lte_moms->h_ij_inv);
+    gkyl_array_release(lte_moms->det_h);
+    gkyl_array_release(lte_moms->pressure_tensor);
+    gkyl_dg_updater_moment_release(lte_moms->M2ijcalc);
```
JunoRavin, line 259: I think we should either call the needed "M2" calculator Pcalc for all of these quantities (so there's not a distinct M2ijcalc for the canonical PB case) or we should have a different name for the needed 2nd moment updater in all these cases (so for the non-relativistic, non-PB case, we should call it M2calc instead of Pcalc). 

Up to you on whether you want to re-use Pcalc in all three cases or call is something different and release it specifically in all three cases.

zero/gkyl_dg_calc_canonical_pb_vars_priv.h
johnson452: Fixed in commit https://github.com/ammarhakim/gkylzero/pull/364/commits/cd5dd0f23cdbc04e19cc228ce3e8d532a945dddc

zero/vlasov_lte_moments.c
johnson452: Fixed in commit https://github.com/ammarhakim/gkylzero/commit/cd5dd0f23cdbc04e19cc228ce3e8d532a945dddc

PR #367, opened 2024-05-22
Can pb implicit bgk

johnson452:
Adds implicit bgk collisions which are first-order strange split from the explicit rk3 step. Tests with the canonical-pb show this works well with sod-shocks in 1x1v to reproduce the results: https://ammar-hakim.org/sj/je/je28/je28-boltz-bgk.html?highlight=sod. 

Can-pb 1x1v sod shock in the fluid limit using implicit BGK: (Note these are the n, ux, T moments)

Equations for this system are described in the attached note (section 5): [Canonical_Poisson_Bracket-4.pdf](https://github.com/ammarhakim/gkylzero/files/15407584/Canonical_Poisson_Bracket-4.pdf).


ammarhakim:
We should not modify rk3. Add a new method that calls the rk3 and then calls a BGK step in it. So it will only be a few extra lines of code. Then choose this scheme if user has said to use the implicit BGK. This way we will not mess up our RK3 stepper. If we have to do other things in RK3 later it will be very confusing to figure out why this BGK implicit step was there.

manauref:
Perhaps rather than a `has_implicit_bgk_scheme` boolean, we should create a new enum which specifies various time stepping options, like RK3_SSP, RK3_SSP_STRANG_SPLIT_COLLISSIONS, etc, and this enum is used to pick the appropriate stepper function.

Because of this work may be extended to LBO in some form, so the infra doesn't need to be BGK-specific.

ammarhakim:
No need for new enums. Just rename the flag to implicit collisions flag so we can use it for lbo also. Lets not introduce more enums at present.Sent from my iPhoneOn May 22, 2024, at 8:04 PM, Manaure Francisquez ***@***.***> wrote:﻿
Perhaps rather than a has_implicit_bgk_scheme boolean, we should create a new enum which specifies various time stepping options, like RK3_SSP, RK3_SSP_STRANG_SPLIT_COLLISSIONS, etc, and this enum is used to pick the appropriate stepper function.
Because of this work may be extended to LBO in some form, so the infra doesn't need to be BGK-specific.

PR #368, opened 2024-05-22
Add a positivity enforcing shift operator 

manauref:
It shifts f to a value `ffloor_fac*max(f)` at positivity control points, which are +/- 1/3 in the p=1 directions, and +/- 1/3 and 0 in the p=2 directions. However, presently ffloor is set to 0.

It also computes the integrated moments of the shift
delta f = f_after_shift - f_before_shift
so that we know how many particles, how much momentum and how much energy we are adding.

To turn it on set `.enforce_positivity = true` in a species table.

Future work:
- We could change this to set thing to positive in other nodes, like quadrature nodes or corners. The latter may be desirable considering that we are not doing exponential reconstruction right now so f at the cell boundary can come out negative, thus convecting negative f to neighbor cells.
- We could use a diffusive shift instead, perhaps replacing f with 0.5*(f + f_flat_within a cell).

FYI @gwhammett


ammarhakim:
Positivity nodes seem risky to use. Can we change this to choose the nodes we want? So either positivity or guass nodesSent from my iPhoneOn May 22, 2024, at 3:10 PM, Manaure Francisquez ***@***.***> wrote:﻿It shifts f to a value ffloor*max(f) at positivity control points, which are +/- 1/3 in the p=1 directions, and +/- 1/3 and 0 in the p=2 directions. However, presently ffloor is set to 0.
It also computes the integrated moments of the shift
delta f = f_after_shift - f_before_shift
so that we know how many particles, how much momentum and how much energy we are adding.
To turn it on set .enforce_positivity = true in a species table.
Future work:

We could change this to set thing to positive in other nodes, like quadrature nodes or corners. The latter may be desirable considering that we are not doing exponential reconstruction right now so f at the cell boundary can come out negative, thus convecting negative f to neighbor cells.
We could use a diffusive shift instead, perhaps replacing f with 0.5*(f + f_flat_within a cell).




You can view, comment on, or merge this pull request online at:
  https://github.com/ammarhakim/gkylzero/pull/368

Commit Summary

  a61b25f  New updater to enforce positivity of f by shifting it up to a floor at posivitiy control nodes-ish (-1/3,0,1/3 for p=2). The updater also computes contributions to the integrated moment of the shift in f, so we can compare against int_mom diagnostics. The updater calculates the floor in _advance, as a (small) fracion of the maximum f in the grid. It might be too ambicious to do all of this in one loop (calculate ffloor, shift f, calculate the int moms of the shift), but I think it's possible. Not yet tested (gkylcas ec0722004018bc96a2ab65532e8bac67cb9651c0).
  fa3df08  Add unit test for positivity_shift_gyrokinetic. Not yet tested on GPU.
  2acc64c  Plug in positivity_shift to the GK app. It seems to get the job done. Tested it with gk_sheath_1x2v at relevant resolution, and only seemed to add 0.01% of the particles already in the sim. Of course, this isn't the best test as it is already well. Behaved. Will wait for the restart branch to get merged before testing in a turbulent sim.
  6f3fd92  Merge gk-g0-app -> gk-g0-app_posShift.
  995570c  Merge branch 'gk-g0-app' into gk-g0-app_posShift
  596551f  Fix GPU build of gk positivity shift updater. 1x2v test seems to produce the same results on 1 and 2 GPUs^Cs on 1 CPU.
  3d9bc68  We found that using a floor of 0.0 instead of some factor times fmax works much better for STEP multispecies Argon simulations. This fix enables us to run past transients that were initially crashing multispecies sims.


File Changes  (17 files)

  
    M
    Makefile
    (4)
  
  
    M
    apps/gk_species.c
    (16)
  
  
    M
    apps/gkyl_gyrokinetic.h
    (53)
  
  
    M
    apps/gkyl_gyrokinetic_priv.h
    (16)
  
  
    M
    apps/gyrokinetic.c
    (67)
  
  
    A
    kernels/positivity_shift/gkyl_positivity_shift_gyrokinetic_kernels.h
    (16)
  
  
    A
    kernels/positivity_shift/positivity_shift_gyrokinetic_1x1v_ser_p1.c
    (70)
  
  
    A
    kernels/positivity_shift/positivity_shift_gyrokinetic_1x2v_ser_p1.c
    (118)
  
  
    A
    kernels/positivity_shift/positivity_shift_gyrokinetic_2x2v_ser_p1.c
    (214)
  
  
    A
    kernels/positivity_shift/positivity_shift_gyrokinetic_3x2v_ser_p1.c
    (406)
  
  
    M
    unit/ctest_integrated_moms.c
    (24)
  
  
    A
    unit/ctest_positivity_shift_gyrokinetic.c
    (293)
  
  
    M
    zero/bc_sheath_gyrokinetic.c
    (2)
  
  
    A
    zero/gkyl_positivity_shift_gyrokinetic.h
    (50)
  
  
    A
    zero/gkyl_positivity_shift_gyrokinetic_priv.h
    (91)
  
  
    A
    zero/positivity_shift_gyrokinetic.c
    (103)
  
  
    A
    zero/positivity_shift_gyrokinetic_cu.cu
    (110)
  


Patch Links:

  https://github.com/ammarhakim/gkylzero/pull/368.patch
  https://github.com/ammarhakim/gkylzero/pull/368.diff

gwhammett:
Thinking through some options:

In Mana's comments, he mentions that future possible work is "change this
to set things to positive in other nodes, like quadrature nodes or
corner".   There are 3 options:

(1) Doing it at positivity control points is the most conservative thing to
do as a first step, i.e., it does the least damage to the F calculated by
the standard DG algorithm, but until other positivity steps are
implemented, it doesn't guarantee that positivity problems won't occur.

(2) Imposing F>=0 at quadrature points will ensure positivity of density
and temperature when calculating collision operators or atomic physics and
radiation terms.  But it won't prevent extrapolations to surfaces that give
negative F, which could cause a negative F into a neighboring cell,
possibly making the mean F in that neighboring cell to go negative.  That
would get corrected the next time the routine is called to enforce
positivity at quadrature points, so perhaps that is sufficient for now.  On
the other hand, once we get rigorous positivity enforcement working, this
would be overly aggressive, because it forces F=0 at quadrature points even
though negative F at quadrature points should be allowed.

(3) Imposing F>=0 at all corners (and a set of interior points to be a
complete basis) would be the most aggressive and the most robust, because F
is positive everywhere and there could never be a negative flux into a cell
to make it negative.  But perhaps this isn't necessary because option (2)
should be sufficient?

So although I initially advocated option (1) as a first step, perhaps it is
best to do option (2) for now, (imposing F>=0 at quadrature points).  Once
we get a rigorous positivity-preservation algorithm implemented, this could
be turned off.

I think at present Mana is imposing this F>=0 constraint only after all 3
Euler stages of a full SSP Runge-Kutta time step. Perhaps that is
sufficient for now, but we might need to impose it after each Euler stage.

ammarhakim:
Also we don’t know what positivity nodes for p=2 are. Hence having an option to use quadrature node is good to have.Sent from my iPhoneOn May 22, 2024, at 7:16 PM, Greg Hammett ***@***.***> wrote:﻿
Thinking through some options:



In Mana's comments, he mentions that future possible work is "change this

to set things to positive in other nodes, like quadrature nodes or

corner".   There are 3 options:



(1) Doing it at positivity control points is the most conservative thing to

do as a first step, i.e., it does the least damage to the F calculated by

the standard DG algorithm, but until other positivity steps are

implemented, it doesn't guarantee that positivity problems won't occur.



(2) Imposing F>=0 at quadrature points will ensure positivity of density

and temperature when calculating collision operators or atomic physics and

radiation terms.  But it won't prevent extrapolations to surfaces that give

negative F, which could cause a negative F into a neighboring cell,

possibly making the mean F in that neighboring cell to go negative.  That

would get corrected the next time the routine is called to enforce

positivity at quadrature points, so perhaps that is sufficient for now.  On

the other hand, once we get rigorous positivity enforcement working, this

would be overly aggressive, because it forces F=0 at quadrature points even

though negative F at quadrature points should be allowed.



(3) Imposing F>=0 at all corners (and a set of interior points to be a

complete basis) would be the most aggressive and the most robust, because F

is positive everywhere and there could never be a negative flux into a cell

to make it negative.  But perhaps this isn't necessary because option (2)

should be sufficient?



So although I initially advocated option (1) as a first step, perhaps it is

best to do option (2) for now, (imposing F>=0 at quadrature points).  Once

we get a rigorous positivity-preservation algorithm implemented, this could

be turned off.



I think at present Mana is imposing this F>=0 constraint only after all 3

Euler stages of a full SSP Runge-Kutta time step. Perhaps that is

sufficient for now, but we might need to impose it after each Euler stage.

gwhammett:
right.

ammarhakim:
Seems there is a conflict. Which should be fixed before merging the PR

gwhammett:
(ignore this comment until after the Memorial Day holiday.)  An earlier comment said "We found that using a floor of 0.0 instead of some factor times fmax works much better for STEP multispecies Argon simulations. This fix enables us to run past transients that were initially crashing multispecies sims."  It will be important to test this positivity-enforcing source again for these STEP cases and Tess's TCV cases, now that it has been changed to enforce positivity at Gaussian quadrature points.  It should still stop the crashes, but the question is how large are the particle and energy errors from this positivity source.  These errors can eventually be reduced, but it will be useful to know how large they are on some cases of interest.

PR #377, opened 2024-05-30
GK comms functions and further standardization of reg tests

manauref:
Create new gyrokinetic_comms functions which package the comm creation for our input tests. Benefits:
1. Removes repetitive 80-100 lines of code from each input file.
2. Reduce the risk of someone making a mistake in this logic in future input files.
3. When we parallelize species, we'll be able to add that additional feature to these new functions with minor modifications to input files.

Also standardize the way number of cells are set in regression tests. This makes it easier to make changes across all reg tests, again reduces the changes of introducing mistakes, etc.

I ran gk_sheath_2x2v on 1 and 2 CPUs or GPUs, reproducing expected results. I also compiled and ran (for 3 steps) every reg test, and through this found there was a bug in the vlasov_lte app for gk neutral species, which is fixed here.

### DIFF
--- a/apps/gk_neut_species_projection.c
+++ b/apps/gk_neut_species_projection.c
@@ -39,7 +39,7 @@ gk_neut_species_projection_init(struct gkyl_gyrokinetic_app *app, struct gk_neut
     struct gkyl_vlasov_lte_proj_on_basis_inp inp_proj = {
       .phase_grid = &s->grid,
       .conf_basis = &app->confBasis,
-      .phase_basis = &app->basis,
+      .phase_basis = &app->neut_basis,
       .conf_range =  &app->local,
       .conf_range_ext = &app->local_ext,
       .vel_range = &s->local_vel,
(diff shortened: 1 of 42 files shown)


akashukla (approved):
Looks good to me based on reviewing the changes in apps/ and a few regression tests I am familiar with. I have made sloppy mistakes creating the comms for reg tests/simulations, so I think this is helpful.

ammarhakim:
Don’t merge yet. I will do it tomorrow Sent from my iPhoneOn May 30, 2024, at 7:11 PM, akashukla ***@***.***> wrote:﻿
@akashukla approved this pull request.

Looks good to me based on reviewing the changes in apps/ and a few regression tests I am familiar with. I have made sloppy mistakes creating the comms for reg tests/simulations, so I think this is helpful.

manauref:
For the sake of discussion (FYI @JunoRavin), adding num_species to the input file context, and passing it to these functions, the creation of the conf-space communicator could be made species-parallelization aware (I think the species comm could/would be created inside the GK app).

I thought about doing it here, even if we are not yet using the species comm, but one thing I was unsure about we should discuss is how to handle the split between gk_species and gk_neut_species.

dingyunl:
I'm trying to run a longer regression test with the asdex geometry and I need the parallelization feature in the input file. Is this branch about ready to merge?

ammarhakim:
@dingyunl the regression tests are already easy to parallelize. This just wraps a few things to make it cleaner. You do not need to wait for this.

PR #383, opened 2024-06-11
Fixing conflicts for pulling main into gk-g0-app.

JunoRavin:


JonathanGorard:
Which conflicts specifically were fixed here? Are there any particularly thorny ones that we ought to know about?

In any case, I'm not sure that creating an intermediate branch with a single monolithic commit in this way is necessarily the correct way to handle this merge, since I worry (unless I'm missing something) that we may end up losing commit history from `main`. Is there a reason why we don't create a PR directly for `main` -> `gk-g0-app`? I realize that there are merge conflicts, so such a PR cannot be merged automatically, but that's where (in my opinion) such conflicts should be resolved.

manauref:
> Which conflicts specifically were fixed here? Are there any particularly thorny ones that we ought to know about?
> 
> In any case, I'm not sure that creating an intermediate branch with a single monolithic commit in this way is necessarily the correct way to handle this merge, since I worry (unless I'm missing something) that we may end up losing commit history from `main`. Is there a reason why we don't create a PR directly for `main` -> `gk-g0-app`? I realize that there are merge conflicts, so such a PR cannot be merged automatically, but that's where (in my opinion) such conflicts should be resolved.

This was my idea, kind of.
I suggested JJ, locally, creates a new branch (e.g. gk-g0-app-merge-main), and merges main into it, just to learn what the conflicts are. However my intention was just to learn what the conflicts were (since online GitHub doesn't tell you what the conflicts are before you create the PR, I think), and then I thought JJ would submit the regular main->gk-g0-app PR and fix the conflicts there as JG suggested. I'm not sure it'll work, or it'll be easy to resolve the conflicts in the PR itself, but there's no harm in trying @JunoRavin

regression/rt_advect_1x.c
```diff
@@ -350,7 +350,7 @@ main(int argc, char **argv)
   gkyl_vlasov_app_cout(app, stdout, "Species collisional moments took %g secs\n", stat.species_coll_mom_tm);
   gkyl_vlasov_app_cout(app, stdout, "Total updates took %g secs\n", stat.total_tm);
 
-  gkyl_vlasov_app_cout(app, stdout, "Number of write calls %ld,\n", stat.nio);
+  gkyl_vlasov_app_cout(app, stdout, "Number of write calls %ld\n", stat.nio);
```
manauref, line 353: I think this file should be renamed rt_vlasov_advect_1x or rt_vlasov_fluid_advect_1x

pcagas:
I might be wrong here but when fixing merge conflicts, you need to be able to commit to the branch you are merging from, right? And we can't do any commits directly to [main](https://github.com/ammarhakim/gkylzero/tree/main)

JunoRavin:
@JonathanGorard The problem is that we can't make any commits or fixes directly to gk-g0-app so I had to figure out a way to bring in where main is somehow and the first thing I thought of was a branch. But @pcagas is right that I need to also merge *this* branch with main so that the system knows it's being pushed up to date with main. 

The merge conflicts that were fixed were as follows:

1. gk-g0-app has made additions to the makefile (for adas) while main had made additions to the makefile for amr. 
2. The vlasov_lte projection and correction objects are taking new inputs in gk-g0-app (to handle projection onto the positivity nodes) while in main they are taking new inputs for the canonical Poisson bracket functionality. 
3. AMR needed changes to all its write methods to handle the fact that the write method now takes metadata (but just passing NULL to the metadata entry).

JunoRavin:
Now this branch knows main has been merged into it.

JonathanGorard:
This looks perfect, thank you @JunoRavin ! Thanks @pcagas for pointing out the issue with what I was proposing before.

Also, @manauref , the correct naming scheme in that case would be `rt_dg_advect_1x` (since we are separating fluid DG tests from kinetic DG tests, and only the latter receive the title "Vlasov"). The reason it has not been renamed is the same as the reason why none of the `rt_diffusion...` tests have been renamed: they currently do not run, and I would prefer not to have them grouped together with other, functional tests until the segfault that they induce has been fixed (likely to be handled as part of the general refactoring in the `dg_10m` branch in the very near future). The fact that they do not follow our updated naming convention reflects the fact that they have not been refactored, and should not be considered part of our testing suite until they are (and work).

JonathanGorard (approved):

ammarhakim (approved):
We need to take a careful look at the LTE project code to see if we can simplify it. Seems the PB, SR and non-SR are all squashed into a single updater, and it may be better to do it a bit more cleaning.

manauref:
> We need to take a careful look at the LTE project code to see if we can simplify it. Seems the PB, SR and non-SR are all squashed into a single updater, and it may be better to do it a bit more cleaning.

strongly in favor of this, especially since we need to be able to add the velocity map to the GK neutral species, but I don't want to affect, or have to check, that I didn't break all these other equation systems

PR #387, opened 2024-06-14
Preserve data in mem_buff upon resizing on the GPU, and test mem_buff

manauref:
1. For GPUs: change gkyl_mem_buff_resize so it preserves the data originally in the buffer, which is the current behavior on CPU.
2. Add a unit test for gkyl_mem_buff.

We are using this object in setting up TS BCs and noticed these missing pieces.

### DIFF
--- a/unit/ctest_alloc.c
+++ b/unit/ctest_alloc.c
@@ -40,8 +40,135 @@ test_aligned_realloc()
   gkyl_aligned_free(rd);
 }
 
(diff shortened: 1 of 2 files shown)


ammarhakim (approved):
This looks fine. However, in general one should refrain from reallocs on GPUs, specially in inner loops.

JunoRavin (approved):

Issue #385, opened 2024-06-13
[DR] Enable cuDSS-based linear solvers (on GPU)


manauref:
Presently we can't compile with cuda >=12.2 without getting lots of warnings about some cuSparse features being deprecated, and when you try to run our cuSolver-based linear solvers unit test, or GK reg tests, they seg fault (see issue #234 ).

On solution proposed here is to solve linear problems
A_i x_i = b_i where i=1,2,...,Nprob
or
A x_j = b_j where j=1,2,...,Nrhs
using [cuDSS](https://developer.nvidia.com/cudss) instead.

The idea is to:
1. Allow users to specify via machine-file flags (e.g. --use-cudss=yes) whether to use cuDSS. When true, gkylzero will download cuDSS into gkylsoft and link to it. This also sets GKYL_HAVE_CUDSS=1
2. Create a zero/cudss_ops.c module that has the same functionality (and hopefully more) functionality that's in cusolver_ops.c
3. Then updaters, like fem_poisson, can choose do perform certain steps with cuSolver or cuDSS based on `GKYL_HAVE_CUDSS=1` like (see around L197 of zero/fem_poisson.c)
```
  if (up->use_gpu) {
#ifdef GKYL_HAVE_CUDSS
    gkyl_cudss_amat_from_triples(up->prob_cu, tri);
#else
    gkyl_cusolver_amat_from_triples(up->prob_cu, tri);
#endif
  }
```
Eventually we may deprecate one of these two solvers, depending on NVIDIA's recommendations and maintenance schedules, and our users experience.

For now we focus on problems where A is time independent. I believe cuDSS will allow us to do the time dependent problem A=A(t) as well (e.g. for time dependent polarization density weights), which wasn't possible with cuSolver. Other benefits of cuDSS include improved performance, more operations happening on the GPU (cuSolver did some things on the host), and potentially better support within NVIDIA.

This is currently being prototyped in the `gk-g0-app_cudss-prototyping` branch, where an initial, simple version of zero/cudss_ops.c has been added, along with a unit test.

ammarhakim:
Comments:

- Move the ifdef preprocessor to the method themselves and not in each usage. So a common method called "cuda_nmat" and not "cusolver_nmat". For example.
- Replace printf to fprintf and use stderror file pointer.

manauref:
Addressed in PR #396

Issue #234, opened 2023-12-29
cuda linear solvers not working with cudatoolkit 12.3


manauref:
When compiling with cudatoolkit 12.3, we are getting a bunch of warnings about deprecated options in cuSPARSE. See below.
Then, unit tests that passed with cudatoolkit 12.0 are not giving `Test interrupted by SIGSEGV` errors. See below.

This may mean that our linear solvers need to be redesigned a little, or different options must be used. It's likely that some of the stuff we used before was unstable (as suggested by a mentor in a hackathon) and has now been removed from cudatoolkit.

Judging by which tests fail it seems that this is an issue with our infrastructure to solve multiple linear problems, as in the GK Poisson case.

Unit test errors:
```
[manaurer@stellar-m01g6 gkylzero]$ ./cuda-build/unit/ctest_linsolvers
Test cusolver_qr...                             [ OK ]
Test cusolver_rf...                             [ OK ]
Test cusolver_ops...                            [ OK ]
Test cusolver_ops_multiple_rhs...                 Test interrupted by SIGSEGV.
Test cusolver_ops_multiple_prob...                Test interrupted by SIGSEGV.
```

```
[manaurer@stellar-m01g6 gkylzero]$ ./cuda-build/unit/ctest_fem_poisson_perp
Test gpu_test_p1_dirichletx_dirichlety...         Test interrupted by SIGSEGV.
Test gpu_test_p1_dirichletx_dirichlety...         Test interrupted by SIGSEGV.
Test gpu_test_p1_dirichletx_periodicy...          Test interrupted by SIGSEGV.
Test gpu_test_p1_periodicx_dirichlety...          Test interrupted by SIGSEGV.
Test gpu_test_p1_dirichletx_neumanny_dirichlety...   Test interrupted by SIGSEGV.
Test gpu_test_p1_dirichletx_dirichlety_neumanny...   Test interrupted by SIGSEGV.
Test gpu_test_p1_neumannx_dirichletx_dirichlety...   Test interrupted by SIGSEGV.
Test gpu_test_p1_dirichletx_neumannx_dirichlety...   Test interrupted by SIGSEGV.
Test gpu_test_p1_neumannx_dirichletx_periodicy...   Test interrupted by SIGSEGV.
```

```
[manaurer@stellar-m01g6 gkylzero]$ ./cuda-build/unit/ctest_fem_parproj
Test gpu_test_1x_p1_nonperiodic...              [ OK ]
[...]

PR #396, opened 2024-06-18
Enable cuDSS solvers (and thus compiling/running gkylzero with cuda >12.1)

manauref:
This PR implements [cuDSS](https://developer.nvidia.com/cudss)-linear solvers, equivalent in capability to our previous cuSolver/cuSparse-based linear solvers on NVIDIA GPUs. It corresponds to the design proposed in Issue #385 . We've done so by:
1. Renaming the `gkyl_cusolver_` methods `gkyl_culinsolver_~` and the `gkyl_cusolver_prob` to `gkyl_culinsolver_prob`.
2. Adding another module, cudss_ops, which offers an alternative definition of these methods and objects, to use cuDSS instead.
3. Which definition of `gkyl_culinsolver_` is used gets decided at build-time, depending on the use or non-use of `--use-cudss` flags in mkdeps and configure scripts.

We could've instead added support for both cuSolver and cuDSS simultaneously through a hierarchy similar to that in gkyl_comm for mpi_comm/nccl_comm. The reason not to do that is that it is highly likely we will deprecate one of these options in the future, which would be very easy with the current structure and would keep the resulting code simpler.

Tests:
- We implemented new cuDSS unit tests, one which is the same as that in the CUDASamples repository and is a standalone test not depending on gkyl objects at all. Another which employs `gkyl_culinsolver_` and solves the SuperLU sample problem.
- Previous tests using cuSolver, i.e. `fem_poisson`, `fem_poisson_vareps`, `fem_poisson_perp`, `fem_helmholtz`, `deflated_fem_poisson`, `fem_parproj`, all pass as they did before.
- Also ran several GK regression tests (e.g. 3x LAPD and 2x gk_sheath tests) and got the same results as before, or within 1e-11.
- Also built the code from scratch with cuda 12.0 and --use-cudss=no, and confirmed that previous functionality is preserved (i.e. ran aforementioned tests).

This change will also allows us to run with cuda >12.1, thus addressing issue #234


ammarhakim:
@manauref  It is hard to see exactly what is going on here. Can you point to one file where the choice between the old solver and new one is being made? I do not see where the ifdef pre-processor directives are used.

manauref:
There's no if-else preprocessor statement, there's instead an `ifdef` at
the top of `zero/cudss_ops.c` (and `zero/gkyl_cudss_ops.h`) and an `ifndef`
at the top of `zero/cusolver_ops.c` (and `zero/gkyl_cusolver_ops.h`)

ammarhakim (approved):
This looks good and is ready for merging.

JunoRavin (approved):

PR #402, opened 2024-06-25
Initial changes to single block app for multib

manauref:
These are the changes to the single block app that we envision are presently needed for multiblocks. It's primarily:
- Implementing the SKIP_BC option so the multib app tells some block boundaries that this is their BC, but when their apply_bc method is called, nothing happens on this boundary (of this block).
- Divide some methods into other smaller methods to expose some operations, namely the calculation of df/dt and the applicaation of BCs, to the multiblock app.


ammarhakim:
There is no need for numbering in the enum gkyl_bc_basic_types of the enums. Enums are numbered automatically.

manauref:
1. That numbering was already there, so I didn't change it. I can remove it if you want.
2. dfdt needs to be exposed because we'd like to have our own MB forward euler, which calls the single block df/dt in each block.
3. Likewise, we want to call the single block apply_bc from the MB (e.g. for apply BCs during initialization), so we need to expose it.

Related to the last two points:
- The reason I put them in the public header was because I was anticipating creating some sample programs were we let ARKODE/SUNDIALS do the time stepping, and exposing df/dt will be needed to make that easier.
- If you'd prefer we could make these private methods, which are only visible to the MB app (and not the user).

ammarhakim:
I do not care about ARCODE. We will cross the bridge when we come to it. These should not be in the public API. I want to see if there are needed at all. I think dfdt is fine (we can call it calc_rhs instead) but need to review. We have a bunch of methods to apply BCs in the private header already and so the apply_bc method is not needed.

manauref:
- I removed the explicit numbering in the bc_basic enum.
- I moved the declarations of dfdt and apply_bc to the private header. Comments on this: a) _rhs is I think less clear for some people because we tend to think of the RHS as being just collisions+sources. b) we need the app_apply_bc function because we only have the species_apply_bc functions, and this new functions allows us to call the same code in MB that we already call in the single B, without duplication and potential divergences.

manauref:
Mod my last comment: for further specificity, i've renamed `gkyl_gyrokinetic_apply_bc` to `gkyl_gyrokinetic_apply_species_bc`, since we don't apply BCs to the fields as an additional step in GK (or in VP actually)

PR #405, opened 2024-06-26
Bump the openmpi version built by gkylzero from 4.1.5 to 4.1.6. This …

manauref:
…avoids the error message

checking if .note.GNU-stack is needed... /Library/Developer/CommandLineTools/usr/bin/objdump: error: 'conftest.o': Invalid/Unsupported object file format no
checking suffix for labels... :
checking prefix for global symbol labels... none
configure: error: Could not determine global symbol label prefix

on newer Macs or MacOS versions.

### DIFF
--- a/install-deps/build-openmpi.sh
+++ b/install-deps/build-openmpi.sh
@@ -3,7 +3,7 @@
 source ./build-opts.sh
 
 # Install prefix
-PREFIX=$GKYLSOFT/openmpi-4.1.5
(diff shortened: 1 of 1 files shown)


ammarhakim (approved):
Looks good and this is ready to merge.

JonathanGorard (approved):
Looks great, builds and runs in my Mac (M3 Pro, Sonoma 14.).

Issue #400, opened 2024-06-24
[DR] cublasDgemm for modal/nodal conversions


johnson452:
My prototyping branch dg_10m_cublasDgemm_prototyping was branched from dg_10m, and I propose to merge this back into dg_10m.

(Similar changes implemented in gk-g0-app-GPU-Hack2024)

Description: 
Adds cu_mat_mm_array(...) to replace existing bottleneck in vlasov-lte-project when converting from nodal to modal quantities. This function calls the cublas function cublasDgemm(…) which performs a matrix multiplications which is approximately 2x faster in production cases for 2x2v can-pb calculations than unrolling the nodal to modal loops with maxima. 

This function, cu_mat_mm_array(...),  will reside in (new) mat_priv header with implementation in mat.c. It uses the cublas_v2 library function cublasv2Gemm(), which computes: alpha\*matrix_multiplication(A,B) + Beta*C = C.  This is particularly efficient since the matrix A is constant in time which converts nodal quantities, B, into modal quantities, C. B and C are gkyl_arrays and no reshaping of the data is required to perform these operations.

Requires:
Created new structure in gkyl_mat.h (implemented in mat.c), gkyl_phase_nodal_to_modal_mem, to store A, cublashandle, transpose, info, alpha, beta. This has been added to gkyl_vlasov_lte_proj_on_basis which uses this information in the gpu implemtation of lte-proj-on-basis to replace the phase modal to nodal calculation.   Structure is populated in Vlasov-proj-lte inew method for the specific nodal to modal phase-space calculation.

Other:
-	Removed exponential factor since it doesn’t matter.
-	Det_h and h_ij_inv calcs moved outside of the loops since they can be pre-computed.

Extra checklist:
[X] Does prototype PR Work? 
[X] Unit test passes: ?
[X] Regression test passes?
[X] Is memory freed?

johnson452:
From our DR meeting (6/24/2024), the following changes were requested
- Move the cuda handles into gkyl_mat_mem structure
- With Openblas, make a CPU version of the matrix multiply, 
- Move the material from gkyl_mat_priv.h into gkyl_mat.h (which is enabled due to the first change - originally impossible because of having the cudahandle in the function sig.) Then remove gkyl_mat_priv.h and change the function signatures to gkyl_...

Further notes:
- The matrix is agnostic to the number of quadrature points and thus better for code maintainability

JunoRavin:
[...]

PR #406, opened 2024-06-27
Dg 10m cublas dgemm production

johnson452:
This is the production version of the cublasDgemm prototyping branch, see [DR](https://github.com/ammarhakim/gkylzero/issues/400). Primarily for computing with cblas/cublas the matrix multiplication: alpha*A(gkyl_mat) * B(gkyl_array/gkyl_mat) + Beta*C(gkyl_array/gkyl_mat) = C(gkyl_array/gkyl_mat). This has been used successfully to reduce the computation time of the nodal to modal conversion in vlasov-lte-proj-on-basis-cu, which we found to be more efficient than loop unrolling on GPU for this specific bottlenecked operation. 

Extensions to gkyl_mat functions:
- _gkyl_mat_mm_array_mem_, new structure to save the matrix A and associated, alpha, beta, transposition information for the matrix multiply.
- **gkyl_mat_mm_array_mem_dev_new**, Builds a new object for computing CPU/GPU side matrix multiplication.
- **gkyl_mat_mm_array_mem_release**, Associated release of the memory.
- **gkyl_mat_mm_array**, Function which does the matrix multiply either or CPU or GPU (depending on build/flags).

New unit tests: (in ctest_mat.c)
- **test_mat_mm_arrays**, Tests CPU side blas matrix multiplication: alpha*A(gkyl_mat) * B(gkyl_array) + Beta*C(gkyl_array) = C(gkyl_array).
- **test_cu_mat_mm**, Tests CPU side blas matrix multiplication: alpha*A(gkyl_mat) * B(gkyl_mat) + Beta*C(gkyl_mat) = C(gkyl_mat).
- **test_cu_mat_mm_arrays**, Tests GPU side blas matrix multiplication using cublas: alpha*A(gkyl_mat) * B(gkyl_array) + Beta*C(gkyl_array) = C(gkyl_array)

### DIFF
--- a/unit/ctest_mat.c
+++ b/unit/ctest_mat.c
@@ -689,7 +689,6 @@ void test_cu_mat_mm_arrays()
   struct gkyl_array *array_ycu = gkyl_array_cu_dev_new(GKYL_DOUBLE, 4, 2);
 
   // fill each A matrix with its number
-  printf("\n");
   for (size_t j=0; j<mat_A->nc; ++j){
     for (size_t i=0; i<mat_A->nr; ++i){
       double a_val = i*3 + j;
@@ -732,6 +731,8 @@ void test_cu_mat_mm_arrays()
 	cublasCreate_v2(&cuh);
 
   cu_mat_mm_array(cuh, alpha, beta, transa, mat_Acu, transb, array_xcu, array_ycu);
+
+  cublasDestroy(cuh);
 
   // copy to host
   gkyl_mat_copy(mat_A, mat_Acu);
--- a/unit/ctest_mat.c
+++ b/unit/ctest_mat.c
@@ -516,11 +516,65 @@ void test_nmat_mm()
       for (size_t i=0; i<y.nr; ++i)
         TEST_CHECK ( expected == gkyl_mat_get(&y, i, j) );
   }
-
-
 }
 
 
+void test_mat_mm_arrays()
+{
+  struct gkyl_mat_mm_array_mem *ctest_prob_mem; 
+  ctest_prob_mem = gkyl_mat_mm_array_mem_dev_new(4, 3, 1.0, 0.0, 
+    GKYL_NO_TRANS, GKYL_NO_TRANS, false);
+
+  struct gkyl_mat *mat_A = ctest_prob_mem->A_ho;
+  struct gkyl_array *array_x = gkyl_array_new(GKYL_DOUBLE, 3, 2);
+  struct gkyl_array *array_y = gkyl_array_new(GKYL_DOUBLE, 4, 2);
+
+  // fill each A matrix with its number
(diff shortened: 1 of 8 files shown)


zero/gkyl_mat.h
```diff
@@ -345,6 +349,43 @@ gkyl_nmat_mem *gkyl_nmat_linsolve_lu_cu_dev_new(size_t num, size_t nrow);
...
+ * @param nr Number of rows in the matrix A
+ * @param nc Number of columns in the matrix A
+ * @param alpha Coefficient infront of A*B
+ * @param beta Coefficient infron of C
+ * @param transa Whether or not to transpose A
+ * @param transb Whether or not to transpose B
+ * @param use_gpu 
+ * @return Preallocated memory
+ */
+gkyl_mat_mm_array_mem *gkyl_mat_mm_array_mem_dev_new(int nr, int nc, double alpha,
```
JunoRavin, line 365: All device allocation calls should be "cu_dev_new" not just "dev_new"

JunoRavin (commented):
Right now you allocate extra memory to transform modal to nodal quantities for h_ij_inv and det_h. I think we should delete this and just use the same loop over configuration space quadrature points in the init method. I don't think we should beg questions in the design why this conversion is done with matrix multiplication while the other configuration space evaluation is done by hand. If you have a reason you would prefer to leave it like this which allows us to document why these configuration space fields are done one way while other configuration space fields are done another way, let me know.

zero/gkyl_mat.h
johnson452: Actually this memory is now both used by the CPU/GPU routines and the memory lives on one or the other, so I believe it should just be: gkyl_mat_mm_array_mem_new

JunoRavin: That's fine!

JunoRavin (approved):

unit/ctest_mat.c
```diff
@@ -733,6 +732,8 @@ void test_cu_mat_mm_arrays()
 
   cu_mat_mm_array(cuh, alpha, beta, transa, mat_Acu, transb, array_xcu, array_ycu);
```
ammarhakim: Is this a public-facing API? I do not think this is correct. We need all methods to be prefixed by the gkyl_ prefix. Also, what is cuh? I thought no handles are to be exposed?

ammarhakim:
I do not think this PR is ready for merge. I am not sure the handles are properly ... handled ... and public facing APIs are not properly names. I am closing this PR for now to prevent anyone from "approving" it (as @JunoRavin approved but still changes are needed).

ammarhakim (commented):
Needs fixes as I indicated in my comment to the PR. Please fix and reopen the PR.

PR #410, opened 2024-07-01
Extensions to, and refactoring/cleanup of, the AMR system

JonathanGorard:
This PR constitutes a major step in the effort to turn the AMR framework for the moment app into a fully robust system that can be used for production-scale simulations (rather than simply a prototyping/research project).

It introduces the basic infrastructure for handling dynamic adaptivity and regridding in mesh refinement simulations (without requiring any modifications to the existing app code, since the `amr/amr_core_...` source files essentially replace all of the time-stepping and simulation logic that presently exists within the app layer), though more work is required on this before it is truly ready for prime time. It also adds fully customizable support for hierarchical nesting of AMR grids, with appropriate modifications to the adaptive time-stepper, and introduces further stabilization to the moment-EM coupling at coarse-fine boundaries (although there is much additional research work to be done on this too, in particular in relation to determining how best to damp the dispersive waves currently being excited at the AMR block boundaries). Finally, it removes a large amount of redundant/unnecessary code from the earlier, prototyping stages of the AMR project, and cleans up a lot of what remains (enforcing greater consistency in style and naming conventions, etc.)

Many new regression tests have been added to test the new functionality, and all of the new AMR (and older GR+AMR) regression tests have now been hooked up to the automatic regression testing system for the moment app. This has required some substantive modifications to the CI pipeline, since correctness and memory leak checking works somewhat differently for AMR/GR output than it does for all other aspects of the moment app.


JunoRavin:
In the sync_blocks methods, such as https://github.com/ammarhakim/gkylzero/blob/1ea596489bfc99b387f1f0a687b1bb4b28edfea1/amr/amr_block.c#L205

I think we should split out the "projection" and "restriction" operators that are filling the coarse or fine blocks into their own functions. This refactor has the advantage that the code becomes more readable without the need for excessive comments (since it will be obvious then from the API that one is moving from either a coarse block to a fine block or vice versa). It also allows you to have flexibility in the restriction operator, either skipping data as you are currently doing or some sort of more sophisticated reconstruction (which maybe permits conservative fluxes at coarse-fine boundaries). It also has a more immediate generalization to DG, as then we can just make the projection/restriction operators DG specific and everything else stays the same in terms of the loops over the grid and copies between different blocks.

JunoRavin:
@JonathanGorard do we want to also do this same clean up in amr_block_coupled.c or do you anticipate this continuing to serve as just a baseline of functionality while we are still debugging coupled multi-fluid calculations with AMR?

I also want to have a sense of how you want to approach the coupled fluid infrastructure. Because I would think at this point you could create an amr_core_maxwell and abstract away the coupled part (since the moment_em_coupling could be structured to just loop over blocks and cells and perform the update, agnostic to the hierarchy since it's an in-place solve).

JonathanGorard:
@JunoRavin Yes, I'm also going to update the `block_coupled`, `patch` and `patch_coupled` infrastructure now to reflect the new refactoring and to pull out the generic AMR projection and restriction operators. I also agree that the coupled fluid infrastructure should be rewritten simply to include core AMR solvers for fluids and Maxwell fields, along with a corresponding modification to `moment_em_coupling` to loop over blocks: I really like that design. The newly refactored `moment_em_coupling` will be designed and written with this modification in mind; once that refactoring is complete, we should expect to revisit the AMR implementation (and the decoupling of these different things will likely make the debugging and stabilization of AMR+EM more straightforward due to the resultant increase in modularization).

JonathanGorard:
All changes discussed with @JunoRavin have now been made, and I've just confirmed on Stellar that all AMR regression tests still pass (both in serial and in parallel).

JunoRavin (approved):

ammarhakim (approved):
This is excellent work, and ready to be merged. I would like to think (after the merge), specifically about how the infra developed here can be used for the block-to-block copies in the multi-block App system. We should see there are common code we can exploit. Ideally, it would be identical, though likely that is not possible.

Issue #414, opened 2024-07-05
[DR] f-lte io and BGK Collision Restructuring


johnson452:
Design Review:
This was branched from dg_10m, and I propose to merge this new branch, dg_10m_f_lte_io-prototyping, back into dg_10m.

Description: 
This branch primarily does the following: 
-	Creates a new structure vm_lte, used within gkyl_vm_species to store moments/projection object/f_lte
-	Used in two places currently: (NEW) io of f_lte, and in the BGK collision routine to create f_lte.
-	New io flag (in the vm_species): output_f_lte 

Additions:
New file: vm_species_lte. This contains the functions: (no gkyl_... since they are private to vm)
```
// Initialize species lte object.
void vm_species_lte_init(…);

// Compute necessary moments for f-lte creation
void vm_species_lte_moms(…);

// Compute f-lte
void vm_species_lte(…);
```

With the new struct:
```
struct vm_lte {  
  struct gkyl_array *f_lte;

  struct vm_species_moment moms; // moments needed in the equilibrium

  // LTE distribution function projection object
  // also corrects the density of projected distribution function
  struct gkyl_vlasov_lte_proj_on_basis *proj_lte; 

  long self_niter; // total number of iterations correcting self collisions

  // Correction updater for insuring LTE distribution has desired LTE (n, V_drift, T/m) moments
  bool correct_all_moms; // boolean if we are correcting all the moments
  struct gkyl_vlasov_lte_correct *corr_lte; 
  gkyl_dynvec corr_stat;
  bool is_first_corr_status_write_call;
};
```

Tests:
- The regression test has been added: rt_can_pb_neut_bgk_sodshock_1x1v_p2, this outputs f_lte

Other Notes:
-	Moments for f_lte are now calculated inside the f. This is to guarantee the moments are always populated and updated before creating f_lte. (So I removed the moment calls before the bgk routines)
-	No longer accumulated for BGK since this is done by the f_lte: species_coll_mom_tm : 0

Extra checklist:
[X] Does prototype PR Work? 
[X] Unit test passes? (Checked ctest_canonical_pb_equilibrium  and ctest_corr_mj_integrated which are quite sensitive)
[X] Regression test passes?
[X] Is memory freed?

JunoRavin:
I think we just call the iteration count niter instead of self_niter.

johnson452:
> I think we just call the iteration count niter instead of self_niter.

[...]

PR #415, opened 2024-07-07
f-lte io and BGK Collision Restructuring

johnson452:
DR: [https://github.com/ammarhakim/gkylzero/issues/414](url) 

Description: 
This branch primarily does the following: 
-	Creates a new structure vm_lte, used within gkyl_vm_species to store moments/projection object/f_lte
-	Used in two places currently: (NEW) io of f_lte, and in the BGK collision routine to create f_lte.
-	New io flag (in the vm_species): output_f_lte 

Additions:
New file: vm_species_lte. This contains the functions: (no gkyl_... since they are private to vm)
```
// Initialize species lte object.
void vm_species_lte_init(…);

// Compute necessary moments for f-lte creation
void vm_species_lte_moms(…);

// Compute f-lte
void vm_species_lte(…);
```

With the new struct:
```
struct vm_lte {  
  struct gkyl_array *f_lte;

  struct vm_species_moment moms; // moments needed in the equilibrium

  // LTE distribution function projection object
  // also corrects the density of projected distribution function
  struct gkyl_vlasov_lte_proj_on_basis *proj_lte; 

  long self_niter; // total number of iterations correcting self collisions

  // Correction updater for insuring LTE distribution has desired LTE (n, V_drift, T/m) moments
  bool correct_all_moms; // boolean if we are correcting all the moments
  struct gkyl_vlasov_lte_correct *corr_lte; 
  gkyl_dynvec corr_stat;
  bool is_first_corr_status_write_call;
};
```

Tests:
- The regression test has been added: rt_can_pb_neut_bgk_sodshock_1x1v_p2, this outputs f_lte

Other Notes:
-	Moments for f_lte are now calculated inside the f. This is to guarantee the moments are always populated and updated before creating f_lte.
-	No longer used for BGK: species_coll_mom_tm : 0
-      self_niter, used to compute the total number of corrective iterations taken when creating the f_lte distribution, has been changed to niter

### DIFF
--- a/apps/gkyl_vlasov.h
+++ b/apps/gkyl_vlasov.h
@@ -151,6 +151,10 @@ struct gkyl_vlasov_species {
   // pointer to the determinant of the spatial metric
   void (*det_h)(double t, const double *xn, double *aout, void *ctx);
 
+  bool output_f_lte; // Boolean for writing out f_lte (used for calculating transport coeff.)
+  int max_iter; //Maximum number of iterations for f_lte correction
+  double iter_eps; //Maximum eps for f_lte correction
+
   // boundary conditions
   enum gkyl_species_bc_type bcx[2], bcy[2], bcz[2];
 };
@@ -287,6 +291,8 @@ struct gkyl_vlasov_stat {
   double species_lbo_coll_diff_tm[GKYL_MAX_SPECIES]; // time to compute LBO diffusion terms
   double species_coll_tm; // total time for collision updater (excluded moments)
 
+  double species_lte_tm; // time needed to compute the lte equilibrium
+
   long niter_self_bgk_corr[GKYL_MAX_SPECIES]; // number of iterations used to correct self collisions in BGK
 
   double species_bc_tm; // time to compute species BCs
@@ -422,6 +428,16 @@ void gkyl_vlasov_app_write_field(gkyl_vlasov_app* app, double tm, int frame);
  * @param frame Frame number
  */
 void gkyl_vlasov_app_write_species(gkyl_vlasov_app* app, int sidx, double tm, int frame);
+
+/**
+ * Write species data to file - for the local equilbrium.
+ * 
+ * @param app App object.
+ * @param sidx Index of species to initialize.
+ * @param tm Time-stamp
+ * @param frame Frame number
+ */
+void gkyl_vlasov_app_write_species_lte(gkyl_vlasov_app* app, int sidx, double tm, int frame);
 
 /**
  * Write species p/gamma to file.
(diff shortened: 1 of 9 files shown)


JunoRavin:
This is looking good but I have a request. I think the vm_species_lte call **should not** take the input distribution, it should take the LTE moments. 

So for example here: https://github.com/ammarhakim/gkylzero/blob/b9d4b10af134b8434639ed6f014193620bb19780/apps/vm_species_lte.c#L88

this call **should not** be inside this update. You should call vm_species_lte elsewhere in the update loop. 

This has two advantages:
1. You eliminate the need to allocate extra memory for cross-species collisions, as then the cross-moments can be passed to the vm_species_lte update method. 
2. You restore the program flow of calling bgk_moms and you can eliminate confusing comments such as here (https://github.com/ammarhakim/gkylzero/blob/b9d4b10af134b8434639ed6f014193620bb19780/apps/vlasov_forward_euler.c#L42). Instead you still call a BGK moms updater, it just diverts to the LTE moment calculator.

johnson452:
> This is looking good but I have a request. I think the vm_species_lte call **should not** take the input distribution, it should take the LTE moments.
> 
> So for example here:
> 
> https://github.com/ammarhakim/gkylzero/blob/b9d4b10af134b8434639ed6f014193620bb19780/apps/vm_species_lte.c#L88
> 
> this call **should not** be inside this update. You should call vm_species_lte_moms elsewhere in the update loop.
> 
> This has two advantages:
> 
> 1. You eliminate the need to allocate extra memory for cross-species collisions, as then the cross-moments can be passed to the vm_species_lte update method.
> 2. You restore the program flow of calling bgk_moms and you can eliminate confusing comments such as here (https://github.com/ammarhakim/gkylzero/blob/b9d4b10af134b8434639ed6f014193620bb19780/apps/vlasov_forward_euler.c#L42
>    ). Instead you still call a BGK moms updater, it just diverts to the LTE moment calculator.

Hi Jimmy, 
This was intentional. The reason is that otherwise this function always depends on calling vm_species_lte_moms(...) directly before hand. Or as you point out, eventually will need to call something like vm_species_lte_cross_moms(...), then handoff the target moms. My worry was that this leaves space for a rather nasty bug. One might accidentally use the incorrect moments by forgetting, misplacing, or changing the distribution function. Hence the moments between the vm_species_lte_moms(...) call and the vm_species_lte(...) call might have changed. 

Would there be a way to package these two calls together, while also addressing your two points, so it doesn't rely on someone using the function to just know to call the moms function? I can change it back, but thought this point was pretty important!

PR #420, opened 2024-07-09
Pass entire structs grid_inp and efit_inp to geometry instead of poin…

akashukla:
…ters to avoid memory leaks. This also required fixing some circular includes beteen gkyl_gk_geometry.h and gkyl_tok_geo/mirror_geo.h. One extra include of gkyl_gk_geometry.h has been added to gkyl_gyrokinetic.h so that the geometry input struct has the definition of the grid_inp and efit_inp structs. All gk regression tests and unit tests have been adjusted to pass the grid_inp and efit_inp structs rather than pointers to those structs. make unit regression works and  example 2x2v mirror and tokamak regression test produce the right geometry.

One idea @JunoRavin  suggested that we could use in the future, especially if the grid_inp and efit_inp structs eventually need to allocate memory is:
Make a general efit_inp struct and then make tok_efit_inp and mirror_efit_inp derived types. We could do the same for grid_inp.


JunoRavin:
One other thing to note about the general efit_inp and grid_inp structs is that it would:
1. Eliminate the need for gkyl_geometry_inp to have definitions of all possible efit and grid input quantities (so e.g., if we add stellarator capabilities we will not need further stellarator specific variants defined in this input struct). 
2. We might be able to make the header includes cleaner by defining the derived efit and grid types in their corresponding headers (so instead of including tok_geo and mirror_geo in gkyl_gk_geometry, these specific tokamak and mirror input types could live in their header files). 

But I do not think this needs to be a priority. It is just a thought about how we could reorganize the input struct to a cleaner API as we continue this work.

PR #423, opened 2024-07-11
Make 2 block - lower and middle - outer SOL multib regression test. A…

akashukla:
…lso fix one error in the physical field BCs for the full step multib test.


manauref:
The files names should be changed to have gk in it

ammarhakim:
This seems to be superseded, and we can reopen this when GK is ready to be merged in.

PR #426, opened 2024-07-16
Gk g0 app gapremoval

akashukla:
This branch's removes the gaps between tokamak blocks at the X-point.

Specifically these changes made in this branch:
1. Tokamak geometry sets the end points for each block in a way that is consistent with the adjacent blocks.
2. Tokamak root finding allows a small tolerance for inexact roots to ensure a root is found at the X-point. This tolerance is very small : when solving for the roots of a quadratic we treat a negative discriminant with absolute value <1e-20 as 0.
3. Tokamak tangent and dual vectors are now calculated using analytical formulas from the cylindrical coordinate representation consistently with the other metric quantities and jacobian.
4. The geometry object now calculates and stores the normal vectors (so we can compare at block boundaries)

Checks: 
1. step regression test runs and produces correct results in serial and in parallel on gpus and cpus. Mirror and LAPD regression test also produce the correct normal vectors.
2. Does not break make check or make unit regression. Issues with compilation of unit tests are pre-existing issues associated with initialization of variable sized arrays (unrelated to this PR).
3. Step regression test is valgrind clean

Note: Automated finding of the X-point and psisep within Gkeyll is not part of this PR. We need to do this but it will be part of a separate work/PR. Right now the X-point finding is done externally.


JunoRavin (approved):

ammarhakim (approved):
This PR is ready to merge. I would like to go over the entire geometry creation sequence of steps at some point, but that need not hold up this merge.

Issue #418, opened 2024-07-09
[DR] Refactor of SR Vlasov for energy conservation and improved robustness of correction algorithm


JunoRavin:
This design review encompasses the two key improvements contained in the branch: https://github.com/ammarhakim/gkylzero/tree/dg_10m_sr_refactor

1. In the attached pdf I outline the proof but the essential idea is that before, the SR Vlasov solve was not energy conserving. The original solver was formulated to separately compute gamma = sqrt(1 + p^2) (the particle Lorentz boost factor) and p/gamma (the relativistic velocity) **independently**. However, our work on Hamiltonian systems demonstrates that these two quantities are not independent and to properly have an energy conserving scheme one must 

- First construct gamma to be continuous in the DG basis (using eval_on_nodes)
- Anywhere p/gamma is needed (streaming term, Lorentz force, and the current density moment for coupling to Maxwell's equations), this expansion **must** be computed as grad_p gamma. 

I have verified that with this construction our scheme is now energy conserving using the relativistic two-stream instability, and we obtain third order accuracy in the convergence of the energy due to the SSP RK3 method employed for time-stepping. See attached figure from the 1x1v regression test. 

2. We were previously having difficulty with the overall robustness of the LTE moment calculation and correction algorithm in relativity, principally due to the fact that in regions of strong gradients the drift velocity may be super-luminal (even if the cell average is less than the speed of light). Constraining the drift velocity to be well-behaved was even harder when trying to compute |V_drift|^2 for computing the Lorentz boost factor and the rest-frame density. This difficulty can be clearly seen in the main branch at the time of this issue (07/09/2024) where the Vlasov SR BGK Sod Shock test **fails** for Nx > 128 due to the initial large flow generated by the pressure gradient. 

**To solve this issue, I have changed the relativistic solver to utilize the spatial component of the bulk four-velocity in the LTE moment calculator and correction algorithm**. The spatial component of the bulk four-velocity is guaranteed to be well behaved as it does not have to be bounded from above, and computing the needed Lorentz boost factor from this quantity is also guaranteed to be positive definite. To achieve this refactor the steps are now:

- Compute V_drift as M1i/M0 using weak division (as before). 
- Check if (1 - |V_drift|^2) > 0 at Gauss-Legendre quadrature points; if it is, utilized the basis_sqrt function which evaluates the square root at Gauss-Legendre points and projects onto the modal basis to construct the basis expansion of 1/GammaV = sqrt(1 - |V_drift|^2)
[...]

Issue #386, opened 2024-06-14
[DR] Extensions to Vlasov and PKPM in the dg_10m branch. 


JunoRavin:
A number of extensions to the Vlasov and PKPM system have been explored in the https://github.com/ammarhakim/gkylzero/tree/dg_10m branch that have yielded new functionality and improvements to robustness and are ready for design review. 

The new features are: 
1. Both the DG fluids functionality in the Vlasov app and the PKPM app now support a first-order operator split + semi-implicit source solve as described in Wang et al. 2020 JCP (https://ammar-hakim.org/sj/_static/files/JCP-2020-Wang.pdf). This algorithm has been extended to our modal DG formalism, in which the matrix for how each modal DG coefficient evolves is constructed following the equations for the linear system (analogous to weak division or the primitive moment solve for the conservative LBO), and then the matrix is inverted with our batched linear solve routines. The standard plasma-beach problem (https://ammar-hakim.org/sj/je/je8/je8-plasmabeach.html) runs with a time step set by the speed of light instead of the plasma frequency in the high density region. The PKPM version has a slightly different cutoff due to the presence of a background Bx (to define the vparallel direction) but nevertheless the launched plasma wave becomes evanescent as expected and the stable time step restriction is relaxed. The new Alfven soliton regression test based on Kakutani & Ono 1969 also stably runs with dx ~ 32 di (where di is the ion inertial length) and dt ~ omega_ci^{-1}/4 (where omega_ci is the ion cyclotron frequency). 
2. The canonical Poisson bracket infrastructure has been extended so that the Hamiltonian is constructed utilizing the form H = 1/2 g^ij w_i w_j, where g^ij is the contravariant components of the metric tensor. This formalism allows simulations of motion on the surface of a sphere, and the vlasov_lte_proj_on_basis routine has been extended to handle this functionality (including the necessary metric components in the exponential). The vlasov_lte_moments routine has also been extended to obtain the correct pressure given the input pressure tensor. 
3. The same first-order operator split which is utilized for the implicit fluid-EM coupling can also be utilized for an implicit BGK collision operator. Simulations which take time steps 1k-10k times larger than the collisional time have been performed. 
4. The PKPM fluid system now utilizes the wv_eqn ten_moment object to perform the Roe solve at the interface (at quadrature points in higher dimensions). This switch has reduced the oscillations significantly in a number of production problems, and simulations which were previously unstable, such as a PKPM firehose instability simulation, now stably run. 
[...]

PR #429, opened 2024-07-21
Refactors and improvements to PKPM, SR Vlasov, and Canonical Poisson Bracket systems

JunoRavin:
This set of commits brings to a conclusion a systematic set of developments on improving the robustness and accuracy of a number of systems within Gkeyll, including the parallel-kinetic-perpendicular-moment model, the special relativistic Vlasov-Maxwell model, and the canonical Poisson Bracket model. 

A number of design reviews are encompassed in this PR including https://github.com/ammarhakim/gkylzero/issues/418, https://github.com/ammarhakim/gkylzero/issues/414, https://github.com/ammarhakim/gkylzero/issues/400, and https://github.com/ammarhakim/gkylzero/issues/386

We summarize here the important changes within this PR, with further details provided by the relevant DRs: 
1. First order operator splits are now supported by Vlasov and PKPM, with the ability to treat implicit BGK collisions in the former and implicit EM source terms in the latter. 
2. PKPM now utilizes wv_ten_moment to do a Roe solve at interfaces, which was found to reduce dispersive errors in a number of nonlinear simulations. 
3. The PKPM "penalization" factor, which by default comes from the 10 moment Roe solve (the jump in the fluctuations determined from the Roe solve) switches back to Lax fluxes under conditions in which the Roe solve would fail (such as when using absorbing boundary conditions and examining plasma sheath problems, since the ghost cell values of the density are identically 0).
4. The canonical Poisson bracket infrastructure has been extended so that the Hamiltonian is constructed utilizing the form H = 1/2 g^ij w_i w_j, where g^ij is the contravariant components of the metric tensor.
5. The formalism of setting up wv_eqn objects on device has been extended to wv_maxwell and wv_ten_moment for DG fluids or PKPM simulations on GPUs (either for the Roe solve or the post hoc characteristic limiter). 
6. SR Vlasov now is energy conserving via a continuous projection of the particle Lorentz boost factor and determining the relativistic velocity from the momentum derivative of this factor. 
7. SR Vlasov now utilizes the spatial component of the four-velocity for constructing the LTE distribution and correcting the moments, which does not have the same positivity and robustness issues from superluminal velocities. 
8. The project LTE updater utilizes new nodal-to-modal transformations on GPUS, utilizing an optimize matrix-matrix routine which directly operates on gkyl_array data. 
9. We can now write out the equivalent LTE distribution for calculating delta f and transport coefficients. 

As part of these refactors, input files were commensurately changed to utilize this new functionality. SR simulations now all utilize the project LTE function (which more robustly handles the normalization factors of the LTE distribution without dealing with finite precision issues associated with evaluating modified Bessel functions) and can take multiple input projection functions for performing streaming instability calculations. PKPM simulations by default utilize the implicit solver for the EM sources, but can set the app level .use_explicit_source flag to utilize fully explicit sources if a user knows they are resolving these quantities in time (since then the scheme is higher order and conserves energy better). We write out the LTE distribution with the species level .output_f_lte flag. 

There is more work to be done, but this branch grew to encompass a large enough number of commits that this code must be reviewed as is before further work should be done.


ammarhakim:
@JunoRavin 

A few generic comments: going forward we should strongly discourage such huge PRs. This has 180 commits, it seems. Though, most of the work here was looked over in various DRs we should still break up our PRs into smaller, easy to review chunks.

I would like to understand your (7) a little better. I recommend submitting a small DR for it where you describe in equations what you did.

ammarhakim:
As this merge is into main, I would recommend checking if we can continue to use G2 Vlasov with this branch. I think it is important to track down the issues we are seeing in G2/G0-Main so we can help those who are running this combo for their work.

ammarhakim (approved):
This PR is ready to merge with the latest G2 main fixes. However, let not have such a huge PR again!

johnson452 (approved):

JonathanGorard (approved):

PR #430, opened 2024-07-23
gk-g0-app-neut

tnbernard:
The ionization model has been updated to allow for ionizations when the electron temperature is less than the ionization energy. A model of charge exchange based on [Bernard et al. PoP 2022](https://doi.org/10.1063/5.0087131) has also been added. Conservation properties have been thoroughly tested and verified in simple and shaped tokamak geometries by @akashukla.

Previously, the ionization term for electron species assuming that the primary and secondary electrons had the same average temperature: 

Screenshot 2024-06-25 at 4 34 24 PM

Now we assume different temperatures for the primary and secondary electrons with a model that conserves energy: 

Screenshot 2024-06-25 at 4 35 47 PM

where Screenshot 2024-06-25 at 4 36 30 PM with

Screenshot 2024-06-25 at 4 37 03 PM ,  
Screenshot 2024-06-25 at 4 37 54 PM. 

The charge exchange model includes changes to zero/dg_cx.c, zero/dg_cx_cu.cu, zero/gkyl_dg_cx.h, and zero/gkyl_dg_cx_priv.h. Kernels for calculating the reaction rate are contained in kernels/neutral/gkyl_dg_cx_kernels.h. The app was also edited in the files apps/gk_species_react.c and apps/gk_neut_species_react.c. 

The charge exchange collision operator is given by
Screenshot 2024-06-25 at 4 45 22 PM
with a relative velocity 
Screenshot 2024-06-25 at 4 45 53 PM. 

The calculation of the relative velocity is done assuming Vlasov Cartesian components for the drift velocity u. That is, the u_parallel from GK species is projected onto the Vlasov components of the vector **u**. This is accomplished using the Cartesian components of the unit b vector which are stored as "bcart" the geometry object. ("bcart" is also used in the recombination collision term for projecting the ion maxwellian on the gk_neut_species grid.) The cross sections are modeled via fitting functions and currently only for hydrogen, deuterium, helium, and neon.


JunoRavin (approved):
This is fantastic work and I see no reason why we shouldn't merge this immediately. Thanks to @tnbernard and @akashukla for such careful and thorough testing.

apps/gk_neut_species_react.c
```diff
@@ -72,60 +85,99 @@ gk_neut_species_react_cross_init(struct gkyl_gyrokinetic_app *app, struct gk_neu
...
+        .grid = &s->grid,
+	.cbasis = &app->confBasis,
+	.pbasis_gk = &app->basis,
+	.pbasis_vl = &app->neut_basis,
+	.conf_rng = &app->local,
+	.conf_rng_ext = &app->local_ext,
+	.phase_rng = &s->local,
+	.mass_ion = react->react_type[i].ion_mass,
+	.mass_neut = react->react_type[i].partner_mass,
+	.type_ion = GKYL_ION_D,
```
akashukla, line 102: Should this really always be GKYL_ION_D? Or should it be based on the input file since now we are supporting neon, etc

akashukla (commented):
Overall, looks good to me! Just some small things:

I went ahead and made one change in calc_metric.c so the correct commit gets used for the sign of the duals (used in calculating bcart for the CX operator). 

I had one other question/comment about how GKYL_ION_D always seems to be used in the C operator in gk_Species_react.c

apps/gk_neut_species_react.c
tnbernard: Fixed in latest commit.

akashukla (approved):

PR #431, opened 2024-07-23
gk-g0-emission-spectrum

kltrbrds:
A full implementation (CPU and GPU) of emission boundary conditions into the gk-g0-app branch (as opposed to the previous partial implementation designed to work solely through g2-g0) is in https://github.com/ammarhakim/gkylzero/tree/gk-g0-emission-spectrum.

The majority of the new proposed infrastructure is self-contained within seven methods. The main app and updater are performed by `vm_species_emission`, `bc_emission_spectrum`, and `bc_emission_elastic`. The first is the g0 app plugin for the emission, the second is the inelastic emission spectrum calculation, and the final is the elastic backscattering calculation. The emission model being implemented matches what is described in Section 4.3 of https://vtechworks.lib.vt.edu/handle/10919/118137, with the modification that the weighted average of the secondary electron yield is done over the boundary flux instead of the skin cell distribution.

Additionally, there are the model methods, `bc_emission`, `spectrum_model`, `yield_model`, and `elastic_model`. These are the initializer and generic model types for each element of the emission functionality, analogous to the `wv_eqn` type in the fluid solver. Different models wrap the generic type, allowing specific model parameters to be initialized without requiring the allocation of large structs directly in the input file. `bc_emission` also includes preset initializers for the three simulation cases included in the above thesis (copper, oxidized lithium, and clean lithium), which allow a material to be selected without the user providing all the required fitting parameters and model information.

The major global change done to the Vlasov-Maxwell app is the modification of the BCs to match what is done in gyrokinetics. Boundary conditions are now passed in the form:
```
.bcx = {
  .lower = { .type = GKYL_SPECIES_ABSORB, },
  .upper = { .type = GKYL_SPECIES_EMISSION,
             .aux_ctx = &bc_ctx, },
},
```
This allows for specification of the auxiliary structures required for various emission models.

Because the impacting species causing emission and the species being emitted are not necessarily the same (such as with ion-impact secondary electron emission), an additional cross-species initialization method is called for the new `vm_species_emission` after all species have initialized. Further, `gkyl_vlasov_app_apply_ic` is modified to compute the boundary condition after IC's are applied to all species, due to the same constraint. Issues were caused in the same method by the fact that IC infrastructure for the `fixed_func` BC assumes `bc_basic` boundary conditions are used regardless of what is specified, so a check of `bc_type` was added there to prevent crashes when emission BCs are used.

High-emission simulations show it is sometimes desirable to scale the emission with time, to  prevent too severe of an initial emitted density gradient from developing in the skin cell. This requires a minor modification to the `forward_euler` method in the app, passing `tcurr` as an additional argument to `vm_species_apply_bc`.

Boundary fluxes were modified to store boundary grids, ranges, and an array of the flux at each edge. These are used extensively by the emission. Previously, the flux was only being temporarily stored in the ghost cells of the full RHS array.

A modification was made to the BGK collision operator as part of this work, allowing relaxation to initial temperature instead of an updated local temperature with specification of the flag `fixed_temp_relax`. This is very useful for sheath work, where maintaining a steady temperature in the presheath is difficult. A regression test `rt_vlasov_sheath_bgk_1x1v_p2` was added to test this functionality.

```
.collisions =  {
      .collision_id = GKYL_BGK_COLLISIONS,
      .self_nu = evalElcNu,
      .ctx = &ctx,
      .fixed_temp_relax = true,
    },
```

The regression test `rt_vlasov_emission_spectrum_1x1v_p2` was added to demonstrate functionality of the emission BC. The revised input file setup gives both means of creating the BC (with the full specification commented out for user reference):
```
  char in_species[1][128] = { "elc" };

  // Copper preset for emission BC where models and parameters are chosen automatically              
  struct gkyl_bc_emission_ctx *bc_ctx = gkyl_bc_emission_secondary_electron_copper_new(ctx.num_emission_species, 0.0, in_species, app_args.use_gpu);

  // Full specification of the BC with user-defined models and parameters                            
  // Uncomment and use this if no material preset is available                                       
  /* struct gkyl_spectrum_model *spectrum_model[1]; */
  /* spectrum_model[0] = gkyl_spectrum_chung_everhart_new(ctx.q0, ctx.phi, app_args.use_gpu); */
  /* struct gkyl_yield_model *yield_model[1]; */
  /* yield_model[0] = gkyl_yield_furman_pivi_new(ctx.q0, ctx.deltahat_ts, ctx.Ehat_ts, ctx.t1, ctx.t2, ctx.t3, ctx.t4, ctx.s, app_args.use_gpu); */
  /* struct gkyl_elastic_model *elastic_model = gkyl_elastic_furman_pivi_new(ctx.q0, ctx.P1_inf, ctx.P1_hat, ctx.E_hat, ctx.W, ctx.p, app_args.use_gpu); */
  /* struct gkyl_bc_emission_ctx *bc_ctx = gkyl_bc_emission_new(ctx.num_emission_species, 0.0, true, spectrum_model, yield_model, elastic_model, in_species); */

```
Deallocations are done through release methods called at the end of the input file:
```
  /* for (int i=0; i<ctx.num_emission_species; ++i) { */
  /*   gkyl_spectrum_model_release(spectrum_model[i]); */
  /*   gkyl_yield_model_release(yield_model[i]); */
  /* } */
  /* gkyl_elastic_model_release(elastic_model); */
  gkyl_bc_emission_release(bc_ctx);
```
Individual model release calls are handled internally (as the allocations are) if a preset is used.


apps/gkyl_vlasov_priv.h
```diff
@@ -611,6 +669,19 @@ void vm_species_moment_calc(const struct vm_species_moment *sm,
 void vm_species_moment_release(const struct gkyl_vlasov_app *app,
   const struct vm_species_moment *sm);
 
+/** vm_species_emission API */
```
JunoRavin, line 672: Let's add the corresponding comments for the emission API here like the other private methods.

apps/gkyl_vlasov_priv.h
```diff
@@ -737,7 +821,7 @@ void vm_species_bgk_release(const struct gkyl_vlasov_app *app, const struct vm_b
  *
  * @param app Vlasov app object
  * @param s Species object 
- * @param bflux Species boundary flux object
+ * @param bflu Sxpecies boundary flux object
```
JunoRavin, line 824: Typo in comment.

apps/vlasov.c
```diff
@@ -1092,7 +1102,7 @@ comm_reduce_app_stat(const gkyl_vlasov_app* app,
   };
 
   int64_t l_red_global[L_END];
-  gkyl_comm_allreduce(app->comm, GKYL_INT_64, GKYL_MAX, L_END, l_red, l_red_global);
+  gkyl_comm_allreduce_host(app->comm, GKYL_INT_64, GKYL_MAX, L_END, l_red, l_red_global);
```
JunoRavin, line 1105: Making a note here in this PR that this fixes Vlasov on GPUs in gk-g0-app, though we are currently refactoring the comm object as part of the multi-block work here: https://github.com/ammarhakim/gkylzero/tree/multib-app so that there will not be "host" methods.

JunoRavin (approved):

ammarhakim (approved):
This looks good and is ready to merge.

PR #434, opened 2024-07-25
Renaming Gkyell Lua executable, getting it ready for use in testing

ammarhakim:
I have modified the name of the xglua exec to glua. Further, a new install target was added to install this exec. The goal is to get some Lua exec installed for assisting in automated testing. The latter remains to be done, but this is the first step for that.


ammarhakim:
Did not mean to merge into main.

Issue #390, opened 2024-06-17
[DR] Refactor the handling of geometric sources in GR (and other things)


JonathanGorard:
Currently, source terms in the moment app are handled through the `gkyl_moment_em_coupling` struct, which populates a forcing vector based on specified couplings between (multiple) fluid species and electromagnetic fields. However, this design is clearly inadequate for various projects that we are currently pursuing and intend to pursue in the near future, since in many cases (such as general relativity), there may exist geometric sources that are _inherent_ to a single fluid species, even in the absence of any other fluids or fields that it is coupled to. The same would also be true if we ever intend to support reactive fluid species, viscous hydrodynamics, etc. in Gkeyll.

One possible design, which was initially discussed, would be to implement a `gkyl_moment_spacetime_coupling`, in the same kind of vein as the existing `gkyl_moment_em_coupling`. However, it seems to me that a more minimalistic and elegant formulation (not to mention more general) would be to introduce a `gkyl_<species>_source` function that computes the forcing vector for a given fluid species, in the same way that we currently have a `gkyl_<species>_flux` function for computing that fluid's flux vector. For all fluid species currently implemented (with the exception of GR Euler), which are governed by homogeneous conservation laws in the absence of other fluids and fields, this forcing vector would default to being the zero vector. Then, we could introduce a pair of new function pointers to the `gkyl_wv_eqn` object: the first called `source_func` (which would default to being a "do nothing" function) which solves the forcing part of the problem using whatever combination of explicit SSP RK3, semi-implicit, fully implicit, etc. is desired; the second called `split_func` (which would simply default to calling the hyperbolic updater in the homogeneous case) which performs whatever kind of operator splitting is desired (Strang splitting, Godunov splitting, etc., in which the hyperbolic updater and the source updater specified via the `source_func` pointer may be called in any order, and with any specified combination of time-steps). This would necessitate adding only two additional lines in the `gkyl_wv_<species>_inew` method for each fluid species (which, as mentioned, could be populated by default so that no existing code would need to be changed):

```
struct gkyl_wv_eqn*
gkyl_wv_<species>_inew(const struct gkyl_wv_<species>_inp *inp)
{
  struct wv_<species> *<species> = gkyl_malloc(sizeof(struct wv_<species>));

  ...

  <species>->eqn.source_func = source_func;
  <species>->eqn.update_func = update_func;

  ...

  return &<species>->eqn;
}
```

[...]

PR #437, opened 2024-07-29
Refactor of moment-EM coupling into separate implicit and explicit source solvers

JonathanGorard:
This PR implements many of the modifications discussed within Design Review #390 - it splits up the previously monolithic `moment_em_coupling.c` into separate implicit and explicit source solvers (located in `sources_implicit.c` and `sources_explicit.c`, respectively), allowing for different families of source terms to be handled by different classes of ODE solver, and with potentially different operator splittings, as required. It also incorporates the `source_func` function pointer into the `gkyl_wv_eqn` object, allowing all fluid species to specify their own forcing vectors which can then be solved explicitly (using e.g. SSP-RK3) or implicitly (using e.g. time-centered Crank-Nicolson), as appropriate. We have included a new reactive Euler fluid species, with an associated ZND detonation test (with a known analytic solution for the von Neumann spike), for testing and validating this new `source_func` system for specifying per-species source terms, which we intend to use for geometric source terms (e.g. in the general relativistic Euler solver), number density and temperature source terms, as well as many other things, moving forward. Broadly speaking, the overarching idea is to enforce a separation between per-species source terms (exposed to the user through the `source_func` pointer) and inter-species source terms (exposed to the user through the traditional `gkyl_moment_em_coupling` structure).

As an illustration of its generality, this PR also adds a couple of new classes of source terms (in addition to the reactive ones used for numerical validation), including frictional source terms for two-species fluid simulations, and geometric source terms for the expanding box formalism. In the latter case, there is a remaining bug wherein the momentum terms are not correctly incorporated into the pressure tensor contributions, although this can easily be fixed. Finally, this PR also folds in some of the Braginskii functionality from @JunoRavin's old branch, and sets up the requisite time-steppers within the new source refactor to integrate these terms correctly.


JonathanGorard:
@JunoRavin has pointed out that the inclusion of the private header `gkyl_moment_em_coupling_priv.h` within the `gkyl_source_explicit.h` and `gkyl_source_implicit.h` header files breaks layering by exposing the internal details of the private `moment_em_coupling` struct via the user-facing API. Instead, we will adopt the same approach used within the rest of the file, in which the `moment_em_coupling` struct is forward-declared within the header files in question, thus ensuring that the private header only needs to be included in the source file containing the implementations.

JunoRavin (approved):
I think this is ready to merge!

Issue #384, opened 2024-06-12
[DR] Twist-shift boundary conditions


manauref:
[UPDATE]: see latest design in my comment on July 31st below.

These boundary conditions are meant for 3D gyrokinetics (though in fact they were also applied to fluids in other codes), and consist of 
```
f(x, y, z_max) = f(x,y-S,z_min)
```
or viceversa
```
f(x, y, z_min) = f(x,y+S,z_max)
```
described, along with its DG algorithm, in M. Francisquez, et al. CPC 298 (2024) 109109 (see eqn 9).

As given in eqn 30 of that paper, the distribution in a the target cell (i,j_tar) is given by a linear combination of mat-vec multiplies including contributions from N_{do}(i) donor cells
```
f_{i,j_tar} = sum_{q=1}^{N_{do}(i)} M_q f_{do,i,j_q}
```
So implementing this BC consists of two things:
1. Constructing the M_q matrices (during initialization).
2. Performing these matrix vector multiplies, and adding their contributions.

We intend to do this building up the capabilities of zero/bc_twistshift.c, which was created last year just to do 2, assuming 1 was done in g2. We'll mention some initial design intentions about each of these below, and edit it as things evolve:

Constructing M_q matrices
---------------------------

We'll store a flat list of donor matrices, created with
```
  up->matsdo_ho = gkyl_nmat_new(total_donor_mats, basis->num_basis, basis->num_basis);
```
for which we'll need to know the total number of donors (matrices or cells, `total_donor_mats`). This is computed after having a (flat) list of the donor cell indices (`cellsDo` in g2, currently `cells_do` in g0). To be precise, `cells_do`, is a flattened 4D array: for each (i,j) location we store the x-y indices of N_do(i) donors.

After knowing who the donors are for each target, we proceed to populate their elements. This involves a sequence of sub-cell integrals, which we have to craft based on a number of 1D projections (onto a DG basis), root finding, and calling 1 of 3 kernels. In g2 this part of the code made heavy use of Lua tables to store scalars, arrays, functions and maybe even strings; we'll have to figure out how to write the C equivalent without such tables, though in some places we may replace similar those data objects with C structs.

The matrices are computed on the host, and are later copied to the GPU with
```
  gkyl_nmat_copy(up->matsdo, up->matsdo_ho);
```

Performing mat-vec multiplies and adding their contributions
-------------------------------------------------------------
[...]

PR #438, opened 2024-07-31
Twist-shift BC updater

manauref:
This is an implementation of the twist-shift (TS) boundary condition (BC) for 5D gyrokinetics. The original algorithms is described in M. Francisquez, et al. CPC 298 (2024) 109109, but we (@akashukla, @jRoeltgen and I) have implemented a simplified version of it, see issue #384  (in particular the comment on July 31 2024), and the comments at the top of zero/bc_twistshift.c. I reproduce a couple of these comments here:
1. Unlike the procedures described in M. Francisquez, et al. CPC 298 (2024) 109109, all subcell integrals are now done with variable y limits. This is possible once we realize that figure 4 is not drawn accurately; the blue lines should be separated by Delta y at all points.
2. This updater only works on 5D distributions. Likely only minor changes are needed to make it work in other dimensions.
3. 99% of the code is written to support a BC, a shift and shear in any direction. Maybe the only thing that needs to change is the permutted range and its use.

There is a unit test, which runs on CPUs and GPUs, and with which I've been able to reproduce figures 6, 9 and 11 from the paper. I've also confirmed the conversation property shown in figure 16a.

The way this works is:
1. One must first apply regular periodicity (outside of this updater).
2. Pass the distribution to this updater, and it will twist-shift the distribution that was previously in the ghost cell (in-place).

See issue #384 for possible improvements, which we should only pursue after some simulations have been done.


tnbernard:
I would like to see a 5D 'clopen' simulation tested before merging, and I am happy to help with this.

manauref:
My logic was to PR the updater alone here. The clopen-specific stuff will come in a separate PR (since that will mostly involve changes to the app).

If you guys want clopen to happen here that's fine by me, but it'll make this PR bigger.

tnbernard:
That makes sense to wait for clopen. What about Fig. 17 in your 2024 paper? Are you able to reproduce that?

manauref:
We are not able to reproduce figure 17 because that requires a delta-f
solver we had in g2 but not in g0. It would take a bit of work (probably a
new app) to do that.

tnbernard (approved):

ammarhakim (approved):
This is ready to merge.

PR #441, opened 2024-08-07
gk-g0-app emission header fix

kltrbrds:
An inclusion call to the private header `gkyl_bc_emission_elastic_priv` in the public header of the same object was causing build issues. This inclusion call has been moved to `bc_emission_elastic`.


ammarhakim (approved):
Thanks for fixing this quickly. I tested the change and it fixes the build.

JunoRavin (approved):

PR #442, opened 2024-08-07
Simplify logic setting CUDSS RPATH in Makefile. Add CUDSS to Makefile…

manauref:
….sample as it was missing. Change cuda version in stellar-amd jobscript to be consistent with the build scripts.


JunoRavin (approved):

ammarhakim (approved):
Ready to be merged.

PR #447, opened 2024-08-19
Adding mixture formalisms, and other minor updates to moments and source coupling

JonathanGorard:
The major functionality change in the `moment_source_extensions` branch is the addition of mixture formalism support (for both full 5-moment mixtures and isothermal mixtures) to the moment app. This change is not entirely trivial, since it required modifying the function signatures of `wv_rotate_to_local`, `wv_rotate_to_global` and `wv_bc_func_t` to ensure that these functions are passed pointers to the base `gkyl_wv_eqn` struct (since otherwise there is no general way to determine how many equations there are within the system when applying boundary conditions or performing local<->global rotations, which is a problem for mixture formalisms because the total number of equations is dependent upon the number of components present in the mixture). These mixture formalisms are based on a conservation law form of the provably thermodynamically-consistent volume fraction-based formalism of [Allaire et al.](http://www.cmap.polytechnique.fr/~allaire/ack.pdf), and have been validated against the 1-dimensional shock tube tests of [Fedkiw et al.](https://www.sciencedirect.com/science/article/pii/S0021999199962368) and the 2-dimensional bubble collapse problems of [Terashima and Tryggvason](https://www.ljll.fr/~frey/papers/applications/Terashima%20H.,%20A%20front-tracking%20ghost-ﬂuid%20method%20for%20ﬂuid%20interfaces%20in%20compressible%20ﬂows.pdf). The inclusion of these formalisms into g0 will make it possible to run simulations involving transport between electrons and ionic mixtures, and to perform comparisons of self-similar diffusion in ionic mixtures against self-similar friction in multi-fluid systems.

The automatic moment and Vlasov regression systems show no changes in correctness or performance as a result of this change, and @JunoRavin has pointed out that, thanks to the additional level of indirection provided by the `gkyl_wv_eqn_rotate_to_local` and `gkyl_wv_eqn_rotate_to_global` functions (which already accepted a pointer to the `gkyl_wv_eqn` struct as their first argument), there should also be no regressions in PKPM functionality, although this should be checked.

Other smaller modifications contained in this PR include fixes to the source solver system so that moment-EM coupling objects are automatically created whenever friction, transport, geometric, etc. sources are present, irrespective of the presence of an electromagnetic field (this allows neutral Rayleigh-Taylor instability tests to run correctly); minor modifications to the explicit frictional solver to allow momentum transport between any combination of 5-moment and isothermal systems; the linkage of mixture formalisms into the AMR subsystem; the movement of all AMR regression tests to their own separate `amr_regression` directory (with corresponding modifications to the Makefile to accommodate this); and cleanup of some of the transport and Braginskii regression tests.


JonathanGorard:
Problem with the Mac build. Closing until I track this down.

JonathanGorard:
Apologies for that - it seems that I missed a cluster of compiler warnings about the modified function signatures. Should be fixed now.

amr/amr_block.c
```diff
@@ -1,4 +1,5 @@
 #include <gkyl_amr_block_priv.h>
+#include <gkyl_wv_euler_mixture_priv.h>
```
JunoRavin, line 2: Why is the euler mixture private header needed here?

apps/moment.c
```diff
@@ -149,10 +152,10 @@ gkyl_moment_app_new(struct gkyl_moment *mom)
 
   // check if we should update sources
   app->update_sources = 0;
-  if (app->has_field && ns>0) {
+  //if (app->has_field && ns>0) {
```
JunoRavin, line 155: Maybe we get rid of these commented out lines and we just note in the comments that we are always running with sources?

zero/sources_explicit.c
```diff
@@ -105,32 +106,73 @@ explicit_frictional_source_update(const gkyl_moment_em_coupling* mom_em, double
...
 
-    for (int i = 0; i < 5; i++) {
+    if (mom_em->param[0].type == GKYL_EQN_EULER) {
+      elc_num_equations = 5;
+    }
+    else if (mom_em->param[0].type == GKYL_EQN_ISO_EULER) {
+      elc_num_equations = 4;
+    }
+
+    double *f_elc_new = gkyl_malloc(sizeof(double[elc_num_equations]));
```
JunoRavin, line 119: We need to avoid these mallocs here. Otherwise our performance can be impacted. We should just make these arrays on the stack since they're small.

zero/moment_em_coupling.c
```diff
@@ -148,11 +159,30 @@ gkyl_moment_em_coupling_explicit_advance(const gkyl_moment_em_coupling* mom_em,
       app_accel_s[i] = gkyl_array_cfetch(app_accel[i], cell_idx);
     }
 
-    double *em_arr = gkyl_array_fetch(em, cell_idx);
-    const double *app_current_arr = gkyl_array_cfetch(app_current, cell_idx);
-    const double *app_current1_arr = gkyl_array_cfetch(app_current1, cell_idx);
-    const double *app_current2_arr = gkyl_array_cfetch(app_current2, cell_idx);
-    const double *ext_em_arr = gkyl_array_cfetch(ext_em, cell_idx);
+    double *em_arr = 0;
```
JunoRavin, line 162: I think we should make these lines ternary operators like we have elsewhere in the code:

double *em_arr = em ? gkyl_array_fetch(em, cell_idx) : 0;

I'm not normally one for compactifying lines, but we do this everywhere else in the code and it really is quite nice.

JunoRavin (approved):
You've addressed all my comments and this is ready to merge!

ammarhakim (approved):
Ready to merge

Issue #448, opened 2024-08-19
[DR] Cubic Representaion of Poloidal Flux for Magnetic Field Calculation


akashukla:
We can now use the cubic interpolator in dg_basis_ops.c to represent the poloidal flux, psi, read from EFIT files. Using a cubic representation of psi makes it easier to calculate the magnetic field since the second derivatives and cross derivatives are continuous. Previously, with our quadratic representation, we has to use recovery to achieve continuity in B.

Field line tracing can also be done using the cubics by passing use_cubics = true in the input file to the geometry grid input. However, it is extremely slow (200x slower for the STEP regression test)  at present and is not recommended.

The prototyping work has been done in this branch: https://github.com/ammarhakim/gkylzero/tree/multib-app_gk_cubicbmag . The changes made are:

1. Geometry methods which use EFIT now calculate B(R,Z) using the cubic representation
2.  There is an option to do field line tracing using the cubic representation using Gkeyll's builtin polynomial root finder. The X-point finder now uses this same option as it is important to be consistent.
3. functions to evaluate the second derivatives of the cubic interpolator were added to the cubic interpolator in dg_basis_ops.c

This change also makes it easy to calculate the magnetic field at R=0 using L'hopital's rule as

$$B_R = \lim_{R \to 0} \frac{1}{R} \frac{\partial \psi}{\partial Z}  = \lim_{R \to 0} \frac{\partial^2 \psi}{\partial Z \partial R},$$

$$B_Z = - \lim_{R \to 0} \frac{1}{R} \frac{\partial \psi}{\partial R} = - \lim_{R \to 0} \frac{\partial^2 \psi}{\partial R^2}.$$

Tests demonstrate that we can go to R= 0 using this method. 

Note: Our old method had a consistency issue: psi(R,Z) used for field line tracing was different than the recovered psi(R,Z) used for magnetic field calculation. Our new method when used as recommended (cubics for magnetic field calculation, and quadratics for field line tracing), has the same consistency issue.  Using the cubics for the field line tracing as well would resolve the consistency issue, but this will be impractical until the root finders are sped up.

akashukla:
Here are some plots of the wham grid and magnetic field : Tracing using quadratics and field using cubics.
![wham_to_axis_bmag](https://github.com/user-attachments/assets/9c869a4b-ad15-4d3c-83ed-6130876b9e97)
[...]

PR #449, opened 2024-08-20
Multib app gk cubicbmag

akashukla:
This PR is associated with Issue (DR) #448  . For geometries read from EFIT, B is now calculated using a cubic representation of the poloidal flux, psi, rather than with a quadratic representation + recovery. There is also now an option to use the cubic representation for field line tracing, although the quadratic representation is used by default since it is faster.

Note about a lot of the cleanup in this PR:

The reason we no longer need the bmag kernels and no longer need F(psi) to be stored by efit updater is that now the calculation of B(R,Z), which uses this EFIT-given quantity F(psi) along with psi(R,Z), happens entirely within the efit updater. This was made very simple by the cubics.

Now the calc_bmag updater simply uses mapc2p given by  the field line tracer along with  B(R,Z) given by  the efit updater to give us B on the computational grid.

Much of the cleanup in this branch is associated with this change. It used to be a lot of work (done by  the calc_bmag updater) to calculate B(R,Z). Now B(R,Z) is calculated by the efit updater immediately after reading the efit file.


zero/gkyl_mirror_geo.h
```diff
@@ -80,6 +89,9 @@ struct gkyl_mirror_geo_grid_inp {
   plate_func plate_func_lower;
   plate_func plate_func_upper;
 
+  bool exact_roots; // If true we will allow approximate roots when no root is found
```
JunoRavin, line 92: This feels like the name of this variable should be approx_roots, not exact_roots, given the comment.

zero/gkyl_efit.h
```diff
@@ -26,28 +26,38 @@ struct gkyl_efit{
...
   struct gkyl_array* fpolflux;
   struct gkyl_array* qflux;
 
   int num_xpts; // Number of X-points
   double *Rxpt; // R coordinates of X points
   double *Zxpt; // Z coordinates of X-points
 
+  int num_xpts_cubic; // Number of X-points of cubic rep
+  double *Rxpt_cubic; // R coordinates of X points of cubic rep
+  double *Zxpt_cubic; // Z coordinates of X-points of cubic rep
```
JunoRavin, line 59: Is there a reason you have separate cubic variables here? Why can't we just reuse Rxpt and Zxpt if you are making them double*? Can't we just make them the right size based on whether or not you're using cubics?

zero/gkyl_mirror_geo.h
akashukla: The comment is wrong, I'll fix this

zero/gkyl_efit.h
akashukla: The reason than Rxpt and Zxpt are double* is because there could be any number of X-points (single null, double null, stellarator with a bunch of Xpoints).

The reason I made separate cubic variables is that I wanted to store both. The efit updater knows nothing about field line tracing : it does not know whether the field line tracer will use the cubic or quadratic representation. So, the efit object stores the x-points of the cubic and quadratic representation just as it stores the cubic and quadratic representations of psi.

The field line tracer takes a user input of whether or not to use the cubics, so the tracer will decide whether to use the cubic rep of psi along with cubic X-points or the quadratic.

This should definitely get cleaner later when we speed up the root finders and only use the cubics. Then we can get rid of all the quadratic stuff. For now storing both seemed like a reasonable design to me.

PR #452, opened 2024-08-22
gk-g0-app cross-species emission fixes

kltrbrds:
A number of minor modifications to the emission BC are necessary for it to work properly when the impacting species and emitted species are different and have different boundary grids.

- The emission spectrum was being projected onto the impacting grid instead of the emitting grid, this has been fixed with both grids being passed to the updater.
- The update range was using the emitted boundary range for the weighting of the yield instead of the impacting boundary range.
- A diagnostic was added to output the emitted distribution in the ghost cell in a separate file `bc_up` or `bc_lo`.
- The relaxation of the BC using `t_bound` wasn't quite working, a check was added to correctly cut off the scaling at the appropriate time.
- A preset for the ion-impact secondary electron emission of copper was added.

### DIFF
--- a/apps/gkyl_vlasov_priv.h
+++ b/apps/gkyl_vlasov_priv.h
@@ -244,6 +244,7 @@ struct vm_emitting_wall {
   double *scale_ptr;
   double t_bound;
   bool elastic;
+  bool write;
 
   struct gkyl_spectrum_model *spectrum_model[GKYL_MAX_SPECIES];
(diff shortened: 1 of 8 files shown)


tnbernard (approved):
These changes look good to me. The diagnostic is very useful, too, and I will implement something similar for recycling BCs on the GK side. I expect this to work smoothly with the recycling BCs and can test it after this PR is merged.

JunoRavin (approved):

PR #454, opened 2024-08-28
Caught a bug in the ion ionization collision term

tnbernard:
Electron prim moms were accidentally being used to project the Maxwellian for the ion ionization collision term, leading to unexpected ion heating.


JunoRavin (approved):

ammarhakim (approved):
PR is ready to merge

PR #457, opened 2024-09-11
Polarization density solving for initial conditions

Maxwell-Rosen:
These changes concern functionality that adjusts one species’s density (ions for mirrors) to match a provided potential. We want to use this feature in the mirror simulations because when we load initial conditions from CQL3D, they make assumptions different from ours, resulting in the Poisson equation not being solved the same way. In CQL3D, they solve for particle distribution function, but the GK modules use the guiding center distribution function. This adjustment is essential because the electron polarization density is negligible compared to the guiding center density, but polarization density is significant for ions. Hence, the ion distribution function must be scaled accordingly to ensure compatibility with the Poisson equation. Essentially, the electron distribution function is a very close approximation to the guiding center distribution function. We want to match the same potential that CQL3D outputs, so we scale the provided ion distribution function according to Poisson’s equation. 

The bulk of the logic here is contained in gk_species_apply_ic_cross. Here, the workflow is as follows.
1. The polarization density (npol) is computed using the gyrokinetic polarization density operator (gkyl_gyrokinetic_pol_density_advance). This accounts for the difference in handling polarization densities between the guiding center and particle distribution functions.
2. The guiding center density is corrected for contributions from other species. The code loops over all species and subtracts the contribution from each species' charge and guiding center density from the total polarization density.
3. The polarization density is then scaled by the inverse of the self-species charge to obtain the guiding center density for the specific species in focus.
4. The factor den_mod scales the distribution function (f) to match the corrected guiding center density. This ensures that the ion distribution function aligns with the potential obtained from CQL3D.

I have tested these changes. My unit test for mirrors works. All mirror regression tests work on CPU and GPU. My initialization protocols for production scale simulations produce good results.

I have added usage of this feature to the 2x and 3x wham regression tests.


Maxwell-Rosen:
I realize that this feature is not used in any regression or unit tests, so I will include some testing

JunoRavin (approved):

unit/ctest_polarization.c
```diff
@@ -0,0 +1,310 @@
+#include "gkyl_array.h"
```
manauref, line 1: @Maxwell-Rosen please adapt this unit test so it also runs on GPUs. See other unit tests, for example ctest_positivity_shift. With one variation, instead of using this
```
  distf_ho = distf;
  if (use_gpu)
    distf_ho = mkarr(false, distf->ncomp, distf->size);
...
  if (use_gpu)
    gkyl_array_release(distf_ho);
```
It is best to use this structure:
```
  distf_ho = use_gpu? mkarr(false, distf->ncomp, distf->size)
                    : gkyl_array_acquire(distf);
...
  gkyl_array_release(distf_ho);
```

apps/gk_species.c
```diff
@@ -407,6 +409,97 @@ gk_species_apply_ic(gkyl_gyrokinetic_app *app, struct gk_species *gks, double t0
...
+    }
+    
+    if (rank == 0) {
+      sprintf(fileNm, fmt, app->name, "den_mod");
+      gkyl_grid_sub_array_write(&app->grid, &app->global, 0,  den_mod_ho, fileNm);
+      sprintf(fileNm, fmt, app->name, "npol");
+      gkyl_grid_sub_array_write(&app->grid, &app->global, 0,  npol_ho, fileNm);
+      sprintf(fileNm, fmt, app->name, "m0");
+      gkyl_grid_sub_array_write(&app->grid, &app->global, 0,  m0_ho, fileNm);
+    }
```
manauref, line 484: Please don't add I/O here. This is only needed during development.

apps/gk_field.c
```diff
@@ -35,6 +35,26 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
...
+  f->init_phi_pol = false;
+  if (f->info.polarization_phi) {
+    // Project the initial potential onto a p+1 tensor basis and compute the polarization
+    // density to use use by species in calculating the initial ion density.
+    f->init_phi_pol = true;
+    struct gkyl_basis phi_pol_basis;
+    gkyl_cart_modal_tensor(&phi_pol_basis, app->cdim, app->poly_order+1);
+
+    f->phi_pol = mkarr(app->use_gpu, phi_pol_basis.num_basis, app->local_ext.volume);
+    struct gkyl_array *phi_pol_ho = mkarr(false, f->phi_pol->ncomp, f->phi_pol->size);
```
manauref, line 47: Change this to
```
struct gkyl_array *phi_pol_ho = app->use_gpu? mkarr(false, f->phi_pol->ncomp, f->phi_pol->size)
                                            : gkyl_array_acquire(f->phi_pol);
```
we need to be more mindful of and save memory

apps/gk_species.c
```diff
@@ -407,6 +409,97 @@ gk_species_apply_ic(gkyl_gyrokinetic_app *app, struct gk_species *gks, double t0
...
+      0, den_mod, 0, npol, 0, gks_self->m0.marr, &app->local);
+    gkyl_dg_mul_conf_phase_op_range(&app->confBasis, &app->basis, gks_self->f,
+      den_mod, gks_self->f, &app->local_ext, &gks_self->local_ext);
+
+          
+    const char *fmt = "%s-%s.gkyl";
+    int sz = gkyl_calc_strlen(fmt, app->name, "den_mod"); 
+    char fileNm[sz+1];
+    int rank;
+    gkyl_comm_get_rank(app->comm, &rank);
```
manauref, line 453: we wont' do I/O here, se comment below

apps/gk_species.c
```diff
@@ -407,6 +409,97 @@ gk_species_apply_ic(gkyl_gyrokinetic_app *app, struct gk_species *gks, double t0
...
+      npol_ho = mkarr(false, app->confBasis.num_basis, app->global_ext.volume);
+      m0_ho = mkarr(false, app->confBasis.num_basis, app->global_ext.volume);
+      gkyl_array_copy(den_mod_ho, den_mod_global);
+      gkyl_array_copy(npol_ho, npol_global);
+      gkyl_array_copy(m0_ho, m0_global);
+    } else {
+      den_mod_ho = den_mod;
+      npol_ho = npol;
+      m0_ho = gks_self->m0.marr;
+    }
```
manauref, line 475: I don't recognize all this comm and global stuff. If you are doing it for I/O, please remove it, see comment below.

regression/rt_gk_wham_2x2v_p1.c
```diff
@@ -110,11 +110,10 @@ struct gkyl_mirror_geo_efit_inp inp = {
 
 struct gkyl_mirror_geo_grid_inp ginp = {
   .rclose = 0.2,
-  .zmin = -2.48,
-  .zmax =  2.48,
+  .zmin = -2.0,
+  .zmax =  2.0,
```
manauref, line 114: why? has WHAM shrunk?

unit/ctest_mirror.c
```diff
@@ -0,0 +1,316 @@
+#include <math.h>
```
manauref, line 1: can you rename this unit test so it's more descriptive? what about the mirror is it testing?
also, add comments in the test describing what it's testing

manauref (changes requested):
See detailed comments. Two additional general things:
1. I think we should get rid of the nG_from_npol boolean in the input file. This is meant to tell the app which species we are going to compute nG from npol for, but I don't see us doing this for the electrons ever or any time soon. We'll only do it for the ions. So it's enough the pass the `polarization_phi` in the field table (maybe even call it polarization_potential for clarity), and then internally do this if charge>0 (how we do this for multiple ion species is another issue we won't deal with now).
2. Please run your unit tests through valgrind and ensure they are clean. If you can run a coarse regression test for a few time steps (at least 2), then run that through valgrind too.

regression/rt_gk_wham_2x2v_p1.c
Maxwell-Rosen: The original EFIT files did not mean to go this long. Wham has been \pm 2 meters all along

unit/ctest_mirror.c
Maxwell-Rosen: ctest_gk_geometry_mirror.c

Maxwell-Rosen:
The wham regression tests and the polarization unit test are all valgrind clean, validated on my Linux personal PC with a good valgrind installation

unit/ctest_polarization.c
```diff
@@ -0,0 +1,424 @@
...
+  struct gkyl_gyrokinetic_pol_density* npol_op = gkyl_gyrokinetic_pol_density_new(basis, grid, use_gpu);
+  struct gkyl_array *npol_ho = mkarr(false, basis.num_basis, localRange_ext.volume);
+  struct gkyl_array *epsilon_ho = mkarr(false, basis.num_basis, localRange_ext.volume);
+
+  struct gkyl_eval_on_nodes *epsilon_proj = gkyl_eval_on_nodes_new(&grid, &basis,
+    1, evalFunc1x_1, NULL);
+  gkyl_eval_on_nodes_advance(epsilon_proj, 0.0, &localRange, epsilon_ho);
+  gkyl_eval_on_nodes_release(epsilon_proj);
+
+  struct gkyl_array *npol_dev = use_gpu? mkarr(true, basis.num_basis, localRange_ext.volume)
```
manauref, line 76: please change your tests so these _dev's don't appear. Only the _ho arrays need a suffix.

In fact, looking at this logic, I think it's not quire right. You are calling it _dev but if use_gpu=false it is not actually a dev array.

unit/ctest_polarization.c
```diff
@@ -0,0 +1,424 @@
...
+
+TEST_LIST = {
+  { "test_1x_flat_cpu", test_1x_flat_cpu },
+  { "test_1x_quad_cpu", test_1x_quad_cpu },
+  { "test_2x_quad_cpu", test_2x_quad_cpu },
+  { "test_3x_flat_cpu", test_3x_flat_cpu },
+  { "test_1x_flat_gpu", test_1x_flat_gpu },
+  { "test_1x_quad_gpu", test_1x_quad_gpu },
+  { "test_2x_quad_gpu", test_2x_quad_gpu },
+  { "test_3x_flat_gpu", test_3x_flat_gpu },
```
manauref, line 422: these need to be placed in an if GKYL_HAVE_CUDA preprocessor statement

manauref (changes requested):

manauref (approved):

Issue #453, opened 2024-08-26
[DR] Upgrade positivity shift to maintain quasineutrality


manauref:
Presently we have an updater, `positivity_shift_gyrokinetic`, which is meant to enforce positivity of the distribution function `f` at certain "check" points (presently Gauss-Legendre quad points, but could be changed to Gauss-Lobatto nodes) . It works as follows:

1. Loop over phase space.
2. In each cell, perform a modal-2-nodal transform & check if $f_{\mathrm{in}}<0$ at check points.
3. If $f_{\mathrm{in}}$ is <0 at a check point, raise $f_{\mathrm{in}}$ to 0 at that point.
4. Perform a nodal-2-modal transform of the distribution $\Rightarrow f_{\mathrm{shifted}}$.
5. If a shift was performed, compute $\Delta  f=f_{\mathrm{shifted}} - f_{\mathrm{in}}$.
6. Compute and output the integrated moments of $\Delta  f$.

This is done independently to each species. As one would expect, this may not preserve quasineutrality in gyrokinetics as one may add electrons but not ions, or viceversa.

We propose to modify this updater so it preserves quasineutrality as follows:
1. Use a two-loop structure, one over conf- and one over vel-space, like we do in computing moments.
2. In the conf-loop, check if the current M0 moment (guiding center density $n_s$) if >0.
3. In the vel-loop, shift `f` if needed (as described above).
4. If M0 was >0 in this conf-cell, re-scale `f` so it has the same M0 as before the shifting. This procedure alone would guarantee (guiding-center) quasineutrality, because the number densities where not changed.
5. If M0 was <0 somewhere, we will have added some (guiding-center) density. After the above procedure is done to each species compute $\Delta  n_s=n_{s,\mathrm{shifted}} - n_{s,\mathrm{in}}$ and rescale each species species as
$$f_{s,\mathrm{new}} = \left[\left(n_{s,\mathrm{shifted}}+h_{sr}\right)/n_{s,\mathrm{shifted}}\right] f_{s,\mathrm{shifted}}$$
where
$$h_{sr}= \max(0, (\Delta n_r^{(0)} - \Delta n_s^{(0)})/|\Delta n_r^{(0)}-\Delta n_s^{(0)}|) * (\Delta n_r-\Delta n_s)$$
[...]

PR #461, opened 2024-09-14
Modify the positivity shift so it can use a Moe-Rossmanith-Seal limiter and preserve quasineutrality

manauref:
The design for this modification is described in issue #453 
Some plots of its test are also shown there. Furthermore, I ran rt_gk_sheath with `.enforce_positivity = true` through valgrind and it came out without any errors.

Note, presently it does regular shifting of f only, not the MRS limiter. To use the MRS limiter one has to change
https://github.com/ammarhakim/gkylzero/blob/f5029fc1502bc9432eb2742b4517999af7d190eb/zero/positivity_shift_gyrokinetic.c#L47
and I didn't make this a user-input because it is not yet clear which option is better, only with time, as people use this, will we be able to tell.

Another change to the app included here is that previously the positivity_shift updater was calculating part of the integrated moments of the shift. Now the updater only does the shifting; its hard for it to also compute diagnostics on the fly because of the potential quasineutrality-enforcing scaling that depends on other species and is computed after using the pos_shift updater. So now the diagnostics, both moments and integrated moments, are computed separately using the same infrastructure previously existing for regular moms and int_moms of f.


JunoRavin:
Can you clarify how frequently we call the positivity shift? I noticed it looked to me like the positivity shift is only being called after the RK3 step. In a flux-based positivity preserving scheme we will want the positivity preservation to happen every forward Euler step (I think the MRS limiter should be performed every forward Euler step for example). 

Am I mis-reading the program flow or is the positivity shift only meant to be a per total time step method instead of a per forward Euler method call?

JunoRavin (approved):
Chatted with @manauref about the current formulation and I understand now that the positivity shift is currently only being applied every RK3 step. When we get a flux-based limiter in place that insures the cell averages stay positive, we will likely want to experiment with where we do further positivity shifting, but that feels like future work to me at the moment.

tnbernard (approved):
Looks like good progress on the positivity issue. I approve this so that it can be applied in production runs for additional testing.

PR #462, opened 2024-09-17
Adapt Vlasov and reg test input file to use new parallelism input

manauref:
instead of .low_inp, .has_low_inp, and .use_gpu.

Tested by running rt_vlasov_weibel in all sorts of serial and parallel CPU/GPU combinations, always producing the same final field. Also did `make regression` to ensure all reg tests compile.

My only doubt is about is with regards to the tests where we separately pass `use_gpu` to `gkyl_wv_euler_new`. Are we handling this correctly? in this commit/PR we only modified the Vlasov app, not the wv_euler objects.

### DIFF
--- a/apps/gkyl_vlasov.h
+++ b/apps/gkyl_vlasov.h
@@ -268,8 +268,6 @@ struct gkyl_vm {
 
   double cfl_frac; // CFL fraction to use (default 1.0)
 
-  bool use_gpu; // Flag to indicate if solver should use GPUs
-
   int num_periodic_dir; // number of periodic directions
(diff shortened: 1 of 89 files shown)


JunoRavin:
So long as the wv_euler object is still taking the use_gpu flag like so:

struct gkyl_wv_eqn *euler = gkyl_wv_euler_new(ctx.gas_gamma, app_args.use_gpu);

everything is fine. This initialization is Vlasov following the design of moments in that certain systems instantiate their equation object outside of the app, and the app acquires a pointer to that equation object internally.

PR #463, opened 2024-09-17
Merge partial work in multib-app branch, primarily to use new cubic psi representation

manauref:
Shouldn't affect functionality of single block sims. I ran rt_gk_sheath_3x2v on CPU and GPU in this branch.


ammarhakim (approved):
Ready to merge. We can take a look in more detail later.

JunoRavin (approved):

Issue #459, opened 2024-09-13
[DR] Improve the way GPU and MPI options are specified in input files, and move decomposition into the app


manauref:
Presently our GK input files specify GPU and multi-CPU options with
```
.use_gpu = app_args.use_gpu,

.has_low_inp = true,
.low_inp = {
  .local_range = decomp->ranges[my_rank],
  .comm = comm
}
```
One core problem here is that the decomp object is only publicly available at the input file level, and we are then not able to do things with it inside the app. We could also improve the fact that this approach requires 3 different inputs by requiring a single table (like the rest of our input file where each feature is specified with a single table). So I propose that we have an interface such as
```
.parallelism = {
  .use_gpu = app_args.use_gpu,
  .use_mpi = app_args.use_mpi,
  .cuts = app_args.cuts,
  .comm = comm, // This is a gkyl_comm without a decomp, created .e.g. w/ gkyl_gyrokinetic_comms_new.
}
```

Internally we will use these arguments to create a decomposition, and duplicate the comm (but assign the decomp to it) with
```
    app->decomp = gkyl_gyrokinetic_comms_decomp_new(cdim, app->grid.cells,
      gk->parallelism.cuts, gk->parallelism.use_mpi, stderr);

    app->comm = gkyl_comm_split_comm(gk->parallelism.comm, 0, app->decomp);
```
(note that we are making `decomp` and app member).

The above simplifies input files, reduces code, and makes room for future functionality which require access to the `decomp` object.

This interface is prototypes in gk-g0-app_parallel_inp.

manauref:
Resolved in PR #469

PR #469, opened 2024-09-18
New interface in input file to specify parallelism options

manauref:
See issue #459 

We changed every regression test input file for moments, pkpm, vlasov and GK.  Now the parallelization options are specified with (e.g. for a 3x sim)
```
.parallelism = {
  .use_gpu = app_args.use_gpu,
  .cuts = {app_args.cuts[0] , app_args.cuts[1], app_args.cuts[2],},
  .comm = comm, // This is a gkyl_comm without a decomp.
}
```
while before we used
```
.use_gpu = app_args.use_gpu,
.has_low_inp = true,
.low_inp = {
  .local_range = decomp->ranges[my_rank],
  .comm = comm
}`
```

We ran some of these tests on single and multiple CPUs or GPUs successfully. Also, `make regression` and `make unit` complete.

This cleans up the specification of whether to use MPI and/or GPUs. It also makes the decomp object internal to (and saved in) the app.

This also makes the necessary change in the multib interface.


JunoRavin (approved):

ammarhakim (approved):
Ready for review

PR #470, opened 2024-09-19
Fix for reaction rates when using non-uniform velocity grids

jRoeltgen:
The reaction rate distribution function was not being scaled by the velocity space jacobian which is needed with non-uniform velocity grids. 
**Note:** (EDIT: Resolved: See comment below) Adding this fix does not result in a change in the resulting electron moments even though this would be expected with a quadratic velocity mapping, so there is likely a remaining bug I'm looking for.


jRoeltgen:
The lack of change in the electron density when adding the scaling by the velocity jacobian is due to the reactions having a minor effect in the change in the electron density over the simulation time in `rt_gk_li_react_3x2v_p1`. Also the density changes by a very small amount ~10<sup>-4</sup>n0.  In another test where the atomic reactions are dominant, a significant change in the electron density was observed.

manauref:
thanks @jRoeltgen 
can you describe that other test or point to its input file? that might be a better regression test for reactions than this gk_li_react then, because this test clearly didn't help us identify a bug.

jRoeltgen:
Added regression/rt_ar_react_1x2v_quadraticV.c as the test described in the prior comment.

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+  double T_src; // Temp density.
+  double vpar_max_elc; // Velocity space extents in vparallel for electrons
+  double mu_max_elc; // Velocity space extents in mu for electrons
+  double vpar_max_ion; // Velocity space extents in vparallel for ions
+  double mu_max_ion; // Velocity space extents in mu for ions
+  double vpar_max_Ar; // Velocity space extents in vparallel for Li ions
+  double mu_max_Ar; // Velocity space extents in mu for Li ions  
+  double finalTime; // end time
+  int numFrames; // number of output frames
+};
```
manauref, line 66: @jRoeltgen please make this test conform to the structure and logic in other tests (e.g. see rt_gk_sheath_3x2v). In particular, do what most other tests do to specify cdim, vdim, number of cells, temporal variables (e.g. num_frames, t_end, etc).

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+      strcpy(reactions[j].ionization[i].ion_nm, ion_nm);
+      strcpy(reactions[j].ionization[i].donor_nm, donor_nm);
+      reactions[j].ionization[i].type_self = j;
+    }
+  }
+  return reactions;
+}
+
+void
+write_data(struct gkyl_tm_trigger *iot, gkyl_gyrokinetic_app *app, double tcurr)
```
manauref, line 582: @jRoeltgen please stick to the structure and workflow of the 2 I/O functions in other regression tests

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+    gkyl_gyrokinetic_app_write(app, tcurr, iot->curr-1);
+    gkyl_gyrokinetic_app_calc_mom(app); gkyl_gyrokinetic_app_write_mom(app, tcurr, iot->curr-1);
+  }
+}
+
+int  
+main(int argc, char **argv)
+{
+  struct gkyl_app_args app_args = parse_app_args(argc, argv);
+
```
manauref, line 594: @jRoeltgen please add parallelism to this tests. All our tests should be able to run in parallel

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+  }
+
+  struct gk_app_ctx ctx = create_ctx(); // context for init functions
+
+  // Reactions for argon isonuclear sequence
+  struct all_of_1species_reactions *reactions = create_isonuclear_react(ctx.massAr, ctx.massElc, 18, GKYL_ION_AR, "elc", "Ar");
+
+  int NX = APP_ARGS_CHOOSE(app_args.xcells[0], 2);
+  int NV = APP_ARGS_CHOOSE(app_args.vcells[0], 6);
+  int NMU = APP_ARGS_CHOOSE(app_args.vcells[1], 4);
```
manauref, line 607: @jRoeltgen please see how number of cells are specified and passed to the app in established tests (e.g. rt_gk_sheath_3x2v_p1) and stick to that

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+     .collisions =  {
+      .collision_id = GKYL_LBO_COLLISIONS,
+      .normNu = true,
+      .n_ref = ctx.n0,
+      .T_ref = ctx.Te,      
+      .self_nu = evalNuElc,
+      .ctx = &ctx,
+      .num_cross_collisions = 3,
+      .collide_with = { "ion", "Ar1", "Ar2" },
+      },
```
manauref, line 642: fix spacing

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+          .ion_nm = "Ar2",
+          .recvr_nm = "Ar1",
+          .charge_state = 1,
+          .ion_mass = ctx.massAr,
+          .elc_mass = ctx.massElc,
+        },
+      },
+      },
+    
+    /*.bcx = {
```
manauref, line 670: delete commented code, or describe in comments with your initials and date why this commented code should really be kept. Same for other commented code below and elsewhere

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+      .normNu = true,
+      .n_ref = ctx.n0Ar2,
+      .T_ref = ctx.TAr,
+      .ctx = &ctx,
+      .self_nu = evalNuIon,
+      .num_cross_collisions = 3,
+      .collide_with = { "elc", "ion", "Ar1" },
+    },
+
+        .react = {
```
manauref, line 822: fix spacing

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+    .kperpSq = pow(ctx.kperp, 2.),
+  };
+
+  // GK app
+  struct gkyl_gk gk = {
+    .name = "ar_react_1x2v_quad_wJacobvel",
+
+    //.cfl_frac = 0.1,
+
+    .cdim = 1, .vdim = 2,
```
manauref, line 875: @jRoeltgen see how cdim and vdim are specified in established codes, or specifically in this branch
https://github.com/ammarhakim/gkylzero/pull/469

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+
+    .num_periodic_dir = 1,
+    .periodic_dirs = {0},
+
+    .num_species = 4,
+    .species = { elc, ion, Ar1, Ar2},
+
+    .field = field,
+
+
```
manauref, line 899: delete

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+      gkyl_gyrokinetic_app_cout(app, stdout, " dt = %g\n", status.dt_actual);
+    }
+    if (!status.success) {
+      gkyl_gyrokinetic_app_cout(app, stdout, "** Update method failed! Aborting simulation ....\n");
+      break;
+    }
+    tcurr += status.dt_actual;
+    dt = status.dt_suggested;
+
+    write_data(&io_trig, app, tcurr);
```
manauref, line 936: follow the same logic/structure in time setting, time stepping, writing data, etc, as established reg tests.

regression/rt_ar_react_1x2v_quadraticV.c
```diff
@@ -0,0 +1,969 @@
...
+
+    step += 1;
+  }
+  gkyl_gyrokinetic_app_calc_field_energy(app, tcurr);
+  gkyl_gyrokinetic_app_write_field_energy(app);
+  gkyl_gyrokinetic_app_stat_write(app);
+
```
manauref, line 943: same here

manauref (changes requested):
thanks for catching this bug. Please change your regression tests so they follow the same structure and workflow as established regression tests, preferably the ones here
https://github.com/ammarhakim/gkylzero/pull/469
then we can merge after that

regression/rt_ar_react_1x2v_quadraticV.c
jRoeltgen: I updated the regression test. To follow https://github.com/ammarhakim/gkylzero/pull/469 required a merge which seems to have introduced a merge conflict which @manauref should probably resolve because of its location.

Issue #456, opened 2024-09-10
[DR] Bringing Vlasov Poisson into the Vlasov app


manauref:
Presently we have a separate Vlasov Poisson app in the gk-g0-app_vp branch (and some sub branches). We'd like to bring this into the main Vlasov app for (at least) 2 reasons:
1. So it's easier to use components common to both solvers (e.g. collisions).
2. So that VP undergoes the same level of scrutiny as other solvers, giving users more peace of mind.

The approach to do so is as follows:
1. A user will indicate in the app input table (`gkyl_vm`) that they wish to run an electrostatic simulation with `is_electrostatic`, such as
```
struct gkyl_vm vm = {
    .name = "vp_sheath_1x1v_p2",

    .cdim = ctx.cdim, .vdim = ctx.vdim,
    .lower = { ctx.x_min },
    .upper = { ctx.x_max },
    .cells = { ctx.Nx },
    .poly_order = ctx.poly_order,
    .basis_type = app_args.basis_type,

    .is_electrostatic = true,
...
```
2. Additionally, input files will specify alternative entries in the field input table (`gkyl_vlasov_field`) like
```
  struct gkyl_vlasov_field field = {
    .epsilon0 = ctx.epsilon0,
    .poisson_bcs = {
      .lo_type = { GKYL_POISSON_DIRICHLET },
      .up_type = { GKYL_POISSON_NEUMANN },
      .lo_value = { 0.0 }, .up_value = { 0.0 }
    },
    // External potentials are optional:
    .ext_potentials = ext_potentials_profiles,
    .ext_potentials_ctx = &ctx,,
    .ext_potentials_evolve = false,
  };
```
3. Internally, the only new object we will create is a `apps/vp_field.c`, which handles the Poisson problem and the external potentials. The Vlasov app will choose between `vm_field` or `vp_field` based on the `is_electrostatic` input (default=false).
4. In order to not further balloon the Vlasov Maxwell solver, to avoid its accidental modification, and make the code more modular, we will create a separate rk3 stepping function for VP. This `update` function will be chosen based on `is_electrostatic` in a similar way to how we choose the operator split update (e.g. for implicit BGK), e.g.
```
  if (app->is_electrostatic) {
    if (app->has_implicit_coll_scheme || app->has_fluid_em_coupling) {
      app->update_func = vlasov_poisson_update_op_split;
    }
    else {
      app->update_func = vlasov_poisson_update_ssp_rk3;
    }
  }
  else {
[...]

PR #473, opened 2024-09-21
Implement a Vlasov-Poisson solver in the Vlasov app

manauref:
A Vlasov Poisson solver is desirable for some electrostatic simulations. In recent times we prototyped implementing this as a separate app in the gk-g0-app_vp branch. In that branch our colleagues used in and reproduced results from Skolar et al. Phys. Plasmas 30, 012504 (2023). One of the complications of that implementation is that it had separate code for things that are identical for Vlasov-Maxwell, like sources, collisions and BCs. This suffered from all the drawbacks of code duplication.

In this work we bring that VP solver into the existing Vlasov app as described in the DR in issue #456. Namely:
1. Users indicate they want VP by saying `.is_electrostatic = true,` in the app table of the input file.
2. The field table of the input file requires Poisson BCs (e.g. like in GK).
3. Based on these inputs, the vlasov app initializes a Vlasov species (with a Vlasov Poisson equation instead of Vlasov Maxwell, which also accepts external potentials phi_ext and A_ext), chooses a VP field (`vp_field.c`) instead of a VM field (`vm_field.c`) (although they both return a vm_field type), and chooses the pointer to the update method so it does the vlasov_poisson_ssp_rk3 update.
4. All other modules (e.g. collisions, sources, BCs, etc) remain the same for both solvers.

There are several new VP regression tests now, testing collisionless physics, collisions, sheaths and emissions. All of them were run on CPUs and GPUs. I ran rt_vlasov_poisson_sheath_Bext_1x2v_p2 on 2 GPUs. Some of these were compared against results from g2 in addition to the aforementioned comparison to Skolar's paper. Furthermore, I ran the following Vm regression tests:
rt_vlasov_advect_1x.c
rt_vlasov_emission_spectrum_1x1v_p2.c
rt_vlasov_landau_damping_1x1v_p2.c
rt_vlasov_neut_bgk_sodshock_1x1v_p2.c
rt_vlasov_sheath_1x1v_p2.c
rt_vlasov_sr_twostream_1x1v.c
rt_vlasov_twostream_p2.c
rt_vlasov_weibel_2x2v_p2.c
on both CPU and GPU, and I ran the weibel tests on 2 GPUs. In all cases the final field was unchanged.

Some additional side remarks:
- FIXED in PR #482 NOW: Most importantly, we need to fix the Makefile issue with long link lines. I think it's a problem here again (I circumvented it by printing the link line to an .sh file and running that)
- There was a bug in emission BCs when running on GPUs which we fixed here.
- Unlike GK, Vlasov applies BCs after forward_euler in the rk3 step, but not after the final state of the stepping is computed. In GK we determined that the only way to do BCs that depends on the fields AND to be able to completely reproduce a state upon restart without ghost cells was to apply BCs after the full rk3 step is completed. I thought the latter is the more correct approach, though I'm not sure to be honest. It's possible that for simple BCs like absorbing it doesn't matter, but for nonlinear (in f) BCs like sheath BCs or emission BCs, it probably does. We can experiment with this outside of this PR in the future.

Some points were feedback is desired:
a) should we have `.is_electrostatic` in the app table as it is now, or should we move it to the field table?
b) as you read the code, look for places where we have if-statements checking whether a field object exists or whether the field is a VP field (GKYL_FIELD_PHI or GKYL_FIELD_PHI_EXT). It'd be good to write things in a way that simplifies this/does away with some of those if-statements.

This PR goes with PR 43 in gkylcas: https://github.com/ammarhakim/gkylcas/pull/43


apps/vlasov.c
```diff
@@ -290,12 +298,11 @@ gkyl_vlasov_app_new(struct gkyl_vm *vm)
...
   // allocated in the previous step
   for (int i=0; i<ns; ++i)
-    if (app->species[i].collision_id == GKYL_LBO_COLLISIONS
-      && app->species[i].lbo.num_cross_collisions)
+    if (app->species[i].collision_id == GKYL_LBO_COLLISIONS &&
+        app->species[i].lbo.num_cross_collisions)
       vm_species_lbo_cross_init(app, &app->species[i], &app->species[i].lbo);
 
-  // initialize each species source terms: this has to be done here
-  // as they may initialize a bflux updater for their source species
```
JunoRavin, line 298: I would like to leave this longer comment in so no future developers try to move the source terms into the species initialization unless @kltrbrds says that this comment can be safely deleted (but I think source terms can still fetch a "source species" if they are using the boundary flux from another species).

apps/vlasov.c
```diff
@@ -384,31 +395,80 @@ vm_find_fluid_species_idx(const gkyl_vlasov_app *app, const char *nm)
...
   struct timespec wtm = gkyl_wall_clock();
-  vm_field_apply_ic(app, app->field, t0);
-  app->stat.init_field_tm += gkyl_time_diff_now_sec(wtm);
 
-  vm_field_apply_bc(app, app->field, app->field->em);
+  if (app->field->field_id == GKYL_FIELD_PHI || app->field->field_id == GKYL_FIELD_PHI_EXT) {
+    struct gkyl_array *distf[app->num_species];
+    for (int i=0; i<app->num_species; ++i)
+      distf[i] = app->species[i].f;
+    vp_calc_field(app, t0, (const struct gkyl_array **) distf);
```
JunoRavin, line 466: What is the cast to a const struct gkyl_array ** for? Shouldn't the function signature be for a gkyl_array* f[] so you can just pass distf (like what we do for passing distribution functions to the forward Euler methods).

apps/vlasov.c
manauref: done

apps/vlasov.c
manauref: if you don't cast it doesn't work. You can try it. You'll get a cascade of error messages and warnings, and as you try to resolve them you'll find that all the arrays of gkyl_arrays (e.g. const struct gkyl_array *fin[]) need to be defined without a const throughout the whole system (e.g. in vlasov_forward_euler), in both VM and VP. This casting is a minimal modification approach, and is also what we used in GK.

JunoRavin (approved):

ammarhakim (approved):
Ready to merge.

PR #474, opened 2024-09-21
Fix a bug in the positivity shift diagnostics and in proj_bimaxwellian_on_basis, and implement FourMoments

manauref:
New feature:
- A FourMoments option for GK which calculates M0, M1, M2par and M2perp together.

Two fixes:
1. Previously we were making the positivity shift diagnostic moments the same as those the user requested. But when the user requests Maxwellian or BiMaxwellian moments one gets a division error at t=0 when computing upar of the positivity shift, because its density is zero. We fix this by simply making the diagnostic for pos shift always be FourMoments.
2. Fix a missing factor of 0.5 in the proj_bimaxwellian_on_basis_lab_mom. It had snuck in when we changed the definition of M2perp.

This goes with PR 44 in gkylcas: https://github.com/ammarhakim/gkylcas/pull/44


JunoRavin (approved):

ammarhakim (approved):
Ready to merge

Issue #445, opened 2024-08-13
[DR] Initializing a 5D distribution function from a 4D distribution in a file


Maxwell-Rosen:
MF 2024/09/12: I have completed a (serial) prototype of this. See my comment below.

For the mirror simulations, we aim to initiate a 3x2v gyrokinetic simulation using initial conditions derived from a 2x2v simulation. In the 2x2v setup, there is no dependence on the cylindrical angle. However, we aim to introduce this angular dependence in the 3x2v simulation by replicating the 2x2v data in each angular slice. Additionally, we would like to apply a user-specified function to this angular component to introduce noise into the initial conditions to trigger instabilities. These simulations should maintain consistent resolution and domain in velocity and physical space coordinates and use the same basis functions as the original data. Similarly, it would be valuable to load a 2x2v simulation from a 1x2v simulation of a single field line and then introduce radial dependence. Developing this infrastructure to initialize higher-dimensional simulations from lower-dimensional ones is crucial for advancing our mirror simulation efforts.

When loading this data, the structure in the code could look like this.
```
struct gkyl_gyrokinetic_species ion = {
   .projection = {
      .proj_id = GKYL_PROJ_FROM_LOWER_DIM,
      .filename = "ion_2x2v.gkyl"
      .func_dependence = user_func,
      .func_ctx = &ctx,
   },
};
```
The user may need to specify the original dimensions of the simulation, but I think that can be read from the file. I'm not sure what the projection routine would look like under the hood.

Maxwell-Rosen:
I have gotten some initialization stuff working inside apps/gk_species_projection.c. I can load a lower-dimensional distribution function, read the grid, and construct lower-dimensional bases and ranges. I did not consider parallelism when creating these ranges and grids. The construction of these objects is identical to how gyrokinetic.c creates an app. It would re-use less code if we called gkyl_gyrokinetic_app_new() and created a new app with the appropriate dimensions instead. It would undoubtedly be less code, but this call has a lot of overhead, and it isn't a good idea.

[...]

PR #479, opened 2024-09-23
Import (possibly lower dimensional) initial distributions from file in GK

manauref:
This implements a feature that allows us to read in an initial distribution function from a file in GK. At present it supports reading:
- A distribution with the same dimensionality and grid.
- A distribution with one fewer dimension, to load 2x distributions as ICs for a 3x sim.
The interface also allows specifying an optional conf-space multiplicative field, or a phase-space additive field.
The design for this work is outlined in issue #445 (see my later comment here https://github.com/ammarhakim/gkylzero/issues/445#issuecomment-2346885508). And this goes with PR 46 in gkylcas: https://github.com/ammarhakim/gkylcas/pull/46

In order to support this we had to:
1. make some functions that handle metadata creation/releasing app-wide (still private), so that they can be used in gk_species.c to read the metadata of the .gkyl file with the ICs, even though these methods are declared in gyrokinetic.c
2. create another communicator with a possibly lower dimensionality. This required extracting `cuts` from the decomp object in the app, for which we implemented a method in rect_decomp which gets the cuts (this method is tested in the rect_decomp unit tests).
3. Implement an updater which translates the DG coefficients of a lower dim field to a higher dim field (and its unit tests).

We tested this feature with both:
- Loading rt_gk_sheath_2x2v_p1 final distributions as ICs for rt_gk_sheath_3x2v_p1.
- Loading 2x WHAM distributions into a 3x sim.

In the not too distant future we may also use this framework to load distributions with the same dimensionality but different resolution (e.g. to initialize a higher resolution sim from a coarser resolution one).

Feedback from reviewers is desired regarding the input file interface, e.g.
```
ion = {
  ...
  .init_from_file = {
    .type = GKYL_IC_IMPORT_F,
    .file_name = "gk_sheath_2x2v_p1-ion_1.gkyl",
    // Optional: specify a conf-space scaling factor and/or a phase-space
    // additive perturbation, so that f=alpha(x)*f_{file}+beta(xv) (must
    // all so change the 'type' entry).
//  .conf_scale = ic_conf_fac, // Function pointer.
//  .conf_scale_ctx = &ctx,
//  .phase_add = {
//    .proj_id = GKYL_PROJ_FUNC,
//    .func = ic_phase_add, // Function pointer.
//    .ctx_func = &ctx,
//  },
  },
  ...
}
```
Specifically, do we want this `.type` enum? if so, are we ok with these option names (i.e. GKYL_IC_IMPORT_F, GKYL_IC_IMPORT_AF, GKYL_IC_IMPORT_AF_B) or do we want others?


JunoRavin:
I think a lot of this work is fantastic. I just want to clarify something in the design because I know a few places use this design now and it is distinct from how I've been writing GPU functionality in these sorts of self-contained updaters. 

The way you handle the GPU functionality is to allocate a kernels struct: https://github.com/ammarhakim/gkylzero/blob/a8193c1d0562a240cea36fbba70fffda4d615f06/zero/translate_dim_gyrokinetic.c#L28

In contrast to making a GPU updater like, for example, what I do with the EM variable calculator (and following the on_dev structure we utilize in dg_eqn, mom_type, etc.): https://github.com/ammarhakim/gkylzero/blob/main/zero/dg_calc_em_vars.c#L20

I understand the logic. In this case, this struct for the transdim transformation is a self-contained object that only needs a specific function pointer which is accessible on device, and one can even pass that function pointer directly to the device kernel like you do here: https://github.com/ammarhakim/gkylzero/blob/a8193c1d0562a240cea36fbba70fffda4d615f06/zero/translate_dim_gyrokinetic_cu.cu#L72

But do we want to generically follow this sort of design for these sorts of self-contained updaters? Part of my logic in making even the self-contained updaters utilize the on_dev and alloc_flags infrastructure we use for dg_eqn, mom_type, etc. was to make it completely transparent to new developers what the GPU extension of a given object should look like, but we now have two different ways of adding GPU functionality, one with structs storing an on_dev pointer and one with structs storing a kernel pointer that could either be on host or on device. I could be persuaded that we should have two distinct mechanisms for putting things on GPUs, but I just want to clarify this design.

manauref:
I'll add that the logic around the kernels struct in this PR is slightly different, better and more compact, than that in older updaters that use the kernel struct, and we should use this structure (if we create objects with the kernels struct) going forward.

JunoRavin (approved):

manauref:
@JunoRavin I had another thought on the topic of creating a device version of the updater as I was creating a new updater but with the structure you pointed to above (e.g. in dg_calc_em_vars). One thing I don't like about that approach is that you essentially end up coding two _new methods, introducing code duplication.

I may use it in a new updater I'm creating elsewhere (which needs more than just a `kernels` struct on the device), but I think another approach is to do something like what GK geometry does, where you create the updater on the host and at the end you copy the whole thing to the device. The complexity of this copy varies by updater though, so in some cases this may not be the easiest way either.

ammarhakim (approved):
Ready to merge

Issue #398, opened 2024-06-20
Error due to long link line: Refactor makefile so we can link more object files


manauref:
In some branches of the code (e.g. gk-g0-app_vp_cudss) compiling the code is resulting in this error after `make install`:

```
...
.cu.o cuda-build/zero/bc_twistshift_cu.cu.o cuda-build/zero/fem_poisson_perp_cu.cu.o cuda-build/zero/bc_basic_cu.cu.o cuda-build/zero/cart_modal_serendip_cu.cu.o cuda-build/zero/dg_diffusion_gyrokinetic_cu.cu.o cuda-build/zero/cart_modal_gkhybrid_cu.cu.o cuda-build/zero/dg_calc_vlasov_gen_geo_vars_cu.cu.o -L/scratch/gpfs/manaurer/gkeyll/code/g0-vp/gkylsoft/superlu/lib -L/scratch/gpfs/manaurer/gkeyll/code/g0-vp/gkylsoft/OpenBLAS/lib -Lcuda-build -L/usr/local/openmpi/cuda-11.1/4.1.1/gcc/lib64 -L/usr/lib64 -L/scratch/gpfs/manaurer/gkeyll/code/g0-vp/gkylsoft/cuDSS/lib/ -L. -lsuperlu /scratch/gpfs/manaurer/gkeyll/code/g0-vp/gkylsoft/OpenBLAS/lib/libopenblas.a  -lcublas -lcusparse -lcusolver -Xlinker "-rpath,/usr/local/openmpi/cuda-11.1/4.1.1/gcc/lib64" -lmpi -lnccl -Xlinker "-rpath,/scratch/gpfs/manaurer/gkeyll/code/g0-vp/gkylsoft/cuDSS/lib/" -lcudss   -lm -lpthread -ldl -o cuda-build/libgkylzero-install.so
make: execvp: /bin/sh: Argument list too long
make: *** [Makefile:364: cuda-build/libgkylzero.so] Error 127
make: *** Waiting for unfinished jobs....
make: *** [Makefile:374: cuda-build/libgkylzero-install.so] Error 127
```

This is on stellar-amd, building with nvcc 12.4.

One could actually reroute the output of `make install` to a file, and then manually copy that link command (~130k characters) into the terminal without a problem. So the issue is not the shell, but that the `make` command cannot execute such long line. Potential solutions include:
1. Refactor the makefile so we have several smaller linking steps.
[...]

PR #482, opened 2024-09-26
Change makefile to avoid long link line error

manauref:
As the makefile is currently written we can encounter an error about long link lines, see issue #398 

In the past we avoided this problem by reducing the number of characters in the folder names build/ and cuda-build/ by renaming them bld/ and cubld/.

Here we offer another solution, where we dump the object files in the the long link line into a temporary file, then invoke that file in the link command.

I tested this on my Mac (make 3.81), on stellar-amd and on perlmutter-gpu (make 4.2.1). They all produced working executables, for regression tests and input files outside of gkylzero/. I also ran them with MPI and with NCCL.


ammarhakim (approved):
Ready for merge

JunoRavin (approved):

PR #486, opened 2024-09-29
Gk g0 app transdim maxwell

Maxwell-Rosen:
Adding a regression test for the translate dimension functionality


regression/rt_gk_wham_translate_dim_gyrokinetic_2to3x.c
```diff
@@ -0,0 +1,1165 @@
+#include <math.h>
```
manauref, line 1: @Maxwell-Rosen please rename this test rt_gk_wham_2xIC_3x2v_p1

translate_dim is just the name of the updater and only makes sense to developers, but the capability involves a bit more than that, and you are testing the whole thing here.

manauref (changes requested):
@Maxwell-Rosen the comments don't mention it, but is the test valgrind clean?

PR #487, opened 2024-10-01
Fix an error in the eval_cubic_wgrad2 function

akashukla:
This is a bug fix that affects mirror geometry on the magnetic axis. Before commit 4ce1631, there was a memory error that obscured this issue. When @manauref was trying to run mirror calculations post this commit, he noticed the error. We have tracked it back to this mistake which fixes the issue.

Error in the eval_cubic_wgrad2 function in dg_basis_ops: Accidentally returned only one of the second derivatives and the mixed partial rather than both second derivatives and the mixed partial. (This is for the 2x case). This was breaking on axis calculation of bmag for the mirror.

This wgrad2 function is supposed to return f, d^2f/dx^2, d^2f/dy^2, and d^2f/dxdy which it now does.

### DIFF
--- a/zero/dg_basis_ops.c
+++ b/zero/dg_basis_ops.c
@@ -622,6 +622,7 @@ eval_cubic_wgrad2(double t, const double *xn, double *fout, void *ctx)
 
   fout[0] = ectx->basis.eval_expand(eta, fdg);
   if (ectx->ndim > 1) {
+    fout[1] = eval_laplacian_expand_2d_tensor_p3(0, eta, fdg);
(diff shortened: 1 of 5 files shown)


manauref (changes requested):
@akashukla my impression is that line 230 of efit.c is still missing a minus sign. Can you check this?

Also, looking at the public header file for dg_basis_ops we find things like
```
  evalf_t eval_cubic_wgrad; // function pointer to evaluate the cubic & its gradient
```
which although say what they do, they don't show the signature, how to use it, or how to interpret the outputs. Could this be improved please?

manauref (approved):

JunoRavin (approved):

PR #492, opened 2024-10-10
All tetrad-related work, GR Maxwell equations, and coupled fluid-Einstein equations

JonathanGorard:
This branch contains several core additions to Gkeyll's burgeoning GR framework, with the principal ones being:

- New functionality for handling Maxwell's equations in curved spacetime (in strongly hyperbolic form) with `wv_gr_maxwell`. Initial prototyping work has been done to try to introduce a fully covariant version of the hyperbolic divergence cleaning approach to the curved spacetime Maxwell equations, although this remains incomplete, and has been disabled at present;
- New functionality for evolving general relativistic Euler equations with an ultra-relativistic equation of state with `wv_gr_ultra_rel_euler` (in contrast to the existing functionality for handling ideal gas equations of state, since in the ultra-relativistic case the fluid density and pressure are no longer algebraically independent, and so the baryonic number density equation does not need to be evolved, and the conservative-to-primitive variable reconstruction algorithm may be dramatically simplified using the approach of [Neilsen and Choptuik](https://arxiv.org/abs/gr-qc/9904052));
- Tetrad-basis representations of all general relativistic equation objects, alongside their standard curved spacetime variants (thus allowing for direct comparison between the two), namely `wv_gr_euler_tetrad`, `wv_gr_ultra_rel_euler_tetrad`, and `wv_gr_maxwell_tetrad`. These equation systems work by calculating a flat spacetime form of the flux and then applying a post-hoc geometric correction, rather than calculating the curved spacetime fluxes directly;
- Initial prototyping work on coupled fluid-Einstein equations in plane-symmetric Gowdy spacetimes using the so-called "Medium" formalism of [Barnes, Lefloch, Schmidt and Stewart](https://iopscience.iop.org/article/10.1088/0264-9381/21/22/003), with `wv_gr_medium`. This has primary been done as a proof-of-concept demonstration that the kinds of finite-volume numerical algorithms currently implemented within the g0 moment app can indeed be used to evolve the Einstein equations in addition to hydrodynamic and electromagnetic ones, and that we need not restrict ourselves only to stationary spacetime metrics going forward. There is much more to be done here, but initial validation of the approach has been successful.

All of the new equation systems described above have been validated using both new unit tests (for instance, allowing one to compare the fluxes obtained through the tetrad-first flat spacetime flux correction approach vs. the fluxes obtained using the direct curved spacetime flux calculation approach) and new regression tests. There has also been substantial cleanup of various aspects of the GR code, including simplification of our Roe solver code (since, thanks to the tetrad-first approach, we need only worry about having an SR Riemann solver working in most cases) and streamlining of our handling of source terms thanks to the new moment source refactor.

In the medium-term, although Roe and Lax fluxes have been useful for initial prototyping and validation of these algorithms, I would ideally like for us to start transitioning to using some kind of intermediate approximate Riemann solver (Lax is typically too diffusive, Roe is typically too fragile and oscillatory), such as HLL, for our general relativistic equation systems. However, I wanted to open a PR for this functionality sooner rather than later, lest the branch become too far out-of-sync with `main`.


ammarhakim (approved):
Ready for merge once Jimmy approves.

JunoRavin:
My only concern is that there are still a number of places in the GR code where we are doing internal mallocs, and I feel like we should avoid this at all costs. Is there a reason we do not allocate these arrays on the stack? https://github.com/ammarhakim/gkylzero/blob/2c9c1047337cf61060900936180c90bd5956f843/zero/wv_gr_euler.c#L23

I understand there are a couple of places, like with the inverse spatial metric, where right now it might still make sense to do a malloc and we can fix this later, but in these functions I do not understand why we are doing mallocs.

JonathanGorard:
@JunoRavin Thanks! Most of the internal mallocs used in things like flux computations, source computations, and primitive variable reconstruction, have now been replaced with stack-allocated arrays. You're correct that there are still a few places, such as in the computation of inverse spatial metric tensors, where it still makes sense to use dynamically allocated memory - eventually I would like to wrap all of these in some kind of `gkyl_tensor` object that handles the memory management automatically (allocating on the stack wherever possible). We should revisit this soon, but for now I think this will do.

JunoRavin (approved):

PR #500, opened 2024-10-22
A unit test with a struct-based approach to arrays of arrays

manauref:
Just a unit test showing how to use structs to wrap arrays to create arrays of gkyl_arrays, or arrays of arrays of gkyl_arrays. 
It's kind of amazing that one can do this, especially since I think (not tested, just though of this now) you could simply use a single struct with recursion and create arrays of arrays of arrays of..

FYI @jRoeltgen


ammarhakim:
I do not see the point of this. This is a nice curiosity but I do not think it is useful to add it to the main. If you want, you can move the struct defn to the unit test and delete the header you added. No such headers in the unit test directory should be added.

manauref:
@ammarhakim if I move it to struct to the test, how do I make it visible to the CUDA kernel? do I define it again in the .cu file?

PR #501, opened 2024-10-22
A unit test with a struct-based approach to arrays of arrays v2

manauref:
Just a unit test showing how to use structs to wrap arrays to create arrays of gkyl_arrays, or arrays of arrays of gkyl_arrays.
It's kind of amazing that one can do this, especially since I think (not tested, just though of this now) you could simply use a single struct with recursion and create arrays of arrays of arrays of..

We are trying to use similar workflows in radiation, so I wanted to demo/test it here.

FYI @jRoeltgen

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,7 +187,7 @@ UNIT_CU_OBJS =
 # There is some problem with the Vlasov and Maxwell kernels that is causing some unit builds to fail
 ifdef USING_NVCC
 #	UNIT_CU_SRCS = $(shell find unit -name *.cu)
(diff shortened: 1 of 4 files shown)


ammarhakim (approved):
I am approving this but I think we need to think more carefully on how to pass multiple arrays to the GPU. For now this is ok but I do not want hacks to propagate into the code. Pretty soon we will forget why we did this.

JunoRavin (approved):

manauref:
i'll mention that @JonathanGorard made the point that maybe we shouldn't use unit/ to place demos. I hear that. If we want to remove this I'm fine with that. Those had to learn this stuff (myself included) I think have learned it and we've put it in the radiation code.

Issue #494, opened 2024-10-15
Positivity fix and Boltzmann Electrons


Maxwell-Rosen:
The positivity fix does not work with Boltzmann Electrons at the moment. There is an issue with the quazineutrality update

`
Thread 1 "sim" received signal SIGSEGV, Segmentation fault.
0x0000155547628f48 in gkyl_positivity_shift_gyrokinetic_quasineutrality_scale_cu(gkyl_positivity_shift_gyrokinetic *, const gkyl_range *, const gkyl_range *, const gkyl_array * __restrict__, const gkyl_array * __restrict__, const gkyl_array * __restrict__, const gkyl_array * __restrict__, gkyl_array * __restrict__) (up=0x1e52f670, conf_rng=0x15079e8, phase_rng=0x1829aa40, delta_m0s=0x1e52c950, delta_m0s_tot=0x1e539630, delta_m0r=0x7900, m0s=0x17a613d0, fs=0x17a673e0) at zero/positivity_shift_gyrokinetic_cu.cu:304
304	  gkyl_positivity_shift_gyrokinetic_quasineutrily_scale_cu_ker<<<nblocks, nthreads>>>(
(cuda-gdb) bt
#0  0x0000155547628f48 in gkyl_positivity_shift_gyrokinetic_quasineutrality_scale_cu(gkyl_positivity_shift_gyrokinetic *, const gkyl_range *, const gkyl_range *, const gkyl_array * __restrict__, const gkyl_array * __restrict__, const gkyl_array * __restrict__, const gkyl_array * __restrict__, gkyl_array * __restrict__) (up=0x1e52f670, conf_rng=0x15079e8, phase_rng=0x1829aa40, 
    delta_m0s=0x1e52c950, delta_m0s_tot=0x1e539630, delta_m0r=0x7900, m0s=0x17a613d0, 
    fs=0x17a673e0) at zero/positivity_shift_gyrokinetic_cu.cu:304
#1  0x0000155542a95b1b in gkyl_positivity_shift_gyrokinetic_quasineutrality_scale (
    up=0x1e52f670, conf_rng=conf_rng@entry=0x15079e8, phase_rng=0x1829aa40, 
    delta_m0s=0x1e52c950, delta_m0s_tot=0x1e539630, delta_m0r=0x7900, m0s=0x17a613d0, 
    fs=0x17a673e0) at zero/positivity_shift_gyrokinetic.c:157
[...]

PR #502, opened 2024-10-23
Single species positivity fix

Maxwell-Rosen:
There is a single line of code changed here. I think that the positivity fix should be fine by just skipping the electron loading in line 2431. The else statement in the loop should just evaluate everything to 1 because it's not loading the moments of the electrons. Addressing issue #494


ammarhakim:
This is somewhat a strange way to do things. What if a ion species has negative charge? For example, it is possible we have negative ion species, a problem, in fact, for which we are preparing a proposal to AFOSR for thruster modeling.

manauref (approved):

JunoRavin (approved):

PR #505, opened 2024-10-24
MB app update and other methods

manauref:
Add the GK multiblock (MB) update method, and I/O methods that mimic the single block methods. These changes are the minimum necessary for the MB to run with a single block, producing identical results to the single block app, while also setting a path for running with multiple blocks with minimal changes.

**Update method:**

This required splitting the forward Euler in the single block app so that we can compute df/dt (in the _rhs method) without performing an actual time step. The single block _rhs method still reduces the CFL frequency in its block, and returns a minimum dt, `dt_min` (we may want to change/move this, more on this below). 

The MB update calls the MB ssp_rk3, and in each stage it does 4 things:
1. It loops over the blocks, calling gyrokinetic_rhs for each. Again, this produces a min dt for each block.
2. Perform a reduction over all ranks (with app->comm) of the min dt.
3. Do a forward euler in each block (and for each species).
4. Call the field solver and apply BCs.

There is something that's not ideal about the current workflow: gyrokinetic_rhs does a reduction of `dt_min` over the block (with allreduce). But then the MB forward_euler, when reducing `dt_min` over all blocks that are not local, it does so using the MB app communicator (`app->comm`) which contains all the ranks. This means that a lot of the ranks participate, unnecessarily, in two reductions. For now this seems fine and is just a likely small inefficiency, but one easy improvement might be to move the reduction over a single block out of gyrokinetic_rhs, and allow the MB app to perform a single reduction over all ranks in every block at once.

Lastly, because we don't yet have a MB field object, step 4 above is currently just using the field object inside the single block app, since this PR is only meant to support a single block for now.

**I/O methods:**

We gave the GK MB API nearly every I/O method that the single block GK app has, allowing for granularity in outputting, restarting, and reporting timings. Many of this _write methods in the MB app simply loop over blocks and call the corresponding method of the single block app in each block.

There's one notable feature in the current implementation: integrated diagnostics are written out per block, i.e. a 2-block step simulation would produce
```
multib_step_2x2v_p1_b0-field_energy.gkyl
multib_step_2x2v_p1_b1-field_energy.gkyl
```
This is because the I/O is much simpler if we do it this way for now, and in postprocessing one can simply add these two dynvectors. Performing the reduction over blocks will be trickier and is postponed until we have a multiblock comm object with more shape.

Tests performed
----------------

Since this is meant to only support a single block sim for now we created the `rt_multib_step_b2_2x2v_p1` regression test, which contains block 2 of a 12-block double null STEP simulation and is meant to be equivalent to the `rt_gk_step_out_2x2v_p1` test (which runs through the single block app).

- The initial conditions for these two tests are identical.
- When run in serial these two tests produce identical results.
- When run in parallel these two tests produce identical results.
- When restarted, in serial or parallel, these two tests produce identical results.

An additional note
------------------
The current GK MB API doesn't give us the flexibility we need for setting BCs for the field solver, so to make the above tests work we added fem_parbc to the input structs and we leverage the fact that the DIRICHLET value is 0, but we are not providing that in the input file (because we can't), the code is simply using some variable that is initialized as 0. I'll describe this problem and possible solutions in a DR.

### DIFF
--- a/apps/gkyl_gyrokinetic_multib.h
+++ b/apps/gkyl_gyrokinetic_multib.h
@@ -169,7 +169,6 @@ struct gkyl_gyrokinetic_multib {
   struct gkyl_comm *comm;  
 };
 
-
 /**
  * Construct a new gk multi-block app.
  *
@@ -209,7 +208,6 @@ void gkyl_gyrokinetic_multib_app_apply_ic_species(gkyl_gyrokinetic_multib_app* a
  * @param t0 Time for initial conditions
  */
 void gkyl_gyrokinetic_multib_app_apply_ic_neut_species(gkyl_gyrokinetic_multib_app* app, int sidx, double t0);
-
 
 /**
  * Initialize field from file
--- a/apps/gkyl_gyrokinetic_multib.h
+++ b/apps/gkyl_gyrokinetic_multib.h
@@ -687,18 +687,18 @@ void gkyl_gyrokinetic_multib_app_write_conf(gkyl_gyrokinetic_multib_app* app, do
 void gkyl_gyrokinetic_multib_app_write(gkyl_gyrokinetic_multib_app* app, double tm, int frame);
 
 /**
- * Read geometry file.
- *
- * @param app App object.
- */
-void gkyl_gyrokinetic_multib_app_read_geometry(gkyl_gyrokinetic_multib_app *app);
-
-/**
  * Write stats to file. Data is written in json format.
  *
  * @param app App object.
  */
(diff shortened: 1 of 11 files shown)


ammarhakim:
I would like the code in this PR to be presented so a proper review can be done. Specially as we did not have a corresponding DR it is hard to fully understand. This is critical code and we need to get it right before we merge it. We should avoid a redo later.

manauref:
Sure, I can present it at an upcoming meeting.

Issue #391, opened 2024-06-17
[DR] Extending radiation to be dependent on electron density


jRoeltgen:
While the electron density dependence for line radiation is generally weak, some (such as Li<sup>0</sup>) have a significant dependence. To include this constructing separate drag terms (vnu, vsqnu, vnu_surf, and vsqnu_surf) for each density is proposed. The drag with the nearest density is chosen cellwise when constructing the overall drag (nvnu, nvsqnu, nvnu_surf, nvsqnu_surf) in dg_calc_dg_rad_vars.c: 

`   const double* ne = gkyl_array_cfetch(n_elc, loc_conf);
    double ne_cell_avg = ne[0]/pow(2.0, cdim/2.0);
    int ne_idx = gkyl_find_nearest_idx(n_elc_rad, ne_cell_avg);
    
    const double* vnu_surf_d = gkyl_array_cfetch(vnu_surf[ne_idx], loc_phase);
    const double* vnu_d = gkyl_array_cfetch(vnu[ne_idx], loc_phase);
    const double* vsqnu_surf_d = gkyl_array_cfetch(vsqnu_surf[ne_idx], loc_phase);  
    const double* vsqnu_d = gkyl_array_cfetch(vsqnu[ne_idx], loc_phase); ` 

The number of densities will be an input by the user, though no change in the input files is needed as num_of_densities already exists in the radiation structure. The densities are then evenly spaced in the number of fits with num_of_densities=1 being the closest to n<sub>e</sub>=10<sup>19</sup>, since what density the fits exist for, is not known until run time.  Current radiation input:

` .radiation = {

      .radiation_id = GKYL_GK_RADIATION, 
      .num_cross_collisions = 0,
      .collide_with = { "Ar0", "Ar4", "Ar8", "Li0", "Li1", "Li2", "C0", "C1", "C2", "C3", "C4", "C5"},
      .z = {18, 18, 18, 3, 3, 3, 6, 6, 6, 6, 6, 6},
      .charge_state = {0, 4, 8, 0, 1, 2, 0, 1, 2, 3, 4, 5, 6},
      .num_of_densities = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}, // Must be 1 for now                                                                                           
      },
`
[...]

PR #509, opened 2024-10-29
Gk g0 app radiation ne dependence

jRoeltgen:
Extending the radiation to allow for multiple densities based on the design review: [https://github.com/ammarhakim/gkylzero/issues/391]. The array of arrays is implemented as a struct of arrays.


apps/gk_species_radiation.c
```diff
@@ -44,25 +46,57 @@ gk_species_radiation_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s
...
+    if (num_dens_per_coll[i] > max_num_densities)
+      max_num_densities = num_dens_per_coll[i];
+  }
+  
   // Make array for cutoff below which radiation is set to 0. Set to 1eV. Keep the radiation from driving Te negative. Only lives on host.
-  rad->vtsq_min_normalized = mkarr(false, 1, rad->num_cross_collisions);
+  rad->vtsq_min_normalized = mkarr(app->use_gpu, rad->num_cross_collisions, max_num_densities);
+  struct gkyl_array* vtsq_min_normalized_host = mkarr(false, rad->num_cross_collisions, max_num_densities);
+  rad->vtsq_min_per_species = mkarr(app->use_gpu, 1, max_num_densities); 
+  double* vtsq_normalized = (double*)malloc(rad->num_cross_collisions*max_num_densities*sizeof(double));
```
manauref, line 86: don't use malloc and free, use gkyl_malloc and gkyl_free.

apps/gk_species_radiation.c
```diff
@@ -178,32 +219,30 @@ gk_species_radiation_moms(gkyl_gyrokinetic_app *app, const struct gk_species *sp
...
-    const double* vtsq_min_normalized_d = gkyl_array_cfetch(rad->vtsq_min_normalized, i);    
-    gkyl_dg_calc_gk_rad_vars_nI_nu_advance(rad->calc_gk_rad_vars[i], 
-      &app->local, &species->local, 
-      rad->vnu_surf[i], rad->vnu[i], 
-      rad->vsqnu_surf[i], rad->vsqnu[i], 
-      rad->moms[i].marr, 
-      rad->nvnu_surf, rad->nvnu, 
-      rad->nvsqnu_surf, rad->nvsqnu,
-      vtsq_min_normalized_d[0], rad->vtsq);
+    gkyl_array_set_offset(rad->vtsq_min_per_species, 1.0, rad->vtsq_min_normalized, i);
```
manauref, line 239: This and line 282 below seem avoidable because these things don't change in time/

Maybe instead of 
```
gkyl_array *vtsq_min_per_species;
```
and 
```
vtsq_min_per_species = mkarr(use_gpu, num_coll, num_dens)
```
it'd be better to create an array of arrays here, with
```
gkyl_array **vtsq_min_per_species;
vtsq_min_per_species = gkyl_malloc(numcoll*sizeof(struct gkyl_array *));
for i = 1, num_dens {
  vtsq_min_per_species[i] = mkarr(use_gpu, num_dens, 1)
}
```
and you'll have to reorganize the gkyl_array_copy and/or some memcpy so that you populate `vtsq_min_per_species` one species at a time. This way you can change avoid the `gkyl_array_set_offset(rad->vtsq_min_per_species` every time, and L245 and L291 would use `rad->vtsq_min_per_species[i]` instead.

data/adas/read_radiation.py
```diff
@@ -0,0 +1,239 @@
+import os
```
manauref, line 1: this file has barely any comments, and no header describing its purpose or what it does. Please at some description near the top

zero/radiation_read.c
```diff
@@ -1,3 +1,4 @@
+#include <gkyl_util.h>
 #include <gkyl_radiation_read.h>
```
manauref, line 2: are there unit tests that just test radiation_read functions? do we know if it is valgrind clean?

apps/gk_species_radiation.c
```diff
@@ -44,25 +46,57 @@ gk_species_radiation_init(struct gkyl_gyrokinetic_app *app, struct gk_species *s
...
   }
 
+  // Updater to compute drag coefficients.
+  rad->calc_gk_rad_vars = gkyl_dg_calc_gk_rad_vars_new(&s->grid, &app->confBasis,
+	&app->basis, s->info.charge, s->info.mass, app->gk_geom, s->vel_map, app->use_gpu);
+
   // Fitting parameters
-  double a[26], alpha[26], beta[26], gamma[26], v0[26];
-  struct all_radiation_states *rad_data=gkyl_read_rad_fit_params();
+  double a[GKYL_MAX_RAD_DENSITIES], alpha[GKYL_MAX_RAD_DENSITIES], beta[GKYL_MAX_RAD_DENSITIES],
```
manauref, line 54: This feedback may be a bit delayed, but please consider changing this n0 and this ne variables to something more descriptive and easier to work with, given that:
1. Single letter variables should not exist (except a temporary variable in a small piece of code, like the for the counter of a small loop).
2. ne is pretty generic. It could mean the electron density, but that is x and t dependent.
Perhaps all these fit parameters should be something like radfit_a for a, etc.

manauref (changes requested):
One additional request:
I found out today that the LBO was using J*M0 in places where it meant to use M0. Please go through the rad app and check that it is not making this same mistake (recall that we evolve J*f, and computing the moments of it gives J*M0, J*M1, J*M2).

zero/radiation_read.c
jRoeltgen: There aren't any unit tests for radiation_read specifically, but all of the functions in it are used by ctest_dg_rad_gyrokinetic. I think it is valgrind clean (it has 1 reachable piece of memory that says it is in acutest).

manauref (changes requested):
Almost ready. Two small requests:
1. gk_species_rad allocates a tremendous amount of memory. I see at least 2 phase-space arrays we'd like to avoid allocating, `integrated_moms_rhs` and `emissivity_rhs`, which are only used for diagnostics. You might be able to find other phase-space arrays you can temporarily use instead of allocating these. Or at the very least consolidate these 2 into 1.
2. `gkyl_find_nearest_idx` was added to gkyl_array.h. Three issues: a) it's name doesn't start with gkyl_array_, b) it was not unit tested in unit/ctest_array, c) it assumes the number of components in the array is 1, so it is not applicable to a gkyl_array. Either address these 3 or just move this function to the radiation code so it is only used/defined there (as a static probably).

jRoeltgen:
@manauref pointed out that a large number of phase space grids are being allocated for the radiation. If this causes it to run out of GPU memory, but not CPU memory, maybe the loop over configuration space in gkyl_dg_calc_gk_rad_vars_nI_nu_advance could only be done on the host. That way the correct density of vnu can be chosen before sending it to the gpu and a maximum of 2 densities per species would be needed on the GPUs.

manauref (approved):

JunoRavin (approved):

PR #511, opened 2024-10-31
Gk g0 app blocked sol comms fieldsolve

akashukla:
Field solve for blocks connected in parallel is ready. If the polarization_bmag used in each block of a 3-block SOL setup is the same, the 3 block setup will produce the same phi as a single block setup.

### DIFF
--- a/apps/gk_field_multib.c
+++ b/apps/gk_field_multib.c
@@ -60,27 +60,27 @@ gk_field_multib_new(const struct gkyl_gyrokinetic_multib *mbinp, struct gkyl_gyr
   }
 
   // Construct the local and global ranges for the allgather
-  mbf->multib_z_ranges = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_range *));
-  mbf->multib_z_ranges_ext = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_range *));
+  mbf->multibz_ranges = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_range *));
+  mbf->multibz_ranges_ext = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_range *));
 
   int nghost[] = { 1, 1 };
   for (int bI= 0; bI<num_local_blocks; bI++) {
     int bid = local_blocks[bI];
-    mbf->multib_z_ranges[bI] = gkyl_malloc( sizeof(struct gkyl_range));
-    mbf->multib_z_ranges_ext[bI] = gkyl_malloc( sizeof(struct gkyl_range));
-    gkyl_multib_comm_conn_create_multib_ranges_in_dir(mbf->multib_z_ranges_ext[bI], mbf->multib_z_ranges[bI], nghost, nconnected[bid], block_list[bid], dir, mbapp->decomp);
+    mbf->multibz_ranges[bI] = gkyl_malloc( sizeof(struct gkyl_range));
+    mbf->multibz_ranges_ext[bI] = gkyl_malloc( sizeof(struct gkyl_range));
+    gkyl_multib_comm_conn_create_multib_ranges_in_dir(mbf->multibz_ranges_ext[bI], mbf->multibz_ranges[bI], nghost, nconnected[bid], block_list[bid], dir, mbapp->decomp);
   }
 
   // Allocate local and global arrays for charge density
-  mbf->rho_c_local_dg = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_array*));
-  mbf->rho_c_global_dg = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_array*));
-  mbf->rho_c_global_smooth = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_array*));
+  mbf->rho_c_local = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_array*));
+  mbf->rho_c_multibz_dg = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_array*));
+  mbf->rho_c_multibz_smooth = gkyl_malloc(num_local_blocks* sizeof(struct gkyl_array*));
 
   for (int bI=0; bI<mbapp->num_local_blocks; ++bI) {
     struct gkyl_gyrokinetic_app *sbapp = mbapp->singleb_apps[bI];
-    mbf->rho_c_local_dg[bI] = gkyl_array_acquire(sbapp->field->rho_c);
-    mbf->rho_c_global_dg[bI] = mkarr(mbapp->use_gpu, sbapp->confBasis.num_basis, mbf->multib_z_ranges_ext[bI]->volume);
-    mbf->rho_c_global_smooth[bI] = mkarr(mbapp->use_gpu, sbapp->confBasis.num_basis, mbf->multib_z_ranges_ext[bI]->volume);
+    mbf->rho_c_local[bI] = gkyl_array_acquire(sbapp->field->rho_c);
+    mbf->rho_c_multibz_dg[bI] = mkarr(mbapp->use_gpu, sbapp->confBasis.num_basis, mbf->multibz_ranges_ext[bI]->volume);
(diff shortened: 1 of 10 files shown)


akashukla:
Since this made a change to the single block app (smoothe rho rather than J*rho), here is a comparison of the field and other quantities for many  regression tests.

I am plotting the field from the final frame. It doesn't appear to change significantly. "Old" is gk-g0-app. "New" is this branch.

Solovev:

image


LTX:

image


Sheath:
image



Mirror:

image




LAPD:
image

akashukla:
Also, single block SOL and 3-block SOL initial fields are the same given the same density ICs:

image

apps/gkyl_gyrokinetic_multib_priv.h
```diff
@@ -58,6 +58,29 @@ struct gyrokinetic_multib_output_meta {
...
 
+
+// field data
+struct gk_field_multib {
+  struct gkyl_gyrokinetic_multib_field info; // data for field
+  enum gkyl_gkfield_id gkfield_id; // type of field
+
+
+  struct gkyl_multib_comm_conn **mbcc_send; // comm object for allgather
+  struct gkyl_multib_comm_conn **mbcc_recv; // comm object for allgather
```
manauref, line 69: these two have the same descriptive comment, please change.

Also, I think we should rename these with more specificity because we may need more mbcc's for other operations (e.g. perp solve), so maybe call them `mbcc_allgatherz_`

apps/gkyl_gyrokinetic_multib_priv.h
```diff
@@ -58,6 +58,29 @@ struct gyrokinetic_multib_output_meta {
...
+  struct gkyl_range **multibz_ranges; // ranges for allgather along z
+  struct gkyl_range **multibz_ranges_ext; // extended ranges for smoothing along z
+  struct gkyl_range **block_subranges; // ranges for copying smooth charge density
```
manauref, line 72: same thing here, specify this is along z, maybe call it `block_subrangesz` ?
In general keep in mind that we soon want to parallelize in x-y too, so we'll have to do some gathers/copies in other directions too. By naming things with specificity you make that other task easier

apps/gkyl_gyrokinetic_multib_priv.h
```diff
@@ -89,3 +112,19 @@ void gyrokinetic_multib_calc_field_and_apply_bc(struct gkyl_gyrokinetic_multib_a
...
+// Initialize multib field object
+struct gk_field_multib* gk_field_multib_new(const struct gkyl_gyrokinetic_multib *mbinp,
+    struct gkyl_gyrokinetic_multib_app *mbapp);
+
+
+// Compute the electrostatic potential
+void gk_field_multib_rhs(gkyl_gyrokinetic_multib_app *mbapp, 
+    struct gk_field_multib *mbf, const struct gkyl_array *fin[]);
+
+// Release resources for multib field
```
manauref, line 129: The comments for these functions are incomplete

apps/gkyl_multib_comm_conn.h
```diff
@@ -123,7 +123,7 @@ int gkyl_multib_comm_conn_array_transfer(struct gkyl_comm *comm,
  * @param dir direction in which blocks are connected
  * @param decomp List of decomposition objects for each block
  */
-void create_cross_ranges_in_dir(struct gkyl_range *cross_range_ext,
+void gkyl_multib_comm_conn_create_multib_ranges_in_dir(struct gkyl_range *cross_range_ext,
```
manauref, line 126: rename cross_range in these functions and comments

unit/mctest_multib_allgather.c
```diff
@@ -961,13 +1046,209 @@ test_L_domain_allgather_dir0_cuts2_par()
...
+    cells[0] = gkyl_range_shape(global_ranges[bI],0);
+    cells[1] = gkyl_range_shape(global_ranges[bI],1);
+    double gridlo[2] = {0.0,0.0};
+    double l0 = ((double) cells[0])/4.0;
+    double l1  = ((double) cells[1])/8.0;
+    double gridup[2] = {l0,l1 };
+    gkyl_rect_grid_init(&grid, 2, gridlo, gridup, cells);
+    char str[50];
+    sprintf(str, "lb%d_r%d.gkyl",bI, my_rank);
+    gkyl_grid_sub_array_write(&grid, global_ranges[bI], 0, array_global[bI], str);
```
manauref, line 1221: this isn't really a unit test if it requires the user to plot the result and check by eye. Can you implement an automatic correctness test?

unit/mctest_multib_allgather.c
```diff
@@ -961,13 +1046,209 @@ test_L_domain_allgather_dir0_cuts2_par()
...
+
 
 TEST_LIST = {
   //{ "test_L_domain_send_connections_dir0_cuts1", test_L_domain_send_connections_dir0_cuts1},
   //{ "test_L_domain_recv_connections_dir0_cuts1", test_L_domain_recv_connections_dir0_cuts1},
   //{ "test_L_domain_send_connections_dir0_cuts2", test_L_domain_send_connections_dir0_cuts2},
   //{ "test_L_domain_send_connections_dir0_cuts2_par", test_L_domain_send_connections_dir0_cuts2_par},
   //{ "test_L_domain_recv_connections_dir0_cuts2_par", test_L_domain_recv_connections_dir0_cuts2_par},
-  { "test_L_domain_allgather_dir0_cuts2_par", test_L_domain_allgather_dir0_cuts2_par}, // 2 ranks
+  //{ "test_L_domain_allgather_dir0_cuts2_par", test_L_domain_allgather_dir0_cuts2_par},
```
manauref, line 1251: can we enable the L domain tests?

apps/gk_field_multib.c
```diff
@@ -0,0 +1,229 @@
+#include <gkyl_alloc.h>
```
manauref, line 1: you have huge lines in this file, please shorten

also, maybe we should call this gk_multib_field, that way if we create more gk_ objects for MB, they'll be grouped together in apps/ (also for the app we placed multib after gyrokinetic)

apps/gk_field_multib.c
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+
+
+
+  // Construct the comm_conns for the allgather
+  mbf->mbcc_send = gkyl_malloc(num_local_blocks * sizeof(struct gkyl_multib_comm_conn *));
+  mbf->mbcc_recv = gkyl_malloc(num_local_blocks * sizeof(struct gkyl_multib_comm_conn *));
+  for (int bI= 0; bI<num_local_blocks; bI++) {
+      int bid = local_blocks[bI];
```
manauref, line 93: wrong indent

apps/gk_field_multib.c
```diff
@@ -0,0 +1,229 @@
...
+    gkyl_free(mbf->multibz_ranges_ext[bI]);
+    gkyl_free(mbf->block_subranges[bI]);
+    gkyl_array_release(mbf->rho_c_local[bI]);
+    gkyl_array_release(mbf->rho_c_multibz_dg[bI]);
+    gkyl_array_release(mbf->rho_c_multibz_smooth[bI]);
+    gkyl_multib_comm_conn_release(mbf->mbcc_send[bI]);
+    gkyl_multib_comm_conn_release(mbf->mbcc_recv[bI]);
+  }
+
+    gkyl_free(mbf->multibz_ranges);
```
manauref, line 219: wrong indent

manauref (changes requested):

akashukla:
This is not ready to merge yet. Specifically, the change to smoothing rho rather than J*rho needs a bit more discussion. Closing this for now.

unit/mctest_multib_allgather.c
manauref: If it's too difficult to make it for any decomp (I suspect it is, and it may be unnecessary) just make it work for at least one decomp, and check that the test is run that way (we do that for MPI/NCCL tests).

PR #525, opened 2024-11-13
Fix VP restart

manauref:
Somehow it slipped under our noses that there were issues with VP restarts. They were working in an earlier iteration, but at some point they broke because they were calling a VM method that doesn't exist for VP.

Fix restarts by:
- Putting an if-statement in the app_read_from_frame method checking that we are not doing VP before calling `vm_field_apply_bc`.
- If doing VP, compute the initial field, external fields, and apply BCs.

Tested with:
- rt_vlasov_weibel
- rt_vlasov_poisson_sheath_Bext
both sims ran with a single run, or with 1 restart, getting visually identical field-energy time traces in each case.

### DIFF
--- a/apps/vlasov.c
+++ b/apps/vlasov.c
@@ -1391,10 +1391,16 @@ gkyl_vlasov_app_from_file_fluid_species(gkyl_vlasov_app *app, int sidx,
 struct gkyl_app_restart_status
 gkyl_vlasov_app_from_frame_field(gkyl_vlasov_app *app, int frame)
 {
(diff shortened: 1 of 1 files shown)


JunoRavin (approved):

manauref:
For the record, in this PR, if you run vp_sheath_Bext with and without a restart you'll see a tiny difference in the field-energy time trace. That is an I/O resolution thing; if you calculate field-energy every time step both sims (w/ and w/o restart) come out identical

kltrbrds (approved):

PR #527, opened 2024-11-13
Gyrokinetic electrostatic field that does not evolve in time

Maxwell-Rosen:
I added a small flag to fix the potential at t=0 and never evolve it on subsequent time steps. It's only 4 lines of code. Regression tests pass on CPU and the code builds just fine.


apps/gyrokinetic.c
```diff
@@ -445,24 +445,26 @@ static void
...
-
-    // Compute biased wall potential if present and time-dependent.
-    // Note: biased wall potential use eval_on_nodes. 
-    // so does copy to GPU every call if app->use_gpu = true.
-    if (app->field->phi_wall_lo_evolve || app->field->phi_wall_up_evolve)
-      gk_field_calc_phi_wall(app, app->field, tcurr);
-
-    // Solve the field equation.
-    gk_field_rhs(app, app->field);
+    if (app->field->is_static == false || app->stat.nup == 0) {
```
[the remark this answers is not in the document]
Maxwell-Rosen: Moved to the initilization step and removed the need for the statistics information

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -751,6 +752,7 @@ struct gkyl_gyrokinetic_app {
   struct gk_geometry *gk_geom;
 
   bool update_field; // are we updating the field?
+  bool is_field_static; // is the field static? Take dphi/dt = 0 and use phi(t=0) as initial condition.
```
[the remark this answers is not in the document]
Maxwell-Rosen: Removed

Maxwell-Rosen:
I discussed using the update_field flag for this feature with Jimmy on Friday. Although it sounded like a good idea at first, further investigation revealed that this flag also removes the field diagnostics for each frame, which is important. Although the field at the initial conditions is the same as the future frames, this data will not get written because writing the data is immediately before the time stepping loop, after the initial conditions are determined.

JunoRavin (approved):

apps/gyrokinetic.c
```diff
@@ -560,6 +560,7 @@ gkyl_gyrokinetic_app_apply_ic(gkyl_gyrokinetic_app* app, double t0)
     }
   }
   calc_field_and_apply_bc(app, 0., distf, distf_neut);
+  app->field->is_static = app->field->info.is_static;
```
manauref, line 563: @Maxwell-Rosen did you check that existing regression tests work with these changes? i.e. when is_static is not specified in the input file?

regression/rt_gk_wham_1x2v_p1.c
```diff
@@ -865,6 +865,7 @@ int main(int argc, char **argv)
     .polarization_bmag = ctx.B_p, 
     .fem_parbc = GKYL_FEM_PARPROJ_NONE,
     .kperpSq = pow(ctx.kperp, 2.),
+    .is_static = true,
```
manauref, line 868: i don't think this belongs here. In general we don't wish to run with static fields, it is only a dev tool, and we want reg tests to test production functionality

manauref (changes requested):

Maxwell-Rosen:
This test produces a field that does not evolve in time. The plot below is the DG coefficients of the field at the beginning and end of the unit test.
image

apps/gyrokinetic.c
Maxwell-Rosen: When not specified, is_static should be set to false by default. Is there a different / more robust way to set this?

JunoRavin (approved):

apps/gyrokinetic.c
```diff
@@ -490,27 +523,7 @@ gkyl_gyrokinetic_app_new(struct gkyl_gk *gk)
...
-    // Compute biased wall potential if present and time-dependent.
-    // Note: biased wall potential use eval_on_nodes. 
-    // so does copy to GPU every call if app->use_gpu = true.
-    if (app->field->phi_wall_lo_evolve || app->field->phi_wall_up_evolve)
-      gk_field_calc_phi_wall(app, app->field, tcurr);
-
-    // Solve the field equation.
-    gk_field_rhs(app, app->field);
-  }
+  app->calc_field_func(app, tcurr, fin);
```
Maxwell-Rosen, line 526: It seems redundant for a function to only wrap a single function. Is this neccisary?

manauref:
Yes it is necessary, because the function (pointer) that is called within that function gets chosen at run time based on input file parameters.One could instead call the function pointer itself, but I believe that design is incorrect. We interact with objects via their API (i.e. their functions), not via their members, which the function pointer is.On Jan 24, 2025, at 1:38 PM, Maxwell Rosen ***@***.***> wrote:﻿
@Maxwell-Rosen commented on this pull request.



In apps/gyrokinetic.c:
> -    // done here as the RHS update for all species should be complete before
-    // boundary fluxes are computed (ion fluxes needed for sheath values) 
-    // and these boundary fluxes are stored temporarily in ghost cells of RHS
-    if (app->field->gkfield_id == GKYL_GK_FIELD_BOLTZMANN)
-      gk_field_calc_ambi_pot_sheath_vals(app, app->field);
-
-    // Compute biased wall potential if present and time-dependent.
-    // Note: biased wall potential use eval_on_nodes. 
-    // so does copy to GPU every call if app->use_gpu = true.
-    if (app->field->phi_wall_lo_evolve || app->field->phi_wall_up_evolve)
-      gk_field_calc_phi_wall(app, app->field, tcurr);
-
-    // Solve the field equation.
-    gk_field_rhs(app, app->field);
-  }
+  app->calc_field_func(app, tcurr, fin);

It seems redundant for a function to only wrap a single function. Is this neccisary?

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -670,6 +673,7 @@ struct gk_field {
 
   struct gkyl_array *phi_host;  // host copy for use IO and initialization
 
+  bool is_static; // is the field static? Take dphi/dt = 0 and use phi(t=0) as initial condition.
```
JunoRavin, line 676: Is this boolean leftover? Now that is_static is just meant to be something in the input file and it looks like the two relevant flags are update_field and calc_init_field?

apps/gyrokinetic.c
```diff
@@ -593,21 +605,39 @@ gkyl_gyrokinetic_app_apply_ic(gkyl_gyrokinetic_app* app, double t0)
...
+  for (int i=0; i<app->num_species; ++i) {
+    gk_species_apply_bc(app, &app->species[i], distf[i]);
+  }
+  for (int i=0; i<app->num_neut_species; ++i) {
+    if (!app->neut_species[i].info.is_static) {
+      gk_neut_species_apply_bc(app, &app->neut_species[i], distf_neut[i]);
+    }
+  }
+
+  app->field->is_static = app->field->info.is_static;
```
JunoRavin, line 640: Seems like we don't need the is_static flag in field anymore between the update_field and calc_init_field flags?

manauref (approved):

Issue #519, opened 2024-11-08
Remove vtsq min calculation from gk_species app and into neutral collisions apps


tnbernard:
In the gk-g0-app branch, vtsq min is currently calculated in app/gk_species.c at lines 421-424 using an outdated formulation. This must be removed and the calculation will be updated following what's done in the LBO and radiation apps.

manauref:
Resolved in PR #533

PR #529, opened 2024-11-14
Move the vtsq_min calculation in the species to the react app. The sp…

manauref:
…ecies shouldn't have it because what vtsq_min to use may be different from one operator to another.

This also addresses issue #519


tnbernard (commented):
The simple regression test rt_gk_sheath_3x2v_p1_cons gives the exact same results as gk-g0-app branch. rt_gk_step_out_2x2v_p1 gives relative differences ~0.01, but this is likely due to vtsq_min being calculated incorrectly before.

PR #531, opened 2024-11-14
Add a simple model of FLR effects in the GK app

manauref:
The simple model of finite Larmor radius (FLR) effects implemented here is summarized in issue #532 .

This PR also adds the ability to:
1. Specify spatially varying BCs for Poisson solves in fem_poisson and deflated_fem_poisson.
2. Specify "biasing planes" in fem_poisson, which enforces a particular value of the solution on specified planes (points in 1x, lines in 2x, planes in 3x).

If the user wants to turn on FLR effects they must specify the `.flr` entry in the species table of the input file, for example:
```
    .flr = {
      .type = GKYL_GK_FLR_PADE_CONST,
      .Tperp = ctx.Ti,
    },
```
which tells the solver which model of FLR to use (we envision a hierarchy of them in the future) and some relevant parameters.

If the `.flr` entry is not provided, the GK solver program flow and results are unaffected.

New unit tests were added to ctest_fem_poisson, and all tests in that file pass on CPU and GPU.

Checked gk_sheath reg tests and they produced the same results as in main, on CPU and GPU.

This goes with [PR 61 in gkylcas](https://github.com/ammarhakim/gkylcas/pull/61).


manauref:
we just use this input to specify the x coordinate of the “plane”, but if you are going to use deflated_fem_poisson, then that updater will make sure to only pass this to the lower and upper fem_poisson solvers.

Antoinehoff:
So I think we do not need to have a spatially dependent gyroradius for this to be useful. At first order, FLR effects will act as a quadratic low pass filter of cutting length $k_\perp \rho_s$. Since fluctuation scale usually increases along the radial coordinate, we can underestimate the gyroradius size so that we do not filter too much. This can be done by setting a high temperature for the `.Tperp` parameter.

By the way, here is what happens if you include FLR effects in the coarse 3x2v TCV baseline:

download-1
_Snapshot of the simulation before adding FLR effects_

download-2
_Snapshot after a restart with FLR effects and_

We see that the small scale turbulent structures are damped. We also observe that a turbulent eddie starts to see itself through the periodic boundary condition in y (not good) but this is physical, not an issue about the FLR implementation. Here I set `.Tperp=100eV` and this overestimate the FLR damping in the core that is much hotter.

apps/gk_field.c
```diff
@@ -10,6 +10,17 @@
 #include <float.h>
 #include <time.h>
 
+static void
+gk_field_invert_flr(gkyl_gyrokinetic_app *app, struct gk_field *field, struct gkyl_array *phi)
+{
+  gkyl_deflated_fem_poisson_advance(field->flr_op, phi, phi, phi);
```
Antoinehoff, line 16: So it is now implemented for the deflated poisson only right?

apps/gk_species.c
```diff
@@ -59,7 +59,8 @@ static double
 gk_species_rhs_dynamic(gkyl_gyrokinetic_app *app, struct gk_species *species,
   const struct gkyl_array *fin, struct gkyl_array *rhs)
 {
-  gkyl_array_set(species->phi, 1.0, app->field->phi_smooth);
```
Antoinehoff, line 62: Why is this removed? it is implicitly done in `species->gyroaverage`?

apps/gkyl_gyrokinetic.h
```diff
@@ -176,6 +176,14 @@ struct gkyl_gyrokinetic_react {
   bool write_diagnostics; // used to write diagnostics from neutral species
 };
 
+// Parameters in FLR effects.
+struct gkyl_gyrokinetic_flr {
+  enum gkyl_gk_flr_type type; 
+  double Tperp; // Perp temperature used to evaluate gyroradius. 
+  double bmag; // Magnetic field used to evaluate gyroradius. If not provided
```
Antoinehoff, line 183: I think it's better to mention the existence of this parameter in the DR.

zero/fem_poisson.c
```diff
@@ -314,11 +418,19 @@ void gkyl_fem_poisson_release(gkyl_fem_poisson *up)
     if (up->isdomperiodic) gkyl_cu_free(up->rhs_avg_cu);
     gkyl_cu_free(up->bcvals_cu);
     gkyl_culinsolver_prob_release(up->prob_cu);
+
+    if (up->num_bias_plane > 0)
```
Antoinehoff, line 422: @JunoRavin will be angry because of the missing '{' 🤭

zero/fem_poisson_cu.cu
```diff
@@ -116,13 +151,24 @@ fem_poisson_choose_kernels_cu(const struct gkyl_basis* basis, const struct gkyl_
...
+  fem_poisson_set_cu_biasker_ptrs<<<1,1>>>(kers, basis->b_type, dim, poly_order, bckey_d);
+
   for (int d=0; d<basis->ndim; d++) {
-    if (bcs->lo_type[d]==GKYL_POISSON_PERIODIC && bcs->up_type[d]==GKYL_POISSON_PERIODIC) { bckey[d] = 0; }
-    else if (bcs->lo_type[d]==GKYL_POISSON_DIRICHLET && bcs->up_type[d]==GKYL_POISSON_DIRICHLET) { bckey[d] = 1; }
-    else if (bcs->lo_type[d]==GKYL_POISSON_DIRICHLET && bcs->up_type[d]==GKYL_POISSON_NEUMANN) { bckey[d] = 2; }
-    else if (bcs->lo_type[d]==GKYL_POISSON_NEUMANN && bcs->up_type[d]==GKYL_POISSON_DIRICHLET) { bckey[d] = 3; }
-    else if (bcs->lo_type[d]==GKYL_POISSON_DIRICHLET && bcs->up_type[d]==GKYL_POISSON_ROBIN) { bckey[d] = 4; }
-    else if (bcs->lo_type[d]==GKYL_POISSON_ROBIN && bcs->up_type[d]==GKYL_POISSON_DIRICHLET) { bckey[d] = 5; }
+         if (bcs->lo_type[d]==GKYL_POISSON_PERIODIC          && bcs->up_type[d]==GKYL_POISSON_PERIODIC         ) { bckey[d] = 0; }
```
Antoinehoff, line 158: I like this style but I was told not to write like that so maybe this should be changed idk. I understand that it is hard to read on github

apps/gk_species.c
JunoRavin: I think this is fine and potentially better, since it converts an input phi to the form that the species expects. If anything we may want to rename species->phi to species->gyro_phi or something.

zero/deflated_fem_poisson.c
```diff
@@ -26,6 +34,12 @@ gkyl_deflated_fem_poisson_new(struct gkyl_rect_grid grid,
   up->num_solves_z = up->local.upper[up->cdim-1] - up->local.lower[up->cdim-1] + 2;
   up->d_fem_data = gkyl_malloc(sizeof(struct deflated_fem_data[up->num_solves_z]));
 
+  // Check if one of the boundaries needs a spatially varying Dirichlet BC.
+  up->isdirichletvar = false;
+  for (int d=0; d<up->cdim; d++) up->isdirichletvar = up->isdirichletvar ||
```
JunoRavin, line 39: I don't think we should be doing these types of single line nested logic statements. This particular line is very hard to read in my opinion. Just make this an if statement to check if the poisson_bc_type is GKYL_POISSON_DIRICHLET_VARYING and if it is, the input is true. 

The only advantage to these types of structures is you can break out of the conditional in one vs. two-three CPU cycles, and I really don't think we should care about this level optimization in the initialization step.

apps/gk_field.c
manauref: yes, FLR operators use deflated_fem_poisson. It's unknown what the requirements are on the discrete FLR operators are right now, so using this or fem_poisson_perp is just as good. At some point we'll have to work through the algorithm to see if something special is needed for FLR ops.

apps/gk_species.c
manauref: I believe this was not just removed but replaced by a function that:
- in the absence of FLR effects it simply assigns species->phi = field->phi_smooth.
- with FLR effects turned on it applies the corresponding FLR operator to phi_smooth to make species->phi the gyroaveraged potential for this species.

dunno whether to rename (e.g. gyro_phi) it or not. I don't feel strongly either way right now

apps/gkyl_gyrokinetic.h
manauref: done

apps/gk_species.c
Antoinehoff: hm I think gyro_phi is quite good, the only problem I would see is that people could think that we have a good model behind this but it will eventually be the case and in the equation it should be gyro_phi.

zero/deflated_fem_poisson.c
manauref: I wasn't trying to optimize, it's just checking every direction and lower/upper BCs, so it has to be done in a loop, and checking lower/upper, and it has to be a cumulative boolean.

apps/gk_species.c
Antoinehoff: Let's call it gyro_phi, it is better so that we can keep track that there is possibly a kind of gyro average that was performed on this.

manauref: k i will change

Antoinehoff (approved):

zero/deflated_fem_poisson.c
JunoRavin: Because the end result is just a true or false for up->isdirichletvar, I would argue the easier thing to understand is:

<code>
for (int d=0; d<up->cdim; d++) {
  if (poisson_bc.lo_type[d] == GKYL_POISSON_DIRICHLET_VARYING ||
     poisson_bc.up_type[d] == GKYL_POISSON_DIRICHLET_VARYING) {
       up->isdirichletvar = true;
   }
}
</code>

This structure is strictly less efficient than what you wrote (it always evaluates the logic statements in each direction instead of immediately exiting the logic statement if up->isdirichletvar has already been set to true), but I think this sort of conditional check is easier on the developer because it makes it more transparent how up->isdirichletvar is set (and avoids logic statements that have to stretch across multiple lines with ors comparing ors).

JunoRavin (approved):

Issue #523, opened 2024-11-11
[DR] Reduced Dimensionality Canonical Poisson Bracket Species


johnson452:
Based on the branch: [can_pb_reduced_dims](https://github.com/ammarhakim/gkylzero/tree/can_pb_reduced_dims). This was original branched from main, but should be merged into main and gk-g0-app branches.

This branch extends the Canonical Poisson Bracket solver to handle 1x2v, 1x3v, and 2x3v cases. (Ser p2, Hybrid p1, and tensor p1/p2). The requirement that a reduced dimensionality can be used is if the hamiltonian does not depend on the coordinate (which also implies the the metric does not depend on this coordinate too).

Since the hamiltonian depends on coordinate-momentum pairs, the code assumes q^i is paired with p_i. Therefore, the ignorable coordinate is presumed to start at the highest index. For instance:

Coordinate Momentum Pairs:
Original can-pb (unchanged): 
(1x1v) : q = [q^0], p = [p_0, p_1]
(2x2v) : q = [q^0, q^1], p = [p_0, p_1]
(3x3v) : q = [q^0, q^1, q^2], p = [p_0, p_1, p_1]

Reduced dimensionality:
(1x2v) : q = [q^0], p = [p_0, p_1]
(1x3v) : q = [q^0], p = [p_0, p_1, p_2]
(2x3v) : q = [q^0, q^1], p = [p_0, p_1, p_2]

These will be demonstrated through example regression tests below. Otherwise, this works exactly the same as the Vlasov species, and along as you hand the correct cdim and vdim, the kernels for ignorable coordinates with can-pb will be automatically selected.

Example regression tests: [Already working]
- 1D Sodshock (r, theta) (1x2v) - y-ignored Flat Space
[regression/rt_can_pb_neut_bgk_sodshock_im_1x2v_p1.c](https://github.com/ammarhakim/gkylzero/blob/can_pb_reduced_dims/regression/rt_can_pb_neut_bgk_sodshock_im_1x2v_p1.c)
- Annular disk (r, theta) (1x2v) - Theta-ignored
[regression/rt_can_pb_bgk_surf_annulus_sodshock_im_1x2v.c](https://github.com/ammarhakim/gkylzero/blob/can_pb_reduced_dims/regression/rt_can_pb_bgk_surf_annulus_sodshock_im_1x2v.c)

Planned tests [WIP]:
[...]

PR #536, opened 2024-11-17
Ignorable Coordinates for Can-pb Updater

johnson452:
See [DR#523](https://github.com/ammarhakim/gkylzero/issues/523) for details.
Merging the production branch into gk-g0-app to avoid having to merge the prototyping branch into main then into gk-g0-app.

WIP Status:
- [FIXED] Need to resolve using conf vs phase basis for kernel selection with @JunoRavin.
- [FIXED] Issue with 2x3v p2 ser (seen in the toroidal test case moments). Besides this, everything else works. I've written another 2x3v test for cylindrical coordinates that works fine. 
- [DONE] Added boundary surf kernels in space. This will enable zero flux BCs for can-pb. They seem reasonable on some non-physical tests cases. Talking with Jimmy, it seems Tess has some for physical regression tests with neutrals using these BCs which we will test after merging.
- [May Need Followup] Need to fix proj_lte for can-pb to correctly have mass factors. This includes modifying the regression/ctest hamiltonians, etc to have the correct 1/m factors and not assume m = 1. 

Checks of additions:
- Verified all old/new can-pb regression tests, can-pb unit tests, and Valgrind clean for sample regression tests.
- Verified CPU and GPU (nvcc + cc on stellar) builds.

Unrelated to my changes: 
- regression/rt_multib_step_sol_2x2v_p1.c and regression/rt_vlasov_lbo_cross_1x1v_p2.c both fail to build with only MPI.


johnson452:
Issue with 2x3v p1/p2 toroidal rt not agrreing was due to insufficient spatial resolution in theta, it's been fixed:
image

JunoRavin (approved):

tnbernard (approved):

Issue #538, opened 2024-11-20
[DR] Incorrect calculation of boundary fluxes in GK solver


manauref:
In the GK solver we compute boundary fluxes for at least two purposes:
1. Using them in the Boltzmann electron model, which obtains the sheath potential from ambipolarity at the sheath entrance: Gamma_e = Gamma_i (here Gamma_s is the particle flux of species s at the sheath entrance).
2. Computing fluxes due to recycling of ions, where the neutral flux into our domain is a function of the ion flux out of our domain.

At least for purpose 1 we were using the presently boundary flux, which is really the z surface term of the GK scaled, e.g. in 1x2v:
```math
\mathrm{bflux}_k  = \frac{2}{\Delta z} \cdot \int_{-1}^{1} \mathrm{d}v_\parallel\,\mathrm{d}\mu~\left(\psi_k \dot{z}\cdot J\cdot B\cdot f\right)_{z=\pm 1}
```
where:
- $\Delta z$: cell length along z.
- $\psi_k$: phase basis function.
- $\dot{z}$: advection speed in the z direction.
- $J$: conf-space jacobian
- $B$: magnetic field amplitude
- $f$: distribution
Later, to compute Gamma, integrate bflux over vpar-mu space, with the appropriate dimensional factors.

The above formula is incorrect if we want the physical particle flux through a surface (or in 1D, through a point). See Dhaeseleer's "Flux coordinates and magnetic field structure", equation 2.5.49a. We are missing a factor of the magnitude of the contravariant unit vector in z. The correct formula should be
```math
\mathrm{bflux}_k  = \frac{2}{\Delta z} \cdot \int_{-1}^{1} \mathrm{d}v_\parallel\,\mathrm{d}\mu~\left(\psi_k \dot{z}\cdot J\cdot |\nabla z|\cdot B\cdot f\right)_{z=\pm 1}
```
One of the consequences of this error, which @Maxwell-Rosen has observed, is that in mirror sims with nonuniform z and Boltzmann electrons the sheath potential comes out different, compared to uniform z sims, and the difference is on the order of $\ln(\vert\nabla z\vert_{\mathrm{uniform}}/\vert\nabla z_{\mathrm{nonuniform}}\vert )$ .

Because of this extra factor of $|\nabla z|$ we can't use the collisionless GK boundary surf kernels, and we'll instead have to create new kernels for these calculations. I propose to change the bflux app and updater so that it uses such new kernels.

[...]

PR #544, opened 2024-11-26
Correct calculation of GK boundary fluxes with curvilinear geometry 

manauref:
This PR addresses issue #538 . This PR also goes with PR 50 of gkylcas: https://github.com/ammarhakim/gkylcas/pull/50

In this work:
1) We give eqn objects a new boundary flux term which is like boundary flux but has an extract factor of $|e^i|$ where e^i is the dual vector along the i-th directin.
2) We generated these new boundary flux terms for GK only, since other models don't have support for curvilinear geometry.
3) Implemented a new boundary_flux updater, which calls these kernels and is similar to ghost_surf_calc, but is more fine grained because it only does one boundary at a time (the app can create one of these for each boundary if desired).
4) Plug in this new boundary_flux updater to the GK app.

In an initial test with mirror geometry and Boltzmann electrons, we see that initial potential has the same sheath potential with uniform and nonuniform z grid in this branch, which was not the case before. 

We also ran the affected tests, rt_gk_ltx_boltz_elc_1x2v and rt_gk_mirror_boltz_elc_1x2v, and produced similar shape final potentials but they were shifted due the shift in the sheath potential with the new boundary fluxes (and the number densities remained the same).


Maxwell-Rosen:
This fix helped clarify the differences between non-uniform and uniform geometry in the potentials. See how the mc2p_fix and mc2p_fix_uniform geometry line up? The line for mc2p without the fix in non-uniform geometry is labeled mc2p. There is a significant change in the densities and temperatures, too. This compares a 144 cell non-uniform grid to a 288 cell uniform grid, initialized with uniform densities. It's very promising that these two plots line up after this bug fix is implemented.

image

ammarhakim (approved):

Maxwell-Rosen:
@tnbernard. It's most important that the potentials agree. This means that the sheath BC is producing the correct potential in both uniform and non-uniform geometry cases after the fix. However, the density and temperature profiles are unchanged with this fix. It means that the non-uniform geometry produces the same moment profiles after the fix. I wouldn't expect the results from the non-uniform grid to be the same as the uniform grid because they are on two different grids.

Issue #520, opened 2024-11-09
[DR] Exact Geometry  for Tokamaks and Mirrors (for as many terms as possible)


akashukla:
In this DR I am suggesting an improvement to the way geometric quantities are calculated for tokamaks and mirrors. I would like to use exact expressions and avoid finite differences wherever possible. This change should result in a large speedup to geometry calculation as well as an improvement in accuracy.

### Overview of Geometry Creation
For geometry, we calculate the tangent vectors 
$$\mathbf e_i = \frac{\partial \mathbf x}{\partial z^i}$$.
So, there are 9 components (9 derivatives) of physical coordinates with respect to computational coordinates that need to be calculated. From these derivatives and the magnetic field strength, we can calculate all of the other necessary geometric quantities.

For mirrors and tokamaks, it makes the most sense to use cylindrical coordinates $(R,Z,\phi)$ as the physical coordinates and we use $(\psi, \alpha, \theta)$ as the computational coordinates. Here $\psi$ is poloidal flux, $\alpha$ is the field line label, and $\theta$ is the normalized poloidal arc length. We calculate the mapping from computational to physical at computational grid nodes using a combination of root finders and integrators.

Currently, in order to get the necessary derivatives, we use finite differences. This is not ideal, because finite differences are not exact. We use very small spacing for the finite differences, which makes them quite accurate, but it would be preferable to use some exact method where possible.

### Suggested Changes
For mirrors and tokamaks we need 9 derivatives: $\frac{\partial R}{\partial \psi}, \frac{\partial Z}{\partial \psi}, \frac{\partial \phi}{\partial \psi}, \frac{\partial R}{\partial \alpha}, \frac{\partial Z}{\partial \alpha}, \frac{\partial \phi}{\partial \alpha}, \frac{\partial R}{\partial \theta}, \frac{\partial Z}{\partial \theta}, \frac{\partial \phi}{\partial \theta}$. We have no analytical expression for the derivatives with respect to $\psi$, so I am not suggesting any change to the way we calculate those derivatives. However, there are analytical expressions for the other 6 derivatives which we can calculate from our exact mapping:

$$  \frac{\partial R}{\partial \theta} = sin[arctan(\frac{\partial R}{\partial Z})] s(\psi) \textrm{;     Eq 61 in the attached document} $$
[...]

PR #549, opened 2024-12-05
Exact expressions  in Tokamak and Mirror Geometry for d/dalpha and d/dtheta derivates 

akashukla:
This goes with DR #520 which  I discussed with Ammar over zoom two weeks ago.

This PR uses exact expressions  in Tokamak and Mirror Geometry for d/dalpha and d/dtheta derivates . This avoids the cost of finite differences and any inaccuracies it causes (however small they may  have been already). This reduces time spent creating geometry by a factor of ~10.

The wham and step 2x2v regression tests give correct results for all geometric quantities (I look at J, g_ij,  b_i, dxdz, and bmag). I  have tested this for some other production scale STEP calculations as well.

I think it will be important to implement this before paring down regression tests since it reduces the time spent generating geometry. It is also important to do this before making any more changes (such  as nonuniform grids)  to geometry.


JunoRavin:
This is awesome work @akashukla. Two big-picture comments: 
1. There are comments in various places which reference equations in I believe your internal notes. Either we should figure out a way to represent the mathematics in the comments with some sort of intelligible shorthand, or we need to add a document to the gkeyll documentation page ASAP with all these equations that we can refer people to (or even making the document its own page like other developer notes). 
2. At this point there are many expressions that are commented out in geometry and it's not clear why we are keeping these legacy expressions around. If we want to keep commented code for the foreseeable future, we should add comments about why certain code is commented out and what we are keeping it around for. If we can't think of a reason and the old code does not spark joy, we should delete it.

JunoRavin (approved):

manauref (approved):

akashukla:
It is stated that C=1 for our choice of coordinates in the pdf document in DR #520 as well as in Noah Mandell's thesis, but as far as I can tell is not explicitly shown. This snippet shows that C=1 for our choice of coordinates.
image (13)

Issue #548, opened 2024-12-03
[DR] G2 -> G0 Merge


JonathanGorard:
This DR is connected to Issue #545 that contains the todo-list for the completion of this DR.

# Motivation and Philosophy

We are now at the stage within the `gk-g0-app_lua_wrappers` branch where the majority of the C app code that currently exists in G0 for Moments, Vlasov (including canonical Poisson bracket), PKPM, and Gyrokinetics, is now wrapped and can be called directly from Lua. There are a couple of remaining holdouts, such as the input structs for reactions or for reading efit files in GK, or the general spacetime structs in GR, but we intend for these to be wrapped and completed within the coming days, so that soon 100% of the G0 functionality will be callable from Lua input files. The goal is that, for every C regression test that currently exists in `/regression`, there should exist a corresponding Lua version in `/regression/lua` that produces identical output (and we are rapidly approaching completion of this task).

This development presents a certain existential question: if all the features of the G0 apps can now be called directly from Lua, then what is the purpose of maintaining the G2 layer at all? Of course, a simple answer might be that there exists functionality in G2 that does not currently exist within G0. However, this simple answer also has a simple resolution: we should simply be able to copy across the old G2 functionality into the G0 repository and, with hopefully fairly minimal modifications, have it "just work". This appears to have been @ammarhakim's experience thus far in copying over the G2 `Tool` system (including the `runregression` script, the `sqlite` database, and the `man`/`woman` system), along with the G2 `Array`, `CartField`, `DynVector`, and `ZeroArray` data structures, into G0. The basic idea is that if there's a useful feature in G2 that doesn't currently exist in G0, then we should design the G0 Lua layer in such a way that it is as easy as possible to copy that feature across.

For this reason, our proposal is to complete this merge of useful G2 functionality into G0, and then to deprecate the G2 repository altogether. The existing G0 repository will then be renamed (perhaps simply to "`gkyl`"), and will become the main/sole entry point to the code, thus simplifying installation and maintenance.

[...]

PR #552, opened 2024-12-06
Intermediate merge of Lua wrappers work

JonathanGorard:
The full details are contained within Design Review #548, and will not be repeated here. This PR is intended to be an intermediate merge of the work that @ammarhakim and I have been doing in the `gk-g0-app_lua_wrappers` branch (completing the Lua wrappers for the Moments, Vlasov, PKPM, and Gyrokinetic apps, and merging certain key bits of infrastructure - such as the Tools system - from G2 into G0) into `gk-g0-app`, in preparation for the eventual merge of the `gk-g0-app` branch into `main` in the next couple of weeks. Once this intermediate merge is complete, our intention is to continue work within the Lua wrappers branch in order to complete some final tasks, namely:

- wrapping for the tokamak and mirror geometry input structs, the reaction struct, and the neutral species struct within the GK Lua app;
- wrapping for the emissive boundary condition struct within the Vlasov and GK Lua apps;
- wrapping for some final equation systems and source terms (e.g. friction terms, Braginskii terms, and general relativistic equation systems) within the Moments Lua app;
- more stringent testing of the Lua wrappers in parallel and on GPUs;

prior to the big `gk-g0-app` -> `main` merge, which will hopefully be completed in time for the holiday break. Since we aim to have most/all users running simulations via the new Lua-C API in the new year, feedback on the top-level design of the API (especially in regards to naming conventions) is particularly welcome, as for obvious reasons we would like to avoid making large changes to this in the future wherever possible. @JunoRavin has already been extremely helpful in providing suggestions in this regard.


regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -202,47 +203,87 @@ create_ctx(void)
...
-  double n_src = app->n_src;
+  double n_peak = app->n_peak;
   double Lz = app->Lz;
 
   double n = 0.0;
 
-  if (fabs(z) < 0.25 * Lz) {
-    n = n_src;
+  if (fabs(z) <= 0.25 * Lz) {
+    n = 0.5 * n_peak * (1.0 + sqrt(1.0 - (z / (0.25 * Lz)) * (z / (0.25 * Lz)))); // Electron total number density (left).
```
manauref, line 217: the "total" and "left"/"right" in this comment and the one below doesn't make sense to me. I'd remove them.

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -202,47 +203,87 @@ create_ctx(void)
...
   }
   else {
-    n = 1.0e-40 * n_src;
+    n = 0.5 * n_peak; // Electron total number density (right).
   }
 
-  // Set source number density.
+  // Set electron total number density.
```
manauref, line 223: remove total

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -202,47 +203,87 @@ create_ctx(void)
...
 void
-evalSourceUparInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT fout, void* ctx)
+evalElcTempInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT fout, void* ctx)
 {
-  // Set source parallel velocity.
+  struct sheath_ctx *app = ctx;
+
+  double Te = app->Te;
+
+  // Set electron total temperature.
```
manauref, line 234: remove total

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -202,47 +203,87 @@ create_ctx(void)
...
+  struct sheath_ctx *app = ctx;
+  double z = xn[0];
+
+  double n_src = app->n_src;
+  double Lz = app->Lz;
+
+  double n = 0.0;
+
+  if (fabs(z) < 0.25 * Lz) {
+    n = n_src; // Electron source total number density (left).
```
manauref, line 257: remove total and left/right in these comments

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -202,47 +203,87 @@ create_ctx(void)
...
+
+void
+evalElcSourceTempInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT fout, void* ctx)
 {
   struct sheath_ctx *app = ctx;
 
   double T_src = app->T_src;
 
-  // Set source temperature.
+  // Set electron source total temperature.
```
manauref, line 274: remove total

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -253,47 +294,76 @@ evalDensityInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT
...
-    n = 0.5 * n_peak * (1.0 + sqrt(1.0 - (z / (0.25 * Lz)) * (z / (0.25 * Lz))));
+    n = 0.5 * n_peak * (1.0 + sqrt(1.0 - (z / (0.25 * Lz)) * (z / (0.25 * Lz)))); // Ion total number density (left).
   }
   else {
-    n = 0.5 * n_peak;
+    n = 0.5 * n_peak; // Ion total number density (right).
   }
 
-  // Set number density.
+  // Set ion total number density.
```
manauref, line 303: remove total and left/right in these comments

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -253,47 +294,76 @@ evalDensityInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT
...
 
 void
-evalUparInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT fout, void* ctx)
+evalIonTempInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT fout, void* ctx)
+{
+  struct sheath_ctx *app = ctx;
+
+  double Ti = app->Ti;
+
+  // Set ion total temperature.
```
manauref, line 314: remove total

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -253,47 +294,76 @@ evalDensityInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT
...
+  double n = 0.0;
+
+  if (fabs(z) < 0.25 * Lz) {
+    n = n_src; // Ion source total number density (left).
+  }
+  else {
+    n = 1.0e-40 * n_src; // Ion source total number density (right).
+  }
+
+  // Set ion source total number density.
```
manauref, line 343: remove total and left/right in these comments

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -253,47 +294,76 @@ evalDensityInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT
...
+evalIonSourceTempInit(double t, const double* GKYL_RESTRICT xn, double* GKYL_RESTRICT fout, void* ctx)
 {
   struct sheath_ctx *app = ctx;
 
-  double Ti = app->Ti;
+  double T_src = app->T_src;
 
-  // Set ion temperature.
-  fout[0] = Ti;
+  // Set ion source total temperature.
```
manauref, line 354: remove total

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -332,28 +402,18 @@ bmag_func(double t, const double* GKYL_RESTRICT zc, double* GKYL_RESTRICT fout,
...
+    }
 
-    gkyl_gyrokinetic_app_calc_field_energy(app, t_curr);
-    gkyl_gyrokinetic_app_write_field_energy(app);
+    gkyl_gyrokinetic_app_write(app, t_curr, iot->curr - 1);
 
-    gkyl_gyrokinetic_app_calc_integrated_mom(app, t_curr);
+    gkyl_gyrokinetic_app_calc_integrated_mom(app, iot->curr - 1);
     gkyl_gyrokinetic_app_write_integrated_mom(app);
   }
```
manauref, line 418: the logic here was carefully chosen so that it produced the intended behavior (for both grid and integrated diagnostics) in:
1. Well behaved single job simulations
2. Restarted simulations
3. Crashing simulations
Did you check that you are reproducing the behavior we had before for these 3 scenarios?

Also, why did you remove the field_energy integrated diag?

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -376,53 +436,124 @@ main(int argc, char **argv)
...
+    ncuts *= cuts[d];
+  }
 
-  // Electron species.
+  if (ncuts != comm_size) {
+    if (my_rank == 0) {
+      fprintf(stderr, "*** Number of ranks, %d, does not match total cuts, %d!\n", comm_size, ncuts);
+    }
+    goto mpifinalize;
+  }
```
manauref, line 517: Why are you bringing these ~70 lines of code back into every regression test (I think, I haven't looked at them all, this is the first one I looked at)? why can't we package these operations into a single function like we had before?

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -487,24 +617,27 @@ main(int argc, char **argv)
...
+    .cells = { NZ },
+
+    .poly_order = ctx.poly_order,
     .basis_type = app_args.basis_type,
+    .cfl_frac = ctx.cfl_frac,
 
     .geometry = {
       .geometry_id = GKYL_MAPC2P,
-      .world = { 0.0, 0.0 },
+      .world = { 0.0, 0.0, 0.0 },
```
manauref, line 639: this doesn't make sense to me. One should only need to pass the world coordinates in the non-simulated dimensions, not all 3

regression/rt_gk_sheath_1x2v_p1.c
```diff
@@ -139,20 +141,19 @@ create_ctx(void)
...
+  double mu_max_elc = (3.0 / 2.0) * 0.5 * mass_elc * pow(4.0 * vte, 2.0) / (2.0 * B0); // Domain boundary (electron velocity space: magnetic moment direction).
   double vpar_max_ion = 4.0 * vti; // Domain boundary (ion velocity space: parallel velocity direction).
-  double mu_max_ion = (3.0 / 2.0) * 0.5 * mass_ion * pow(4.0 * vti,2) / (2.0 * B0); // Domain boundary (ion velocity space: magnetic moment direction).
+  double mu_max_ion = (3.0 / 2.0) * 0.5 * mass_ion * pow(4.0 * vti, 2.0) / (2.0 * B0); // Domain boundary (ion velocity space: magnetic moment direction).
+  int poly_order = 1; // Polynomial order.
+  double cfl_frac = 1.0; // CFL coefficient.
 
   double t_end = 6.0e-6; // Final simulation time.
   int num_frames = 1; // Number of output frames.
-  int int_diag_calc_num = num_frames*100;
```
manauref, line 148: why are you getting rid of this?

manauref (changes requested):
I haven't had a chance to go through everything, but here's some initial input. I'll add more comments later.

kltrbrds:
The structure for the moments app and the Lua input files looks good to me. Minor comment, but I'm seeing a mix of camel case and underscores in the Lua input files (in variable names, not core code). I believe originally we had attempted to maintain some consistency in the use of the former in Lua G2 and the latter in G0 even for variables, is that still the case?

akashukla (commented):
Just a  question: why are none of the mirror or tokamak input files wrapped in lua?

ammarhakim:
The tokamak and mirror tests are being completed. This is an intermediate merge so people have a chance to look at things. When all said and done we hope to have every C regression test have a corresponding Lua regression test.

regression/rt_gk_sheath_1x2v_p1.c
JonathanGorard: The "left" and "right" descriptors can certainly be removed (I think they are vestigial remnants from some other test, apologies). The "total" descriptor is a convention that I've attempted to enforce throughout the code in order to distinguish between quantities that are being assigned componentwise vs. in one shot. For instance, "Set total drift velocity" means setting all components of the drift velocity simultaneously, while "Set drift velocity (x-direction)" means only setting the x-component. There are quantities, like distribution function values, that are always single-valued in the GK case (and so will always be listed as "total", which may appear redundant), but which are not always single-valued in other cases (such as in PKPM, where the F0 and G functions are set separately). I'd always prefer to err on the side of disambiguation where possible, hence why these comments have been written in this way.
If there is a reason why GK should be regarded as a special case in this respect then I'm happy to hear it; otherwise this is the rationale for the convention.

regression/rt_gk_sheath_1x2v_p1.c
JonathanGorard: The boilerplate that has been introduced into the write method here was the boilerplate that you yourself wrote, and which was subsequently standardized across the entire G0 system. Are you saying that it was subsequently discovered that this code was inadequate in some regard (either for restarts or crashes)? If so, why wasn't it modified globally rather than just in GK?

If this is the case, then we should establish precisely what was wrong and why this change is necessary, and then make this change _globally_. And we should figure out why it wasn't caught before.

As for why the field energy integrated diagnostics were removed: there is no principled reason, and we can add these to the Lua app driver (in which case it makes sense to add them to the C tests too), but then we would need to decide whether they would be calculated and written for all tests, or whether we want some kind of Lua interface that determines when and if they are calculated and written, and a design for this needs to be established.

regression/rt_gk_sheath_1x2v_p1.c
JonathanGorard: The idea is that the program flow within a pure C test should be identical (or as close to identical as possible) to the program flow within a Lua-driven test, so as to facilitate the debugging of the Lua wrappers, since going forward this will be the primary/sole purposes of the C tests. So the simple answer is that these guards are used in the C tests because they're also what appear in the Lua program flow. It may be possible that the program flow in the Lua drivers can be simplified (it's very likely that it can), but only then should the C tests be modified to remove this.

[Incidentally, there are places within GK regression tests - especially on the multiblock side - where inadequate guarding has been used, and as a result the code does not compile without MPI, or other such pathologies have been introduced. This kind of thing is pretty disastrous, and is precisely why I would rather make such changes _very cautiously_. It is far better to have too much code than to have broken builds.]

regression/rt_gk_sheath_1x2v_p1.c
JonathanGorard: Oops, sorry - I think I added this to the Lua tests (and therefore to the C tests) as part of tracking down some Valgrind uninitialized memory error. It turned out not to be the problem. Will remove this, thanks!

regression/rt_gk_sheath_1x2v_p1.c
JonathanGorard: See above for my comment regarding integrated diagnostics. Either we calculate these things uniformly across every test, or we need to design an appropriate Lua interface so that these kinds of options can be passed in. We can't have random bespoke code in a C test if the whole idea is to facilitate comparisons against Lua tests that do not contain such bespoke code.

JonathanGorard:
@kltrbrds Right - this is a good point which I should have clarified within the Design Review. The fields in the Lua-C API itself should always use camelcase (e.g. we use `tEnd` to designate the end time of the simulation). However, because I wanted to try to make the Lua input files match the C input files to the greatest extent possible, in order to facilitate direct comparisons, I used the underscore convention (e.g. `t_end`) when defining the global variables at the beginning of the input file. Hence why we have lots of lines like `tEnd = t_end`, etc.

This is perhaps a slightly silly way to do it (it made sense to me at the time...), so I'm happy to modify if other people think that this is unnecessarily confusing.

@akashukla That's simply a byproduct of the fact that there are three major things on the GK side that have not yet been fully wrapped and tested: mirror and tokamak geometry structs, reaction structs, and neutral species structs. Once this is done, all of the tests you mention will also have their own Lua-wrapped versions.

JunoRavin:
Such a phenomenal amount of work! 

Couple of quick things on the PKPM side that I just want us to document but do not affect my approval:
1. I very badly kept calling the velocity dimension in PKPM C tests "vx" and the number of velocity space cells "Nvx" or "Nv." These names should be changed to vpar and Nvpar to be consistent with GK (and since the PKPM velocity dimension is actually vpar, not vx). 
2. There are PKPM tests that define G, the second distribution function (based on the L1 Laguerre coefficient) as T*T*F0. I realize these tests all have T = m = 1.0. But we should always define in our input files vt = sqrt(T/m) and G = vt*vt*F0 (we do this in other tests where the units matter, which is good!).

JunoRavin (approved):

regression/lua/rt_gk_bgk_cross_relax_1x2v_p1.lua
```diff
@@ -0,0 +1,221 @@
...
+      end,
+
+      numCrossCollisions = 1,
+      collideWith = { "ion" },
+
+      correctAllMoments = true,
+      iterationEpsilon = 1.0e-12,
+      maxIterations = 10,
+      useLastConverged = true
+    },
```
manauref, line 152: This isn't for this PR, but (FYI @JunoRavin ) I think this set of inputs for correction operations should be grouped in a single struct/table. That way we can pass those (optional) inputs as a group to various places where we do corrections (e.g. initialization, BGK, reactions, etc)

manauref (changes requested):
Thank you for undertaking this. As I mentioned many months ago, bringing back automated testing is one of the top (maybe the top) most important thing for (at least GK) Gkeyll.
For the most part I'd like to vouch for restoring certain things that were changed in this PR:
1. GK reg tests had a way to do integrated diagnostics at a different rate. I think we should restore that.
2. There's at least one test (rt_gk_sheath_3x2v_p1) where we do phase-space diagnostics at a different rate too. If you are going to standardize things across the board, then I think that should be the standard. These things are important for production science and help us not fill our disk space.
3.  The logic surrounding integrated diagnostics was carefully chosen (by me). I'd like to double check: have you ensured that the same behavior is reproduced for single sims, restart sims, and crashing sims?
4. You brought back a bunch of lines (~80?) into GK regression tests that handle communicator creation. This was packaged neatly into the `gkyl_gyrokinetic_comms_new` function. I think we should bring that back. If you are worried about it not being standardized with other apps, then I think we should do similar things for other apps (e.g. I had created one for Vlasov and used in Vlasov Poission reg tests).
Other observations/questions:
a) If the main avenue for testing will be through Lua, then I think all the C tests in `regression/` should be moved to a `c/` folder, so that the organization doesn't give one the impression that the C files are what matters, and that the `lua/` folder is just some auxiliary thing.
b) There are no vlasov_poisson Lua tests. Why is that?
c) There are no gk_d3d_iwl Lua tests. You might want to check that.
d) Have you tried running single and/or multi GPU sims through Lua?

JonathanGorard:
@manauref Thanks. I've attempted to address points 1-4 in replies to your inline comments. Regarding your other observations:

(a) I agree - I think that we should do something like this as part of the planned directory reorganization, so that we are not (inadvertently) privileging one language's test suite over another.

(b) All of the Vlasov-Poisson C tests have counterparts in the Lua layer. See `rt_vp_landau_damping_1x1v_p1.lua`, `rt_vp_landau_damping_1x1v_p2.lua`, `rt_vp_sheath_1x1v_p2.lua`, `rt_vp_sheath_Aext_1x2v_p2.lua`, `rt_vp_sheath_Bext_1x2v_p2.lua`, `rt_vp_sheath_feedback_sources_1x1v_p2.lua` and `rt_vp_sheath_feedback_sources_bgk_1x1v_p2.lua`.

(c) I don't think there is anything to "check" here: these tests have simply not been moved to Lua yet.

(d) Not yet. Extensive parallel and GPU testing of the Lua layer is at the top of the agenda for the next merge.

regression/lua/rt_pkpm_em_advect_p1.lua
```diff
@@ -0,0 +1,131 @@
...
+      local Ez = 0.0 -- Total electric field (z-direction).
+
+      local Bx = 0.0 -- Total magnetic field (x-direction).
+      local By = 0.0 -- Total magnetic field (y-direction).
+      local Bz = 0.0 -- Total magnetic field (z-direction).
+
+      return Ex, Ey, Ez, Bx, By, Bz, 0.0, 0.0
+    end,
+
+    externalFieldInit = function (t, xn)
```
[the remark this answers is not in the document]
JonathanGorard: This is a good point - there is a wrapped "evolveExternalField" flag but I neglected to set it (or its C counterpart) for this test. Will fix. Thanks!

regression/rt_gk_sheath_1x2v_p1.c
JonathanGorard: Oh don't even get me started on our use of "vector" for things that are really covectors or pseudovectors. I've given @ammarhakim grief about this before ;)

I thought that it would be useful (as it is in things like 10-moments) to distinguish explicitly between, say, a parallel vs. perpendicular temperature and an isotropic/total temperature. If you say that this distinction is not useful or meaningful in GK, then I can't profess to understand, but I'm not going to contradict you.

manauref: in GK we simply say the temperature, and I would stick with that. Or the scalar temperature (instead of the tensor temperature).

The left/right above just doesn't make sense to me, sorry. Likewise, so far the distinction between vectors, pseudovectors, covectors, etc, doesn't arise in GK at the input file level, so I would not make the language in the input file more complicated than it needs to be, just because you are using other language with other models. They are different models, with different languages, and more importantly, different communities of people that speak differently (sometimes about the same thing for the bane of us all)

JonathanGorard: Already commented that the left/right thing was a vestigial remnant.

Just because you don't discuss these distinctions doesn't mean that they don't exist, or are not important. There are many places, for instance on the GK side, where a variable will be called something like `T_elc`, and it's not clear whether it's an isotropic temperature that should appear in (e.g.) the calculation of a thermal velocity, or a parallel temperature that should appear (e.g.) in the calculation of a velocity space extent. Very often it's used as both, which is clearly completely wrong.

We should be aiming to minimize these kinds of glaring conceptual errors, not promote them.

regression/rt_gk_sheath_1x2v_p1.c
manauref: But the logic in the GK reg tests here seems to have changed slightly compared to what I wrote (see for example rt_gk_sheath_2x2v_p1 in gk-g0-app), and my memory of it is that, after some trial and error, that exact logic is what was needed to produce the intended behavior in the 3 cases mentioned above for both grid and integrated diagnostics. Maybe I'm wrong, or it's possible that the new logic used here is equivalent, but I would just like to know if it is and/or if it behaves the same way.

Field energy diag: I would put it in every reg test. It's a useful tool (because it's more sensitive than the integrated moments). I thought that was the case already, and if it wasn't I'm sorry I didn't update that.

regression/rt_gk_sheath_1x2v_p1.c
manauref: I think we should simplify the Lua interface then. I don't think we should bring back these ~80 lines of code to every C reg test, and I don't think any Lua input file should have equivalently clunky code. I'm sure a simple gyrokinetic_comms_new function can be created in Lua too (and equivalently for other apps).

regression/rt_gk_sheath_1x2v_p1.c
manauref: FYI as a reminder, I think I saw this in a few tests, maybe both C and Lua, in case you forget

regression/rt_gk_sheath_1x2v_p1.c
manauref: this isn't bespoke code, it's important production-worthy code. If by bespoke you mean single-purpose, then the IWL feature in GK is bespoke, but that is the avenue for a whole research program (with clopen sims)!
Back to this point: if your motivation is to standardize across other models or across all GK tests, then I think the standard should be to allow for different diag rate for integrated, conf and phase space diagnostics, as I think is the case in rt_gk_sheath_3x2v_p1. That is truly the most valuable way to portray and use the GK solver

manauref:
> @manauref Thanks. I've attempted to address points 1-4 in replies to your inline comments. Regarding your other observations:
> 
> (a) I agree - I think that we should do something like this as part of the planned directory reorganization, so that we are not (inadvertently) privileging one language's test suite over another.
> 
> (b) All of the Vlasov-Poisson C tests have counterparts in the Lua layer. See `rt_vp_landau_damping_1x1v_p1.lua`, `rt_vp_landau_damping_1x1v_p2.lua`, `rt_vp_sheath_1x1v_p2.lua`, `rt_vp_sheath_Aext_1x2v_p2.lua`, `rt_vp_sheath_Bext_1x2v_p2.lua`, `rt_vp_sheath_feedback_sources_1x1v_p2.lua` and `rt_vp_sheath_feedback_sources_bgk_1x1v_p2.lua`.
> 
> (c) I don't think there is anything to "check" here: these tests have simply not been moved to Lua yet.
> 
> (d) Not yet. Extensive parallel and GPU testing of the Lua layer is at the top of the agenda for the next merge.

@JonathanGorard regarding (c)
I single out d3d_iwl because that uses additional BC interface that is worth checking (that it runs). But if you haven't gotten there and want to wait for the next PR that's fine.

manauref:
To a GK user it is clear that T_e is the scalar electron temperature.

We often use T_e to define parallel velocity grid extents, and that is
perfectly fine and in accordance with conventions in the GK field. So I
don't see a problem here.

I understand your point, but I'm arguing from the point of view of a GK
person, someone who did some GK in undergrad, gradschool or postdoc,
outside of Gkeyll. Sticking to that is preferable for these input file
things.

lua/Comm/gkyl_mpi_funcs.c
```diff
@@ -0,0 +1,120 @@
+// Gkyl ------------------------------------------------------------------------
```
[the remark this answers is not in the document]
ammarhakim: 1. We can't run Tools without MPI. 
2. No, unless we revive the cuda-info Tool or need to run a Tool on GPUs

lua/DataStruct/CartField.lua
```diff
@@ -0,0 +1,889 @@
...
+         self._decompNeigh:calcFaceCommNeigh(ghost[1], ghost[2])
+      end
+
+      local localExtRange = self._localExtRange
+      local indexer = self:genIndexer()
+      local decomposedRange = self._grid:decomposedRange()
+      local myId = self._grid:subGridId() -- Grid ID on this processor.
+      local neigIds = self._decompNeigh:neighborData(myId) -- List of neighbors.
+
+      -- Create buffers for periodic copy if Mpi.Comm_size(nodeComm) = 1.
```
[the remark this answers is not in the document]
ammarhakim: For now, yes

lua/Grid/RectCart.lua
```diff
@@ -0,0 +1,359 @@
...
+
+-- RectCart --------------------------------------------------------------------
+--
+-- A uniform Cartesian grid
+--------------------------------------------------------------------------------
+
+-- Uniform Cartesian grid
+local RectCart = Proto()
+
+function RectCart:init(tbl)
```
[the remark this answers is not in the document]
ammarhakim: This could be cleaned up later. For now this is needed to get to a working Tool system.

regression/lua/rt_gk_bgk_cross_relax_1x2v_p1.lua
```diff
@@ -0,0 +1,221 @@
...
+
+    evolve = true, -- Evolve species?
+    diagnostics = { "M0", "M1", "M2", "M2par", "M2perp", "BiMaxwellianMoments" }
+  },
+
+  skipField = true,
```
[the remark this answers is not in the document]
ammarhakim: Not really. We want to keep this.

PR #555, opened 2024-12-12
Gk g0 app host allgather

Maxwell-Rosen:
Added a host allgather for GPU simulations. This is needed to gather a global representation of Bmag for the non-uniform grids. I added 4 unit tests for this, 2 MPI, 2 NCCL. They all pass on Stellar in 2 and 4 CPUs, and on 2 GPUs. 

The style is copied from the differences between bcast and bcast_host. Almost everything is nearly identical, except for how the unit tests work.

### DIFF
--- a/zero/gkyl_comm_priv.h
+++ b/zero/gkyl_comm_priv.h
@@ -95,6 +95,7 @@ struct gkyl_comm_priv {
   // FOLLOWING NEED A DECOMP 
 
   gkyl_array_allgather_t gkyl_array_allgather; // gather local arrays to global array
+  gkyl_array_allgather_t gkyl_array_allgather_host; // ?? gather local arrays to global array on host
 
   gkyl_array_bcast_t gkyl_array_bcast; // broadcast array to other processes
   gkyl_array_bcast_t gkyl_array_bcast_host; // ?? broadcast host side array to other processes
--- a/unit/mctest_mpi_comm.c
+++ b/unit/mctest_mpi_comm.c
@@ -256,6 +256,137 @@ mpi_n4_allgather_2d()
   gkyl_array_release(arr_global);   
 }
 
+
+static void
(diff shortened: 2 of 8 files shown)


zero/gkyl_comm_priv.h
```diff
@@ -95,6 +95,7 @@ struct gkyl_comm_priv {
   // FOLLOWING NEED A DECOMP 
 
   gkyl_array_allgather_t gkyl_array_allgather; // gather local arrays to global array
+  gkyl_array_allgather_t gkyl_array_allgather_host; // ?? gather local arrays to global array on host
```
manauref, line 98: @Maxwell-Rosen what's up with the ??

unit/ctest_null_comm.c
```diff
@@ -79,9 +79,68 @@ test_1d()
     }
   }
 
+  ///////// Test gkyl_comm_array_allgather //////////////
```
manauref, line 82: Please reformat this comment to remove //////// since we don't do that elsewhere.

Also, instead of lengthening this test, can you please create new tests for these methods? you could potentially keep allgather/allgather_host in the same test (same with bcast).

manauref (changes requested):
Seems good. Just requested a couple of small changes (make sure the unit tests pass after such changes).

zero/gkyl_comm_priv.h
Maxwell-Rosen: Not sure. I didn't put them there. @ammarhakim put these ??

unit/ctest_null_comm.c
Maxwell-Rosen: I thought this was the right thing to do when writing the test, but this is how the test was organized to begin with. I'll reorganize this

manauref (approved):

JunoRavin (approved):

Issue #559, opened 2024-12-19
GK L2 norm diagnostic


manauref:
We wish to add an L2 norm diagnostic in the gyrokinetic app. The L2 norm is defined as

$$ ||f||_2 = \int \mathrm{d}^3\mathbf{R} \mathrm{d}^3\mathbf{v}~|f|^2 $$

where $R$ is the guiding center position and $v$ the velocity, or once you take into account all the coordinate transforms and Jacobians it is

$$ ||f||_2 = \frac{2\pi}{m}\int\mathrm{d}^3\mathbf{x}\mathrm{d}\eta\mathrm{d}\kappa~f F $$

where $(\mathbf{x},\eta,\kappa)$ are the computational coordinates along configuration, $v_\parallel$ and $\mu$ directions, respectively, and $F=J_xJ_v B f$. Here $J_x$ and $J_v$ are the Jacobians of the $R(x)$ and $(v_\parallel(\eta),\mu(\kappa))$ mappings, respectively.

For this diagnostic we will:
- Modify the array_integrate updater so it can compute the weighted $f^2$ integral.
- Compute the L2 norm as the weighted integral of $F^2$, with a $1/(J_xB)$ weight.

This diagnostic is prototyped in the gk-g0-app_l2mod_prototype branch.

There may be better ways to compute $||f||_2$, but for now this is an OK diagnostic.

manauref:
Addressed in PR #560

PR #560, opened 2024-12-19
L2 norm diagnostic for gyrokinetics

manauref:
This adds the option to calculate and write an (integrated) diagnostic providing the L2 norm of the distribution vs. time. See the DR in issue #559 .

This doesn't change any core code, it only adds a weighted f^2 integral to the array_integrate updater, and (optional) diagnostic functions to the GK app. We did reorganize the array_integrate updater so that it is simpler and easier to extend.

This goes with PR 52 in gkylcas (https://github.com/ammarhakim/gkylcas/pull/52).


tnbernard (approved):

zero/gkyl_array_integrate_priv.h
```diff
@@ -59,18 +89,30 @@ struct gkyl_array_integrate {
...
-void gkyl_array_integrate_choose_kernel(enum gkyl_array_integrate_op op, const struct gkyl_basis *basis, struct gkyl_array_integrate *up)
+void gkyl_array_integrate_choose_kernel(enum gkyl_array_integrate_op op,
+  const struct gkyl_basis *basis, struct gkyl_array_integrate *up)
 {
   int ndim = basis->ndim, poly_order = basis->poly_order;
 
-  if (op == GKYL_ARRAY_INTEGRATE_OP_GRAD_SQ) {
+  if (op == GKYL_ARRAY_INTEGRATE_OP_NONE) {
+    up->kernel = gkyl_array_integrate_none_ker_list_ser[ndim-1].kernels[poly_order-1];
+  } else if (op == GKYL_ARRAY_INTEGRATE_OP_ABS) {
```
JunoRavin, line 99: We should move these else if and else statements onto new lines to be consistent with the style elsewhere.

zero/array_integrate_cu.cu
```diff
@@ -15,14 +15,25 @@ gkyl_array_integrate_set_ker_cu(struct gkyl_array_integrate *up, enum gkyl_array
 {
   int ndim = basis.ndim, poly_order = basis.poly_order;
 
-  if (op == GKYL_ARRAY_INTEGRATE_OP_GRAD_SQ) {
+  if (op == GKYL_ARRAY_INTEGRATE_OP_NONE) {
+    up->kernel = gkyl_array_integrate_none_ker_list_ser[ndim-1].kernels[poly_order-1];
+  } else if (op == GKYL_ARRAY_INTEGRATE_OP_ABS) {
```
JunoRavin, line 20: Same thing here is the new lines for the else if and else.

zero/array_integrate.c
```diff
@@ -51,15 +34,18 @@ gkyl_array_integrate_new(const struct gkyl_rect_grid *grid, const struct gkyl_ba
...
 {
 #ifdef GKYL_HAVE_CUDA
   if (up->use_gpu) {
-    gkyl_array_integrate_advance_cu(up, fin, factor, weight, range, out);
+    gkyl_array_integrate_advance_cu(up, fin, factor, weight, range, weight_range, out);
     return;
   }
 #endif
 
+  int widx[GKYL_MAX_DIM] = {-1};
```
JunoRavin, line 47: Setting widx[GKYL_MAX_DIM] = {-1} doesn't initialize the entries to -1. Is there a reason you aren't setting this to 0? If you need it to be -1 it has to be a loop.

manauref: I've taken to initializing multi-dim indices to -1 instead of 0 because they fail when 0 doesn't (because 0 is a ghost cell), helping mind things that are wrong more easily.

zero/gkyl_array_integrate_priv.h
manauref: done

JunoRavin (approved):

Issue #546, opened 2024-11-29
[DR] Non-uniform grids in the gyrokinetic app


Maxwell-Rosen:
Associated PR #547 This DR encompasses the need for non-uniform geometries in the numerical and analytic geometries that we deal with in Gkeyll. This work stems from a desire to pack cells near the mirror throat for magnetic mirror / WHAM simulations. @akashukla implemented the general geometry, but it assigned the grid a uniform spacing along a field line.

Official notes: https://www.overleaf.com/read/bmpfcgwtrjwj#53a369

Two methods to deform the grid are implemented. A user can specify a mapping from the input file to deform the grid, or there can be a non-uniform mapping dependent on the global magnetic field. The magnetic field-dependent mapping involves initializing a uniform geometry magnetic field and then reading it to determine the appropriate deformation. There is a numeric mapping that minimizes variation of B along a cell, and a reduced order polynomial-based map is used to accomplish this same task. This mapping may also have parameters set by the user. Since geometry is defined in 3x and then deflated for lower cdim simulations, this non-uniform mapping must be defined in 3x. 

I introduced a new structure to the zero folder called `position_map`. This contains information about the non-uniform mapping and information from within the geometry.

A separate position_map_set function is created to mirror the velocity_map interface, where the initialization function does not take in an array input. Instead, we need to read information from the geometry because this map is set in the geometry.

A new array is added to GK geometry called mc2nu. This object mirrors mapc2p, except instead of populating it with cylindrical coordinates, mc2nu is populated with the field-aligned coordinates of every point. This is necessary for the position_map object. Each point in computational space always maps to 3 coordinates in non-uniform space (psi, alpha, theta) to locate the point.

The initial conditions must be mapped to the non-uniform geometry. A method `proj_on_basis_c2p_position_func` is introduced to the `gk_species_projection.c` files to deform a grid using the position map. This initial condition mapping is why it's so important to store the c2fa mapping.

Initial conditions are appropriately mapped with the non-uniform deformation.

In an input file, users may specify using the non-uniform map by specifying this structure.

```
struct gkyl_gk app_inp = {
    .geometry = {
      .position_map_info = {
        .id = GKYL_PMAP_CONSTANT_DB_NUMERIC,
        .map_strength = 0.9999,
      },
    },
}
```

[...]

PR #561, opened 2024-12-20
Computational to non-uniform mapping for gyrokinetic simulations

Maxwell-Rosen:
Associated DR #546. This has included 12 regression tests in mapc2p, tokamak and mirror geometry. There is a unit test for the zero/position_map.c functionality. Tests have been successfully completed on CPU, GPU, multi-CPU, and multi-GPU. The tests are valgrind clean, although the STEP unit tests are not valgrind clean on gk-g0-app.

Notes are contained at this overleaf link.
https://www.overleaf.com/read/bmpfcgwtrjwj#53a369


zero/position_map.c
```diff
@@ -0,0 +1,227 @@
...
+
+  struct gkyl_position_map *gpm_out = gpm;
+  return gpm_out;
+}
+
+void
+gkyl_position_map_set(struct gkyl_position_map* gpm, struct gkyl_array* mc2nu)
+{
+  // Should be a copy, but there are issues I will look into later
+  gkyl_array_release(gpm->mc2nu);
```
Maxwell-Rosen, line 131: Use copy instead of acquire

Maxwell-Rosen:
Test the root finding map more. Test 2x and 3x wham. Make plots of the magnetic field. Check with a mirror not located at pi/2.

Maxwell-Rosen:
Here is an example of the non-uniform map working on the STEP geometry to minimize the magnetic field variation.

image
image

Maxwell-Rosen:
Here is a non-uniform map for WHAM that makes dB/dTheta = constant

image
image

Maxwell-Rosen:
[Non_uniform_mirror_grids (6).pdf](https://github.com/user-attachments/files/18324165/Non_uniform_mirror_grids.6.pdf)

zero/gkyl_position_map.h
```diff
@@ -0,0 +1,131 @@
...
+};
+
+struct gkyl_position_map_inp {
+  enum gkyl_position_map_id id;
+  mc2nu_t map_x; // Map in the first direction (x, psi, r, etc.)
+  mc2nu_t map_y; // Map in the second direction (y, alpha, theta, etc.)
+  mc2nu_t map_z; // Map in the third direction (z, field line length, etc.)
+  void *ctx_x;  // Context for map_x.
+  void *ctx_y;  // Context for map_y.
+  void *ctx_z;  // Context for map_z.
```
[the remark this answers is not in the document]
Maxwell-Rosen: So in code, this would look something like this
```
mc2nu_t *maps[3]; 
void *ctx[3];
```
A user would set
```
.geometry = {
      .position_map_info = {
        .id = GKYL_PMAP_FUNC,
        .maps = { NULL, NULL, nonuniform_position_map_z},
        .ctx =     { NULL, NULL, ctx},
      },
}
```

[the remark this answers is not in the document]
manauref: That's right.

ammarhakim (changes requested):
Please do not use the keyword NULL. Use 0 instead. Also, use array of function pointers in the proper way.

JunoRavin:
Summarizing an offline conversation between @Maxwell-Rosen and I:

General feedback: since we always order the coordinates as (position space, velocity space) (position space is first in the coordinate ordering) everywhere you need both the configuration space and velocity space mappings, or include both headers, or specify both structs, let’s order things such that the position space map is passed/acted on first. So for example in the phase mapping function in gk_species_projection.c, evaluate the position_map function first, and then the velocity_map function. This way, we can keep our natural coordinate orderings everywhere (and no future developer wonders if there’s a reason the velocity map is being acted upon first).

Maxwell-Rosen:
Summarizing another conversation with Jimmy offline, I have been allocating and freeing the global bmag inside the geometry generation, which seems out of place. I implemented it this way because I had been using whether the global bag inside position_map was allocated to determine whether we should do the geometry-dependent optimization inside position_map. Instead, these objects should be allocated inside the position map, and a flag can be set whether we should optimize.

Jimmy also notes that this operation is repeated across many different geometry types, and code is being copied. The geometry generation in gyrokinetic.c could be packaged and moved somewhere else to make it more compact and readable. This is left to future work but is worth revisiting.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+
+void
+calculate_mirror_throat_location_polynomial(struct gkyl_position_map_const_B_ctx *constB_ctx, struct gkyl_bmag_ctx *bmag_ctx)
+{
+  // Assumes symmetry along theta centered at 0 and two local maxima in Bmag that are symmetric
+
+  enum { X_IDX, Y_IDX, Z_IDX }; // arrangement of cartesian coordinates
+  double psi = constB_ctx->psi;
+  double alpha = constB_ctx->alpha;
+  double *xp = malloc(3*sizeof(double));
```
manauref, line 40: here and line 49 below, please don't use malloc for small arrays whose size is known at compile time. Just use double xp[3].

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+
+void
+position_map_constB_z_polynomial(double t, const double *xn, double *fout, void *ctx)
+{
+  // Converts our uniform coordinate along field line length to a non-uniform coordinate
+  // according to the polynomial mapping of arbitrary order.
+  // Notation: I switch from theta to z here. Both are computational coordinate.
+  struct gkyl_position_map_const_B_ctx *app = ctx;
+  int n_ex = app->map_order_expander;//app->mapping_order_expander;
+  int n_ct = app->map_order_center;//app->mapping_order_center;
```
manauref, line 84: there's some commented code here. Remove if not needed.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+  free(xp);
+  free(fout);
+}
+
+void
+position_map_constB_z_polynomial(double t, const double *xn, double *fout, void *ctx)
+{
+  // Converts our uniform coordinate along field line length to a non-uniform coordinate
+  // according to the polynomial mapping of arbitrary order.
+  // Notation: I switch from theta to z here. Both are computational coordinate.
```
manauref, line 81: The reader doesn't know who "I" is. Either speak in 3rd person or initialize and date your comment, e.g. MR 2024/12/25.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+void
+calculate_optimal_mapping_polynomial(struct gkyl_position_map_const_B_ctx *constB_ctx, struct gkyl_bmag_ctx *bmag_ctx)
+{
+  // Could be refined further by doing midpoint root finding, like the mirror throat finding does
+  // Expander region
+  enum { X_IDX, Y_IDX, Z_IDX }; // arrangement of cartesian coordinates
+  double psi = constB_ctx->psi;
+  double alpha = constB_ctx->alpha;
+  double *xp = malloc(3*sizeof(double));
+  double *fout = malloc(3*sizeof(double));
```
manauref, line 136: don't use mallocs, see earlier comment

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+  else
+  {
+    nonuniform_coordinate = uniform_coordinate;
+  }
+  fout[0] = nonuniform_coordinate;
+}
+
+void
+calculate_optimal_mapping_polynomial(struct gkyl_position_map_const_B_ctx *constB_ctx, struct gkyl_bmag_ctx *bmag_ctx)
+{
```
manauref, line 129: add comment describing what this function does.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+ * object when the number of references has decreased to zero.
+ *
+ * @param ref Reference counter for this object.
+ */
+void gkyl_position_map_free(const struct gkyl_ref_count *ref);
+
+// Utility functions for polynomial based B mapping
+
+void
+calculate_mirror_throat_location_polynomial(struct gkyl_position_map_const_B_ctx *constB_ctx, struct gkyl_bmag_ctx *bmag_ctx)
```
manauref, line 33: add comment describing what this function does.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+
+// Utility functions for numeric root finding B mapping
+
+
+double
+calc_bmag_global_derivative(double xn, void *ctx)
+{
+  struct gkyl_position_map *gpm = ctx;
+  struct gkyl_bmag_ctx *bmag_ctx = gpm->bmag_ctx;
+  double h = 1e-6;
```
manauref, line 256: why 1e-6? should this be multiplied by something else? what if my field line is 2e-6 long?

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+  free(xp);
+  free(fout);
+}
+
+
+// Utility functions for numeric root finding B mapping
+
+
+double
+calc_bmag_global_derivative(double xn, void *ctx)
```
manauref, line 252: add comment describing what this function does.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+  double Bmag_plus = fout[0];
+  xh[2] = xn - 2*h;
+  gkyl_calc_bmag_global(0.0, xh, fout, bmag_ctx);
+  double Bmag_minus = fout[0];
+  return (Bmag_plus - Bmag_minus) / (h);
+}
+
+void
+find_B_field_extrema(struct gkyl_position_map *gpm)
+{
```
manauref, line 272: add comment describing what this function does.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+  }
+  int num_boundaries = gpm->constB_ctx->N_theta_boundaries;
+  double dB_cell = B_total_change / (num_boundaries-1);
+
+  gpm->constB_ctx->dB_cell = dB_cell;
+
+}
+
+double
+position_map_numeric_optimization_function(double theta, void *ctx)
```
manauref, line 358: add comment describing what this function does.

zero/gkyl_position_map_priv.h
```diff
@@ -0,0 +1,448 @@
...
+  double dB_target = ridders_ctx->dB_target;
+  double dB_global_lower = ridders_ctx->dB_global_lower;
+  double B_lower_region = ridders_ctx->B_lower_region;
+
+  double result = fabs(Bmag - B_lower_region) + dB_global_lower - dB_target;
+  return result;
+}
+
+void
+position_map_constB_z_numeric(double t, const double *xn, double *fout, void *ctx)
```
manauref, line 382: add comment describing what this function does.

zero/gkyl_calc_bmag.h
```diff
@@ -8,7 +8,19 @@
 
 // Object type
 typedef struct gkyl_calc_bmag gkyl_calc_bmag;
-typedef struct bmag_ctx bmag_ctx;
+
+struct gkyl_bmag_ctx{
```
manauref, line 12: Add a comment describing that this struct is for, and each of its contents.

zero/gkyl_calc_bmag.h
```diff
@@ -8,7 +8,19 @@
...
+   const struct gkyl_rect_grid* grid;
+   const struct gkyl_rect_grid* cgrid;
+   const struct gkyl_range* range;
+   const struct gkyl_range* crange;
+   const struct gkyl_range* crange_global;
+   const struct gkyl_basis* basis;
+   const struct gkyl_basis* cbasis;
+   const struct gkyl_array* bmagdg;
+   const struct gkyl_array* bmag;
```
manauref, line 21: what's the difference between these two?

zero/gkyl_gk_geometry.h
```diff
@@ -26,6 +27,7 @@ struct gk_geometry {
   // GK Equation and Poisson Equation and to apply certain BC's
   // The first 20 are defined on the configuration space domain. The last is a single element.
   struct gkyl_array* mc2p; // 3 components. Cartesian X,Y, and Z
+  struct gkyl_array* mc2nu_pos; // 3 components. Computational to field aligned mapping
```
manauref, line 30: shouldn't this comment be "computational to nonuniform" ?

zero/gkyl_gk_geometry.h
```diff
@@ -76,6 +78,8 @@ struct gkyl_gk_geometry_inp {
   struct gkyl_efit_inp efit_info; // context with RZ data such as efit file for a tokamak or mirror
   struct gkyl_tok_geo_grid_inp tok_grid_info; // context for tokamak geometry with computational domain info
   struct gkyl_mirror_geo_grid_inp mirror_grid_info; // context for mirror geometry with computational domain info
+  struct gkyl_position_map *position_map; // position map object
+  struct gkyl_comm *comm; // communicator object
```
manauref, line 82: I don't think this is right. Instead of storing a pointer to comm, why can't you pass comm to the respective functions as an argument? we don't store comm anywhere (in zero), and parallel operations are technically always orchestrated at the app level, not at this level.

zero/gkyl_gk_geometry_mirror.h
```diff
@@ -3,9 +3,8 @@
 #include <gkyl_gk_geometry.h>
 /**
  * Create a new geometry object using mirror input (efit)
+ * Wrapper for gkyl_gk_geometry_mirror_advance, neccisary for the non-uniform grids
```
manauref, line 6: This added comment seems misleading. Isn't this used/needed if we are not using nonuniform grids? To call it a wrapper for nonuniform grids seems wrong.

zero/gkyl_mirror_geo.h
```diff
@@ -166,8 +167,9 @@ double gkyl_mirror_geo_integrate_psi_contour(const struct gkyl_mirror_geo *geo,
  * @param ddtheta_nodal output nodal field containing dphi/dtheta = s(psi)/R|grad(psi)|, dR/dtheta and dZ/dtheta
  */
 void gkyl_mirror_geo_calc(struct gk_geometry* up, struct gkyl_range *nrange, double dzc[3], 
-  struct gkyl_mirror_geo *geo, struct gkyl_mirror_geo_grid_inp *inp, 
-  struct gkyl_array *mc2p_nodal_fd, struct gkyl_array *mc2p_nodal, struct gkyl_array *mc2p, struct gkyl_array *ddtheta_nodal);
+  struct gkyl_mirror_geo *geo, struct gkyl_mirror_geo_grid_inp *inp,
```
manauref, line 170: here and elsewhere: update comment with description of new arguments.

zero/gkyl_tok_geo.h
```diff
@@ -218,8 +219,13 @@ void gkyl_tok_geo_mapc2p(const struct gkyl_tok_geo *geo, const struct gkyl_tok_g
  * @param mc2p_nodal output nodal mapc2p field R,Z,phi)
  * @param mc2p On output, the DG representation of mapc2p ((R,Z,phi)
  * @param ddtheta_nodal output nodal field containing dphi/dtheta = s(psi)/R|grad(psi)|, dR/dtheta and dZ/dtheta
+ * @param position_map position map object
```
manauref, line 222: update comment with new arguments.

manauref (changes requested):

zero/gkyl_calc_bmag.h
Maxwell-Rosen: @akashukla bmagdg is the RZ representation of bmag from the efit file and bmag is in computational coordinates. They are both DG arrays. I find this naming also bad. It should be bmag_rz or something.

zero/gkyl_gk_geometry.h
Maxwell-Rosen: It's because this struct `gkyl_gk_geometry_inp` is the input to all the `gkyl_geometry_new` functions in `gyrokinetic.c`. The only input is this struct. Rather than making many arguments, I pass it this way. It makes the input more flexible also.

This comm is used for the bmag allgather to make bmag global. This operation happens at the zero level because it's in geometry.

zero/gkyl_mirror_geo.h
Maxwell-Rosen: This must have gotten screwed up by the recent merge

zero/gkyl_gk_geometry.h
JunoRavin: We do sometimes store comm in zero-level objects depending on what parallel operations need to be done at the zero level, such as here in wave_prop.c because it's a single step method that must do the allreduce internally for what the stable time step is: https://github.com/ammarhakim/gkylzero/blob/gk-g0-app/zero/wave_prop.c#L29

So in principal there is no issue with the comm being an internal object of a zero-level struct. So long as things are still done with acquires and releases so the memory is tracked carefully.

JunoRavin:
@Maxwell-Rosen In looking over your responses to @manauref's comments I realized something else: please give all the small functions defined in the private header the static keyword. Private headers (usually) need to have static functions in case the private header is included multiple places so that the compiler doesn't think there are duplicate symbols (this issue shows up a lot on GPUs when both a .c and .cu file need to include the private header; although there is no position_map_cu.cu function, I think we should institute a discipline that these sorts of private header functions are static).

zero/position_map.c
```diff
@@ -0,0 +1,189 @@
...
+  gpm->bmag_ctx->bmag = gkyl_array_new(GKYL_DOUBLE, basis.num_basis, global_ext.volume);
+  gpm->to_optimize = false;
+
+  gpm->constB_ctx = gkyl_malloc(sizeof(struct gkyl_position_map_const_B_ctx));
+
+  for (int i = 0; i < 3; i++){
+    gpm->maps[i] = gkyl_position_map_identity;
+    gpm->ctxs[i] = 0;
+    gpm->constB_ctx->maps_backup[i] = gkyl_position_map_identity;
+    gpm->constB_ctx->ctxs_backup[i] = 0;
```
manauref, line 41: what are these backups used for? can they be removed? if they are needed in optimization of the map, can't these pointers just be defined locally/temporarily in the optimization funciton?

zero/position_map.c
```diff
@@ -0,0 +1,189 @@
...
+      for (int i = 0; i < 3; i++){
+        if (pmap_info.maps[i] != 0)
+        { gpm->maps[i] = pmap_info.maps[i];
+          gpm->ctxs[i] = pmap_info.ctxs[i];
+        }
+      }
+
+    case GKYL_PMAP_CONSTANT_DB_POLYNOMIAL:
+
+      for (int i = 0; i < 2; i++){
```
manauref, line 56: Why is there a 2 here (and in line 66) but there's a 3 in line 47?

zero/gkyl_position_map.h
```diff
@@ -0,0 +1,130 @@
...
+  GKYL_PMAP_USER_INPUT = 0, // Function projection. User specified. Default
+  GKYL_PMAP_CONSTANT_DB_POLYNOMIAL, // Makes a uniform dB in each cell. Polynomial approximation, assuming 2 local maxima in Bmag
+  GKYL_PMAP_CONSTANT_DB_NUMERIC, // Makes a uniform dB in each cell, but calculates the dB numerically
+};
+
+typedef void (*mc2nu_t)(double t, const double *GKYL_RESTRICT xn, double *GKYL_RESTRICT fout, void *ctx);
+
+struct gkyl_position_map_inp {
+  enum gkyl_position_map_id id;
+  mc2nu_t maps[3]; // Position mapping in each position direction.
```
manauref, line 20: here or somewhere else you must make it clear to the user that even if it's a 1x or 2x sim they must provide a map in all 3 computational coordinate directions, unless a uniform map is desired. Otherwise some people might be confused to see .map[2] = ... in a 1x input file, or they may incorrectly set .map[0] in a 1x input file. This was a common error in the early days of geometry, so let's avoid that this time.

manauref (changes requested):
This is very good work Max. Thank you for undertaking this.

I added relatively small comments throughout. I only have 2 more requests before merging:
1. Check that 6 regression tests are valgrind clean, a tokamak one, a mapc2p one, and a mirror one, each with uniform and nonuniform grid. Running them for 3 time steps is sufficient.
2. Work with @pcagas and others to make sure we can plot data from a sim with nonuniform configuration space, and show the corresponding pgkyl command here (like we did in PR #359 ). Otherwise we are introducing huge changes that we cannot use effectively (i.e. to obtain AND convey better results).

I'll do a couple more checks myself too.

zero/position_map.c
Maxwell-Rosen: The backups are used for if you want to use a user-defined position map in x and/or y, but a magnetic field dependent map in z. We need a backup because we need to generate the geometry twice. The first time needs uniform everything, but the second time we need non-uniform everything.

zero/position_map.c
Maxwell-Rosen: Only the 0 and 1 indices are used for the backup because the 2nd index is used with the magnetic field dependent map, so only those 2 parts need to be assigned.

zero/gkyl_position_map.h
Maxwell-Rosen: In my regression tests, I specify .map[2] = function. This works just fine in my implementation because they are all initialized to identity maps and only assigned if the user gives a function. I should add a comment that they have to specify the position in 3x rather than the deflated geometry

ammarhakim (approved):

JunoRavin (approved):

manauref:
I believe this PR changed the evolution of the field in rt_gk_step_out_2x2v_p1, see below:
image

Issue #450, opened 2024-08-21
[DR] Simplification of how static GK species are handled


tnbernard:
We propose the following to handle static gk and gk_neut species (idea courtesy of @JunoRavin): 

Right now, we are having to introduce if statements in a bunch of places and some of these places are not necessarily sequential (don’t allocate certain internal arrays, don’t allocate certain updaters, etc)
What we should do:

1. In gk_species.c and gk_neut_species.c make new functions:  
    a) `gk_species_static_new` and `gk_neut_species_static_new`
    b) `gk_species_static_rhs` and `gk_neut_species_static_rhs`
2. Check in the init method if a species is static, if it is, call the static new method and set a function pointer to the correct rhs method

This has a couple of advantages:
1. No more if statements
2. Allows us to use pointers to just set things like f1 = f and fnew = f (so we don’t have to do a bunch of complex logic for whether f1 and fnew are allocated, we can just keep using f)

The DR will be updated with a link to a branch once development on this task has begun.

manauref:
One potential disadvantage of this approach is that there may be a lot of code duplication between `gk_species_static_new` and `gk_species_new`. But perhaps the common blocks can be canned and called equally by both functions.

JunoRavin:
DR implemented in https://github.com/ammarhakim/gkylzero/pull/583

PR #567, opened 2025-01-13
static species functionality to gk-g0-app branch

tnbernard:
This is related to DR: https://github.com/ammarhakim/gkylzero/issues/450

This restructures how the static functionality for gk_neutral species is applied by creating new static_init and static_rhs methods  which are called. @manauref, some code duplication was inevitable by restructuring in this way, and you can comment as you see fit. 

Static functionality for gk_species is also included, with an example of how to apply it in rt_gk_static_3x2v_p1. Be sure to include `.skip_field = true,` in the app definition. A sample of affected tests were run with Valgrind and are clean: rt_gk_static_3x2v_p1, rt_gk_step_out_2x2v_pt, and rt_gk_rad_low_Te_1x2v_p1.


JunoRavin:
I really like this dichotomy between the static initialization (and setting a specific RHS pointer that is then generically called). I think we should do a few things:
1. We should utilize this same logic for the integrated moments so we can eliminate the cumbersome nested if statements in gyrokinetic.c for whether to compute an integrated diagnostic only if the species is not static. Let's make an integrated diagnostic function pointer too (@manauref already did this in Vlasov for the field energy since the field energy is different between Vlasov Poisson and Vlasov Maxwell). 
2. Let's get rid of everything that doesn't need to be allocated in a static species. It looks like things like equation objects and fields are being created even though I imagine we do not need to call these things. If we need to, I would suggest in the same way we have a static_init we should have a static_release that releases the smaller big of allocated memory.

apps/gk_neut_species.c
```diff
@@ -16,7 +16,7 @@
 void
 gk_neut_species_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app, struct gk_neut_species *s)
 {
-  int cdim = app->cdim, vdim = app->vdim+1; // neutral species are 3v
+    int cdim = app->cdim, vdim = app->vdim+1; // neutral species are 3v
```
manauref, line 19: new indent

tnbernard:
@JunoRavin, the equation object needs to be initialized for static GK in order to calculate the boundary flux that is needed for the recycling BCs. There is likely a more efficient way to do this so that it's only initialized when recycling bcs are present.

apps/gk_neut_species.c
```diff
@@ -288,6 +287,105 @@ gk_neut_species_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app, struc
   }
 }
 
+void
+gk_neut_species_static_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app, struct gk_neut_species *s)
```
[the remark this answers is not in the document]
tnbernard: Fixed in [54e6fc0](https://github.com/ammarhakim/gkylzero/pull/567/commits/54e6fc0d7589e76cd30a6662b7a3e61df38e518e)

apps/gyrokinetic.c
```diff
@@ -514,7 +518,9 @@ gyrokinetic_calc_field_and_apply_bc(gkyl_gyrokinetic_app* app, double tcurr,
 
   // Apply boundary conditions.
   for (int i=0; i<app->num_species; ++i) {
-    gk_species_apply_bc(app, &app->species[i], distf[i]);
+    if (!app->species[i].info.is_static) {
```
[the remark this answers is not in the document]
tnbernard: Addressed for gk_neut_species in [54e6fc0](https://github.com/ammarhakim/gkylzero/pull/567/commits/54e6fc0d7589e76cd30a6662b7a3e61df38e518e)

apps/gyrokinetic.c
```diff
@@ -2224,7 +2236,7 @@ gyrokinetic_rhs(gkyl_gyrokinetic_app* app, double tcurr, double dt,
   // Compute RHS of Gyrokinetic equation.
   for (int i=0; i<app->num_species; ++i) {
     struct gk_species *s = &app->species[i];
-    double dt1 = gk_species_rhs(app, s, fin[i], fout[i]);
+    double dt1 = app->species[i].rhs_func(app, s, fin[i], fout[i]);
```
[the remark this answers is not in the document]
tnbernard: See [54e6fc0](https://github.com/ammarhakim/gkylzero/pull/567/commits/54e6fc0d7589e76cd30a6662b7a3e61df38e518e) for how this was changed for gk_neut_species

apps/gk_neut_species.c
tnbernard: Fixed in [54e6fc0](https://github.com/ammarhakim/gkylzero/pull/567/commits/54e6fc0d7589e76cd30a6662b7a3e61df38e518e)

apps/gk_neut_species.c
```diff
@@ -401,8 +328,16 @@ double
...
 {
-  double omega_cfl = 1/DBL_MAX;
+  double omega_cfl = species->cfl_func(app, species, fin, rhs);
+  return app->cfl/omega_cfl;
+}
 
+// Compute the RHS for species update, returning maximum stable
+// time-step.
+double
+gk_neut_species_omega_cfl(gkyl_gyrokinetic_app *app, struct gk_neut_species *species,
```
JunoRavin, line 338: This naming convention seems confusing to me that we perform the RHS update in a function labeled omega_cfl? Why don't we keep the original rhs naming convention and just the have the static RHS method return DBL_MAX with a comment like //Nothing to do; species static.

apps/gk_neut_species.c
```diff
@@ -562,15 +523,11 @@ gk_neut_species_release(const gkyl_gyrokinetic_app* app, const struct gk_neut_sp
...
-    gkyl_free(s->red_integ_diag);
-    gkyl_free(s->red_integ_diag_global);
-  }
+}
+
+// release resources for species
+void
+gk_neut_species_static_release_func(const gkyl_gyrokinetic_app* app, const struct gk_neut_species *s)
+{
+  // do nothing
```
JunoRavin, line 532: I don't think this is right. The static species method still has memory to be released like the f array and equation object correct?

tnbernard: It does. It gets released in gk_neut_species_release.

apps/gyrokinetic.c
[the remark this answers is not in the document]
tnbernard: After discussions with @JunoRavin, we decided to keep this structure to avoid additional indirectives in the species app and to be consistent with what is being done elsewhere in the vlasov app. The formatting was cleaned up some in gyrokinetic.c. See [73a3f7b](https://github.com/ammarhakim/gkylzero/pull/567/commits/73a3f7bfe6786c41244e800b065198fcc93ee9ca)

apps/gk_neut_species.c
tnbernard: Fixed in [73a3f7b](https://github.com/ammarhakim/gkylzero/pull/567/commits/73a3f7bfe6786c41244e800b065198fcc93ee9ca).

apps/gk_neut_species.c
tnbernard: Resolved in [73a3f7b](https://github.com/ammarhakim/gkylzero/pull/567/commits/73a3f7bfe6786c41244e800b065198fcc93ee9ca)

apps/gyrokinetic.c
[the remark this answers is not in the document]
tnbernard: Recent commits put these function pointers back at the species level.

apps/gk_species.c
```diff
@@ -750,6 +945,18 @@ gk_species_rhs(gkyl_gyrokinetic_app *app, struct gk_species *species,
   return app->cfl/omega_cfl_ho[0];
 }
 
+double
+gk_species_static_rhs(gkyl_gyrokinetic_app *app, struct gk_species *species,
+  const struct gkyl_array *fin, struct gkyl_array *rhs)
+{
+  gkyl_array_set(species->phi, 1.0, app->field->phi_smooth);
```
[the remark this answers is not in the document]
tnbernard: This was a mistake and is now removed.

apps/gyrokinetic.c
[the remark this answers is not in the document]
tnbernard: Addressed for gk_species in [ab8a838](https://github.com/ammarhakim/gkylzero/pull/567/commits/ab8a838d811ad20a031ce55c5d8353319870783c)

apps/gk_species.c
```diff
@@ -641,6 +644,193 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
   }
 }
 
+// initialize static species object
+void
```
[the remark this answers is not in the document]
tnbernard: Fixed with [ee38339](https://github.com/ammarhakim/gkylzero/pull/567/commits/ee38339113abf3f03c5dc8a307d8c97a2fb3cc37)

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1561,14 +1682,56 @@ double gk_neut_species_rhs(gkyl_gyrokinetic_app *app, struct gk_neut_species *sp
...
+ * @param species Pointer to neutral species
+ * @param fin Input distribution function
+ * @param rhs On output, the RHS from the neutral species object (df/dt)
+ * @return Maximum stable time-step
+ */
+double gk_neut_species_dynamic_rhs(gkyl_gyrokinetic_app *app, struct gk_neut_species *species,
+  const struct gkyl_array *fin, struct gkyl_array *rhs);
+
+/**
+ * Pass constant for max dt when neutral species is static
```
JunoRavin, line 1697: I think we can make this comment more descriptive. Something like:

"Compute RHS in the case when species is static. Only returns max double for time step estimate." 

Or something like that.

apps/gyrokinetic.c
```diff
@@ -403,12 +403,20 @@ gkyl_gyrokinetic_app_new_solver(struct gkyl_gk *gk, gkyl_gyrokinetic_app *app)
   }
 
   // initialize each species
-  for (int i=0; i<ns; ++i) 
-    gk_species_init(gk, app, &app->species[i]);
+  for (int i=0; i<ns; ++i)
```
JunoRavin, line 406: Since these are nested conditionals with a for loop, even though you don't need the braces, for clarity let's use braces everywhere like:

for {
  if {
  }
  else {
  }
}

apps/gk_species.c
```diff
@@ -302,6 +291,125 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
...
+  int cdim = app->cdim, vdim = app->vdim;
+  int pdim = cdim+vdim;
+
+  int cells[GKYL_MAX_DIM], ghost[GKYL_MAX_DIM];
+  double lower[GKYL_MAX_DIM], upper[GKYL_MAX_DIM];
+
+  int cells_vel[GKYL_MAX_DIM], ghost_vel[GKYL_MAX_DIM];
+  double lower_vel[GKYL_MAX_DIM], upper_vel[GKYL_MAX_DIM];
+
+  for (int d=0; d<cdim; ++d) {
```
JunoRavin, line 378: Do the species and neut_species need to reset things like cells, lower, etc.? Since the grid is already setup by common this seems unnecessary. Maybe you need to set the ghost array for use in the boundary condition updaters, but I think that is it.

apps/gyrokinetic_update_ssp_rk3.c
```diff
@@ -166,10 +158,12 @@ gyrokinetic_update_ssp_rk3(gkyl_gyrokinetic_app* app, double dt0)
         }
         else {
           for (int i=0; i<app->num_species; ++i) {
-            array_combine(app->species[i].f1,
-              1.0/3.0, app->species[i].f, 2.0/3.0, app->species[i].fnew, &app->species[i].local_ext);
-            gkyl_array_copy_range(app->species[i].f, app->species[i].f1, &app->species[i].local_ext);
-          }
+	    if (!app->species[i].info.is_static) {
```
JunoRavin, line 161: You don't have to go into info anymore right? You added an is_static flag to the species struct?

tnbernard:
The following tests are valgrind clean: rt_gk_static_3x2v_p1 and rt_gk_step_out_2x2v_p1. This branch is ready for merge.

apps/gyrokinetic_update_ssp_rk3.c
tnbernard: No, it still lives in info.

manauref (changes requested):
There are two main comments I wanna make
1. I like the suffix _dynamic to constrast _static. We should use that more. When I get a chance I'll incorporate that into the dynamic/static field capability. Likewise, where you see two options, please use _dynamic, rather than leaving the dynamic option without a suffix.
2. Speaking within app only, disregarding the user: Several of your _dynamic and _static methods are public, meaning that gyrokinetic.c sees them. I think this is not correct. gyrokinetic.c and update_ssp_rk3.c should only see non _dynamic/_static methods, whose contents are left to the species to define.

manauref:
One more thing: can you run some regression tests through valgrind (running it for 3 time steps is enough) to make sure we are not messing up memory please? I suggest running at lease

rt_gk_asdex_3x2v_p1
rt_gk_ion_sound_nonuniformv_1x2v_p1
rt_gk_lapd_cyl_3x2v_p1
rt_gk_ltx_boltz_elc_1x2v_p1
rt_gk_sheath_2x2v_p1
rt_gk_sheath_nonuniformv_2x2v_p1
rt_gk_wham_2x2v_p1
and maybe
rt_gk_step_out_2x2v_p1
but this last one may take a long while to run through valgrind

tnbernard:
I will need to replace this branch with a more recent, improved branch.

PR #569, opened 2025-01-14
Hackathon 2024 branch including parallelized by quadrature point projection routines, refactored reactions, and implicit BGK

JunoRavin:
This branch brings together the work of the 2024 hackathon and other collected developments of the second half of 2024 which support @dingyunl's work on implicit BGK and @tnbernard's work on neutral-plasma coupling. This branch also re-implements the development work of @Maxwell-Rosen on an improved GPU parallelization strategy for conf-phase space weak multiplies by defining a new parallelized accumulation method that is both thread-safe and between 20-40x faster for production runs (thus leading to significant speed ups in, e.g., reactions which must repeatedly do conf-phase space weak multiplies to get the ionization, recombination, and charge exchange updates). 

The main updates are as follows:
1. The old proj_on_maxwellian and proj_on_bimaxwellian routines are deprecated, with the split formalized into the gk_maxwellian and vlasov_lte updaters that include the new GPU parallelization strategy which:
a. Gets the configuration space quantities at configuration space quadrature points
b. Gets the distribution function at phase space quadrature points (thus in general launching tens to hundreds of more threads)
c. Uses the cublas_mm optimized matrix-matrix multiply routine to convert the quadrature point representation back to the modal representation. 
2. Since more memory is now allocated (there is a phase space array internal to the projection routines to optimize the quadrature point to modal representation computations), I have split off the Maxwellian projection routines from their internal allocation to have a single "LTE" object in both gyrokinetics and Vlasov (LTE = local thermodynamic equilibrium). Thus a gk_species and gk_neut_species always make an LTE object, and this LTE object is manipulated by BGK collisions and neutral reactions, and could be utilized by other parts of the app too such as the initial projection to minimize the amount of memory being allocated. I elected to utilize the LTE name across Vlasov and GK because abbreviations of Maxwellian would not work (max and maxwell could be confusing) and the full name "maxwellian" everywhere was making certain functions have overly verbose names in my opinion. 
3. A new set of bin_op routines for conf-phase space multiplies which are accumulation methods have been added, with the option on GPUs for the accumulation methods to be parallelized by component. Before when @Maxwell-Rosen experimented with this feature, he was experiencing race conditions from the in-line weak multiplications we do. In many places, such as when we multiply by Jacobians, we still need in-line operations, but in other places such as neutral reactions, it is safe to do accumulations (the final result of reactions is an accumulation onto the rhs array) and with accumulations we can safely assume distinct arrays have been allocated and no race conditions will occur. We can thus speed up certain configuration space-phase space weak multiplications by factors of 20-40 and significantly reduce the cost of operations such as reactions. 
4. Reactions have been refactored to
a. Utilize the new projection routines (specifically the LTE updater for obtaining the Maxwellians from either the neutral species or plasma species)
b. Correct the moments of the projected Maxwellians if desired. 
c. Perform a set of accumulations onto the total reaction distribution function and the RHS array so we can utilized the parallelized by component conf-phase weak multiply routines
d. Split off the primitive moment computations from the reaction rate computations in anticipation of the canonical PB refactor of neutrals, which do not require as many manipulations of the moments to perform the plasma-neutral coupling. 
5. A new gyrokinetic operator split method has been added to perform the implicit BGK step following an RK3 step. This function is identical to the vlasov_update_op_split which has successfully utilized an implicit BGK update implemented by @johnson452 for high collisionality kinetic neutral calculations. 

The vast majority of the added files in this PR are the new bin_op kernels (both accumulate and accumulate_comp_par) for all our supported basis types and polynomial orders. Some final clean up is needed before the merge but comments are welcome.


manauref:
Lesson for the future: this is a canonical example of how not to manage branches and PR:
- 103 commits (oldest being 8 months old)
- 321 files changes
- >23,000 lines added
- >14,000 lines modified
Recall that after PR #359 there was some discontent with big PR (or that specifically big PR), and the DR/PR policy was put in place in part to prevent this situation.

This PR could've easily been managed as 3-4 separate ones, and we should do that more in the future.

regression/rt_gk_bgk_3x2v_p1.c
```diff
@@ -427,16 +428,17 @@ main(int argc, char **argv)
       .temp = evalTempElcInit,
       .ctx_temp = &ctx,
     },
+
+    .correct_all_moms = true, 
+    .use_last_converged = true, 
+    .iter_eps = 1e-12,
+    .max_iter = 10,
```
manauref, line 435: These inputs should be consolidated into a `maxwellian_correction_inp` struct, rather than placed loosely here, and then we can add that struct as an input to ICs, collisions, sources, etc (wherever the user wants to make corrections).

manauref (changes requested):
Have you tried BGK with nu(x,t) now? there are a lot of changes and we should make sure that still works (maybe make sure it works with implicit BGK too)

JunoRavin:
Yes BGK with nu(x,t) works now. @dingyunl has been running with this capability in this branch. So long as nu(x,t) is still just a piecewise constant this functionality is supported.

JonathanGorard:
> Lesson for the future: this is a canonical example of how not to manage branches and PR:
> 
> * 103 commits (oldest being 8 months old)
> * 321 files changes
> * 23,000+ lines added
> * 14,000+ lines modified
> 
> Recall that after PR #359 there was some discontent with big PR (or that specifically big PR), and the DR/PR policy was put in place in part to prevent this situation.
> 
> This PR could've easily been managed as 3-4 separate ones, and we should do that more in the future.

This is not a constructive comment - this branch was actively discussed within leadership meetings more-or-less since its inception, and, crucially, all aspects of the design were made transparent from the beginning, and were openly discussed with relevant members of the core architecture team throughout the entire process. Long-lived branches (though of course not universally desirable) are not in themselves a problem, and indeed are sometimes inevitable due to external factors: this is one such case.

manauref:
This comment is not meant to be constructive for this PR, but a lesson for
the future as the comment says.

dingyunl:
This is something Mana and I discussed last week. I tried to test the new omega_H restriction implemented on gk-g0-app by restarting a test rt_gk_bgk_im_asdex_extend_2xIC_3x2v_p1.c on a local branch with the latest gk-g0-app merged into gk-g0-app-gpu-hack2024 from the frame before the crash. However, phi in the frame I restarted from and the initial phi in the restart simulation are not consistent. 
gk_bgk_im_asdex_extend_2xIC_3x2v_p1-field_44_xz
gk_bgk_im_asdex_extend_2xIC_3x2v_p1-field_44_xz_initial
I have been using the old version of the gk-g0-app-gpu-hack2024 (commit 839d809a6a4998f845b3bc7d574ea1f94e55fba7 by Jimmy on Jan 14th), and I don't know how much it matters and if it is solved.

ammarhakim:
> > Lesson for the future: this is a canonical example of how not to manage branches and PR:
> > 
> > * 103 commits (oldest being 8 months old)
> > * 321 files changes
> > * 23,000+ lines added
> > * 14,000+ lines modified
> > 
> > Recall that after PR #359 there was some discontent with big PR (or that specifically big PR), and the DR/PR policy was put in place in part to prevent this situation.
> > This PR could've easily been managed as 3-4 separate ones, and we should do that more in the future.
> 
> This is not a constructive comment - this branch was actively discussed within leadership meetings more-or-less since its inception, and, crucially, all aspects of the design were made transparent from the beginning, and were openly discussed with relevant members of the core architecture team throughout the entire process. Long-lived branches (though of course not universally desirable) are not in themselves a problem, and indeed are sometimes inevitable due to external factors: this is one such case.

I agree with @JonathanGorard, and to a certain extent @manauref also. In general, we should not allow branches to live for so long. However, in some situations things evolve in a way that this becomes inevitable. This is the case here, and also in other parts of the code (for example, the gk-g0-app branch). 

One thing, though, that helps here is that, as @JonathanGorard mentioned, is that this work has had considerable scrutiny and many of the changes were discussed in detail. So, overall I think this branch is fine. Of course, shorter lived branches are better, but this work is fine also.

As to constructiveness of comments: it is best to express such views in the G-Leadership meetings. Let us leave it at that.

JunoRavin:
Just adding a comment here, as well as the recent commit, for a public record of a design decision I made after discussions with @tnbernard and @akashukla.

By default now, *no* moments divide out the Jacobian after they are computed, so as you can see in https://github.com/ammarhakim/gkylzero/pull/569/commits/8efb59cf33752b3069a30fa8a32e1181ea1843cf
I needed to make a small bug fix after missing this change in the merge of the gk-g0-app into the hackathon branch following the write refactor. 

The reason for this design change was two-fold:
1. Having some moments already have the Jacobian divided out while others did not required us to internally keep track of who has the Jacobian and who doesn't (and thus writes needed conditionals like the above). 
2. *Not* dividing out the Jacobian by default allows a nice manipulation of updaters such as reactions to utilize Jn in their final RHS update. So before many of the reaction updaters required three weak multiplications: J, n, and sigma (the reaction rate). Now they just need two: Jn and sigma. I think this helps a lot in clarity of the update and also reduces cost so @tnbernard, @akashukla, and I see this as a win-win.

akashukla (dismissed):
3x Step Conservation Tests pass in this branch.

ammarhakim:
I would like this branch merged ASAP. Please let us do this before we meet on Wednesday for the GK meeting, if not **much** earlier. 

Some of the comments here and elsewhere are constructive and should be addressed. Others are not. In these situations it is best to merge and continue testing in production branches.

ammarhakim (approved):
This work is ready to merge. We will continue with further debugging in the production branch. At this point, due to the nature of the work we need work used so we discover the remaining issues. Not much will be gained from waiting much longer.

@manauref Seems the blocking changes have been completed. Please unblock if this is the case. We have two other approval.

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1540,6 +1652,15 @@ void gk_species_apply_bc(gkyl_gyrokinetic_app *app, const struct gk_species *spe
...
 void gk_species_coll_tm(gkyl_gyrokinetic_app *app);
 
+/**
+ * Fill stat object in app with total number of iterations
+ * used to correct moments in LTE projection object.
+ * Also fills stat object with number of times correction object called. 
+ *
+ * @param app App object to update stat timers
+ */
+void gk_species_n_iter_corr(gkyl_gyrokinetic_app *app);
```
manauref, line 1662: can you rename this something like gk_species_maxwellian_corr_n_iter ?

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1643,6 +1764,107 @@ void gk_neut_species_moment_calc(const struct gk_species_moment *sm,
...
+
+/**
+ * Initialize species lte object.
+ *
+ * @param app Gyrokinetic app object
+ * @param s Neutral species object 
+ * @param lte Neutral species lte object
+ * @param corr_inp Input struct with moment correction inputs
+ */
+void gk_neut_species_lte_init(struct gkyl_gyrokinetic_app *app, struct gk_neut_species *s,
```
manauref, line 1777: same as my comment for gk_species_lte above: we are not going to be using Maxwell-Juttner distributions here, so let's just call this _maxwellian instead of _lte. Hence rename
- gk_neut_species_lte_init -> gk_neut_species_maxwellian_init
- gk_neut_species_lte_from_moms -> gk_neut_species_maxwellian_from_moms
- gk_neut_species_lte -> gk_neut_species_maxwellian_from_distf
- gk_neut_species_lte_write_max_corr_status -> gk_neut_species_maxwellian_write_corr_status
- gk_neut_species_lte_release -> gk_neut_species_maxwellian_release

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1870,6 +2105,15 @@ double gk_neut_species_rhs(gkyl_gyrokinetic_app *app, struct gk_neut_species *sp
...
 void gk_neut_species_apply_bc(gkyl_gyrokinetic_app *app, const struct gk_neut_species *species, struct gkyl_array *f);
 
+/**
+ * Fill stat object in app with total number of iterations
+ * used to correct moments in LTE projection object for neutral species.
+ * Also fills stat object with number of times correction object called. 
+ *
+ * @param app App object to update stat timers
+ */
+void gk_neut_species_n_iter_corr(gkyl_gyrokinetic_app *app);
```
manauref, line 2115: -> gk_neut_species_maxwellian_corr_n_iter

manauref (changes requested):
Aside from the minor things I mentioned throughout the files, I think we should make 3 main things with this PR before we merge (I'm happy to help, just let me know):
1. I think we should remove the LTE terminology from the GK app. a) The reasons that terminology came about (to accommodate Maxwell-Boltzmann and Maxwell-Juttner distributions) are not valid here, and we are only going to be using Maxwellians in GK. b) It's best to favor the language that is common in GK literature/community/education, so it's easier for people to get into and use the code. So for both gk_species and gk_neut_species,  we should rename lte -> maxwellian (at the app level, not in zero since I understand we are trying to reuse the lte updater also used by the Vlasov app).
2. Can you check that: a) the new and modified unit tests are valgrind clean? b) have you checked relevant regression tests to make sure they produce expected results and are still valgrind clean? I recommend looking at (checking the field_energy is often sufficient, and for valgrind just running 3 time steps `-s 3` is enough)
rt_gk_d3d_iwl_2x2v_p1
rt_gk_d3d_iwl_3x2v_p1
rt_gk_ltx_boltz_elc_1x2v_p1
rt_gk_rad_nonuniformv_1x2v_p1
rt_gk_sheath_nonuniformv_1x2v_p1
rt_gk_step_out_2x2v_p1
rt_gk_wham_nonuniformx_2x2v_p1_numeric
and a couple of the new BGK tests, maybe one for ASDEX and one implicit.
3. Where is the maxima for all the changes to the kernels? i.e. for the BGK, bin_op and neutral kernels. I didn’t see any mention of an accompanying gkylcas PR.

ammarhakim:
As this discussion has gone on for so long, I would like to step in and make the following comments:

- I want the name "lte" to remain unchanged. We all know the GK terminology and this name is perfectly appropriate here. To suggest that one is not aware of GK literature is somewhat condescending. **I do not want this argument to continue, and I do not want such comments in the future**. 
- **I want this branch merged ASAP**. We should just run a few tests and merge, then we can rebranch
- As guidelines for the future: I think these discussions are becoming extremely low-level, drawn-out, and unhelpful.

I do not want any more comments after this. Please merge and let us take the discussion to a future, in-person meeting.

unit/ctest_mom_gyrokinetic.c
```diff
@@ -945,21 +970,21 @@ void test_2x2v_p2_cu() { test_2x2v(2, true); }
...
+// { "test_1x1v_p2", test_1x1v_p2 },
   { "test_1x2v_p1", test_1x2v_p1 },
-  { "test_1x1v_p2", test_1x1v_p2 },
-  { "test_1x2v_p2", test_1x2v_p2 },
+// { "test_1x2v_p2", test_1x2v_p2 },
   { "test_2x2v_p1", test_2x2v_p1 },
-  { "test_2x2v_p2", test_2x2v_p2 },
-//  { "test_3x2v_p1", test_3x2v_p1 },
+// { "test_2x2v_p2", test_2x2v_p2 },
+// { "test_3x2v_p1", test_3x2v_p1 },
```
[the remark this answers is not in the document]
JunoRavin: This wasn't me and in fact there are no 3x tests in the file, so I'm deleting this line.

regression/rt_gk_bgk_im_cross_relax_1x2v_p1.c
```diff
@@ -0,0 +1,714 @@
...
+      .upar = evalElcUparInit,
+      .ctx_upar = &ctx,
+      .correct_all_moms = true,
+    },
+    .correct = {
+      .correct_all_moms = true, 
+      .use_last_converged = true, 
+      .iter_eps = 1e-12,
+      .max_iter = 10,
+    },
```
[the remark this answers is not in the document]
JunoRavin: Right now, this is now the way the correction updater works. These parameters such as use_last_converged, iter_eps, and max_iter are stored within the updater when the updater is created. 

So if one creates the correction updater, then every object which projects Maxwellians (BGK, reactions, etc.) will correct that Maxwellian. 

The reason for this is because the correction updater has its own internal projection updater for Maxwellians, and now on GPUs this allocates extra memory to optimize the quadrature point evaluation and utilize the optimized matrix-matrix multiplication routines. Thus, right now I don't want to create a correction object for every single operator that could use it. 

Going forward, a strategy could be:
1. Make one correction updater with some specified parameters (specific max_iter, iter_eps, and use_last_converged). 
2. Allow updaters to determine individually whether they will use the updater with a boolean stored in their respective struct (so every operator, BGK, reactions, etc. for every type of operation being done, can specify this input boolean)

But a better strategy might be to revisit the implicit scheme @johnson452 worked out for general reactions and to keep things as is (since this requires that every operator being implicitly updated is corrected).

JunoRavin:
I have checked the tests requested and found machine precision differences between the hackathon branch and gk-g0-app, which is to be expected because the projection routines have all changed, but everything looks good. Simulations are also valgrind clean except for a valgrind issue in the twist-shift boundary conditions, but this issue is in gk-g0-app too and needs a more careful look in a new branch.

JunoRavin:
Final notes before the merge: 
1. @tnbernard, @akashukla, and @dingyunl have all run fairly strenuous tests of this branch (in @dingyunl's case, full production calculations utilizing geometry, implicit BGK collisions, nonuniform velocity space, etc.). While of course automatic testing will be an important component of PRs going forward to insure no edge cases are missed (such as the recent small breaking of zero flux boundary conditions), this branch has not been completely isolated in its development and I have leaned on various folks to help me test the full system. 
2. The charge exchange CAS needs to be updated to go along with this branch. I did not modify the charge exchange maxima because there was an outstanding PR for how the charge exchange kernels were organized, and the only change in this branch is one in which a few arrays are renamed and indexed slightly differently. Since that PR is now merged, @tnbernard and I will take care of this as soon as we can. 
3. The bin_op CAS is already in gk-g0-app in gkylcas. This is because @Maxwell-Rosen filed the PR for this CAS when he originally did the work on this front, however this CAS was merged before realizing that what was happening in gkylzero had race conditions. I did some slightly polishing of the CAS directly in the gk-g0-app branch of gkylcas, but just noting that this CAS is already there. 
4. The BGK cross primitive moments CAS is here: https://github.com/ammarhakim/gkylcas/pull/57

Thanks to everyone for their feedback and reviews on this PR. Many people contributed over these past few months and I greatly appreciate it.

PR #571, opened 2025-01-15
Fix a small oversight in the position map usage in the unit tests

Maxwell-Rosen:
Consider the case where the position map is not provided as an input to the geometry generation. This often happens in a unit test for the geometry modules. The geometry needs to call the position map functions, even if they are identity maps. For this, a temporary position map must be created and destroyed.

### DIFF
--- a/regression/rt_gk_wham_nonuniformx_1x2v_p1_numeric.c
+++ b/regression/rt_gk_wham_nonuniformx_1x2v_p1_numeric.c
@@ -449,7 +449,7 @@ create_ctx(void)
   double mu_max_elc = me * pow(3. * vte, 2.) / (2. * B_p);
   double vpar_max_ion = 20 * vti;
   double mu_max_ion = mi * pow(3. * vti, 2.) / (2. * B_p);
-  int Nz = 128;
+  int Nz = 64;
   int Nvpar = 32; // Number of cells in the paralell velocity direction 96
   int Nmu = 32;  // Number of cells in the mu direction 192
   int poly_order = 1;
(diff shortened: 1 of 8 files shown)


Maxwell-Rosen:
I ran make check and everything passes, except for a few unit tests that fail on commit c9e1a83b0ef40768f6f9726a1b7d7f27eccbf464 which is the merge before this change was implemented. For instance ctest_proj_bimaxwellian_on_basis was broken and is still broken. It's unaffected by this commit. The regression tests for step and gk_sheath look fine too

JunoRavin:
I am fixing ctest_proj_bimaxwellian_on_basis in https://github.com/ammarhakim/gkylzero/pull/569 if we don't want to worry about that test at the moment

JunoRavin:
@Maxwell-Rosen is this the right logic? In unit tests we create identity velocity maps. Should we not be instantiating identity position maps in unit tests and then we don't need this logic? @ammarhakim @manauref be curious on your perspective on this.

Maxwell-Rosen:
I started by changing all the unit tests to initialize an identity position map, but it got very tedious and I realized about 30 to 50 unit tests need to be changed. I thought of this solution because it seems more foolproof than requiring a specific input. I think this is more robust design.

Maxwell-Rosen:
I ran gk_wham with and without non-uniform grids and ran all unit tests. Some things were a little different on Stellar, and there was a minor bug that I think was due to the optimizers. I ran some GPU unit tests, and they look good, but there isn't any way to `make check` on a GPU node on Stellar. I wondered how to use the ci for parallel regression testing on a compute node.

JunoRavin (approved):

Issue #554, opened 2024-12-11
[DR] Refine/coarsen GK distributions upon restarts


manauref:
The ability to refine/coarsen simulations upon restart is helpful for:
- Performing earlier stages with a coarser mesh to run them faster/cheaply, only using high res later when certain things (like global profiles or low-k modes) have saturated.
- Performing earlier stages with a different resolution that may be more robust against transients, and restarting with a different resolution suitable for later stages.

Here we propose to:
1. Leverage the existing infrastructure, used to load lower dimensional simulations for a higher dimensional simulation (see PR #479 ), to load a distribution with the same grid extents but a different resolution/number of cells. That is, we would restart not with `-r` in the command line, but by using (e.g. for electrons)
```
.init_from_file = {
  .file_name = "sim_with_different_res-elc_10.gkyl,
  .type = GKYL_IC_IMPORT_F,
}.
```
in the species table in the input file.
2. Introduce a new updater, `dg_interpolate`, which interpolates a DG field from a donor to a target grid, where these two differ only in the number of cells. This operator is based on ideas of weak equality between overlapping grids, like what we used in multigrid in g2 or in twist-shift. In a nutshell if you consider this diagram
<img width="645" alt="GridTransfer" src="https://github.com/user-attachments/assets/484c3a38-49c1-4fb4-a058-3df9535945ca" />
and we are refining (going from $K_i^{\ell+1}$ to $K_{2i}^{\ell}$, we project the solution in $K_i^{\ell+1}$ onto the basis in the space of $K_{2i}^{\ell}$, being careful in accounting for the different definitions of the logical coordinates in each of these cells. With this formulation, if applied to a kinetic distribution function, the moments are exactly conserved.

Previously it was shown, in earlier maxima, that one can create a general kernel in which interpolation between any two grids, in any dimensionality and any disparity between cell numbers, is possible. However the kernels that come out of it are THE LARGEST kernels I have seen, with 2D interpolation having lines that are tens of thousands or maybe 100k characters long. Even after I did **some** simplification of maxima I was not able to generate reasonable higher dimensional kernels. So the design I propose now, and is prototyped in the `gk-g0-app_interp` branch, is to only generate kernels that do 1D interpolation (for higher dimensional fields, e.g. 2x2v), and `dg_interpolate` calls these kernels in the 
```
static void
dg_interpolate_advance_1x(gkyl_dg_interpolate* up,
  const struct gkyl_range *range_do, const struct gkyl_range *range_tar,
[...]

PR #574, opened 2025-01-21
GK mesh refinement or coarsening upon restart

manauref:
This follows the DR in issue #554 (and goes with [PR 54 in gkylcas](https://github.com/ammarhakim/gkylcas/pull/54)) regarding the ability to refine or coarsen the mesh when restarting a GK simulation.

With this capability the API
```
.init_from_file = {
  .file_name = "sim_with_different_res-elc_10.gkyl,
  .type = GKYL_IC_IMPORT_F,
}
```
in the species table allows the user to provide a distribution function to restart from that may have different resolution (lower or higher), and the GK app will perform `dg_interpolate` to interpolate this distribution onto the mesh specified in the input file. This `dg_interpolate` updater performs DG interpolation while preserving the moments, so that the total integrated moments are conserved to machine precision.

Here I addressed some points listed at the end of issue #554:
- Periodic directions: the grid extents are not allowed to change, so the skin cell of the new grid doesn't overlap with the ghost cell of the donor grid, so there's no need to use (periodic) ghost cells in the interpolation. We simply fill the interior and the app applies periodic BCs to the interpolated field.
- Allowing -r to use dg_interpolate: we punt on this for now; `-r` and `init_from_file` remain separate (and mutually exclusive) options.
- We have not optimized memory allocation to use this in the time loop.
- This is not meant to be used in parallel because in general one would have to gather the distribution onto a single process, which may take too much memory.
- Interpolating J*f or f: more on this below

**Interpolating J*f or f**

Here J is the total Jacobian, which for now is jacobgeo*bmag. I performed a test where I compared these two approaches. That is:
1. Option 1: what the code does now, the first simulation writes J*f, and the second simulation (with a different grid) reads J*f and interpolates it to the new grid.
2. Option 2: the first simulation writes J*f, and the second simulation (with a different grid) multiplies it by 1/J (specifically, jacobgeo_inv written out by the first simulation), interpolates f, and multiplies it by J (jacobgeo of discretized on the new grid).
I tried this on rt_gk_step_2x2v_p1, where the first simulation ran to t=0.5e-6 with 4x8x16x8 cells and the second with 8x16x16x8 to t=1e-16. It appears that Option 2 doesn't conserve moments exactly. Here's M0 at t=0.5e-6:
Screenshot 2025-01-21 at 11 17 26 AM
Here are the slices through theta=0
Screenshot 2025-01-21 at 11 20 29 AM
Screenshot 2025-01-21 at 11 21 07 AM
The difference looks small, but the error in the integrated M0

- Option 1: (3.0311900516472448e+18-3.0311900516472453e+18)/3.0311900516472453e+18 = -1.65e-16.
- Option 2: (3.0134323968372219e+18-3.0311900516472453e+18)/3.0311900516472453e+18 = -0.0059.
is only machine precision for Option 1. Similar results are seen for the integrated M1.

When we look at the integrated M2 however we see that neither option conserves integrated energy:
- Option 1: ((1.4736549181729155e+32+3.0091772356612618e+32)-(1.4736549181729164e+32+3.0138748701741898e+32))/((1.4736549181729164e+32+3.0138748701741898e+32)) = -0.0010.
- Option 2: ((1.4637562061864033e+32+2.9913452194474142e+32)-(1.4736549181729164e+32+3.0138748701741898e+32))/(1.4736549181729164e+32+3.0138748701741898e+32) = -0.0072.
Qualitatively the M2s look very similar:
Screenshot 2025-01-21 at 12 00 31 PM
I believe this is due to the fact that we are using a new bmag which messes with the conservation of the v^2=vpar^2+2*mu*B/m moment (M2). What I mean exactly is that bmag is discretized on the new grid according to its own procedures and criteria, and that's what's used to compute the M2 moment on the new grid. But our interpolation is done via weak projection, which essentially conserves the moments of vpar and mu, not the moments of mu*B. I can strengthen this conclusion by only looking at the error in the integrated M2par:

- Option 1: ((1.4736549181729155e+32)-(1.4736549181729164e+32))/((1.4736549181729164e+32)) = -6.11e-16.
- Option 2: ((1.4637562061864033e+32)-(1.4736549181729164e+32))/(1.4736549181729164e+32) = -0.0067.

So that with option 1 at least the parallel kinetic energy is conserved. So we'll go with that for now. (This does make me wonder whether it is possible to populate the DG coefficients with mu in the polynomial by projecting onto mu*B instead of just mu.)


zero/gkyl_dg_interpolate.h
```diff
@@ -0,0 +1,50 @@
...
+#include <gkyl_gk_geometry.h>
+#include <gkyl_velocity_map.h>
+#include <gkyl_range.h>
+#include <gkyl_array.h>
+
+// Object type.
+typedef struct gkyl_dg_interpolate gkyl_dg_interpolate;
+
+/**
+ * Create a new upater to interpolate a field from a donor grid to a target
```
JunoRavin, line 14: small typo in comment (updater)

zero/gkyl_dg_interpolate.h
```diff
@@ -0,0 +1,50 @@
...
+ * grid with the same extents but different resolution.
+ *
+ * @param cdim Configuration-space dimensions.
+ * @param pbasis Phase-space basis.
+ * @param grid_do Donor grid.
+ * @param grid_tar Target grid.
+ * @param range_do Phase-space range in donor grid.
```
JunoRavin, line 21: Do you mean to label these ranges as phase-space ranges? I know right now the experimentation is just with the gyrokinetic distribution function, but this seems more generic.

zero/gkyl_dg_interpolate.h
```diff
@@ -0,0 +1,50 @@
...
+#include <gkyl_velocity_map.h>
+#include <gkyl_range.h>
+#include <gkyl_array.h>
+
+// Object type.
+typedef struct gkyl_dg_interpolate gkyl_dg_interpolate;
+
+/**
+ * Create a new upater to interpolate a field from a donor grid to a target
+ * grid with the same extents but different resolution.
```
JunoRavin, line 15: we have restrictions right now on what different resolutions we support right? Can we add comments about this restriction?

zero/gkyl_dg_interpolate_priv.h
```diff
@@ -0,0 +1,169 @@
...
+      assert(false);
+      break;
+  }
+
+  return 0;
+}
+
+static int dg_interp_prime_factors(int n, int *pfs, int pfs_size)
+{
+  // Find the prime factos of number `n`, and put them into `pfs`. We assume
```
JunoRavin, line 107: Typo in comments (factors)

zero/dg_interpolate.c
```diff
@@ -0,0 +1,287 @@
...
+  else {
+    // Mesh coarsening, or no change in resolution (dxRat=1).
+    intStencilSize = 1+ceil(1*(1/up->dxRat-floor(1/up->dxRat)));
+  }
+
+  // This updater will loop through the donor grid. At each cell it will give
+  // the interpolation kernel the target grid cells, one at a time, and the
+  // kernel will add the corresponding contributions to the target field DG
+  // coefficients in that each target grid cell.
+  // There are 3 stencils or regions. For example, in 2D the regions are
```
JunoRavin, line 130: Why in your 2D drawing here are the regions only with respect to x? Aren't there 9 regions in 2D because you could be at a corner in both x and y?

zero/dg_interpolate.c
```diff
@@ -0,0 +1,287 @@
...
+  memcpy(&up->ranges[up->num_interp_dirs], range_tar, sizeof(struct gkyl_range));
+
+  // Create an interpolation updater for each interpolation.
+  up->interp_ops = gkyl_malloc(up->num_interp_dirs * sizeof(struct gkyl_dg_interpolate*));
+  if (up->num_interp_dirs == 1) {
+    up->interp_ops[0] = up;
+  }
+  else {
+    for (int k=0; k<up->num_interp_dirs; k++)
+      up->interp_ops[k] = gkyl_dg_interpolate_new(cdim, pbasis, &up->grids[k], &up->grids[k+1],
```
JunoRavin, line 80: Is there a reason this needs to be a recursive object instead of just storing the information you need in each direction? I understand your doing a bunch of sequential 1D interpolations, but what does making this a recursive object like this get you? Doesn't that mean a lot more memory gets allocated?

zero/gkyl_dg_interpolate.h
manauref: no, will fix.

zero/gkyl_dg_interpolate.h
manauref: Actually we did in g2, or in the initial implementation in g0. We were only able to do grids with prime factors of 2 and 3, e.g. Nx=12 or 16. But last weekend I checked that the current implementation also works with grids that have factors of 5 in them (e.g. Nx=10 or 5). I didn't try factors of 7, 11, 13, etc because we rarely or never use those, but my bet is that it'll work.

zero/gkyl_dg_interpolate.h
manauref: Actually, maybe this was intentional to signal that the current updater only works for phase-space quantities, in fact it only works for GkHybrid basis, because the kernel it selects is a gkhybrid kernel. We'd have to generate Ser and Hybrid kernels (easy) and have the updater pick the correct one based on basis.b_type for it to work with other basis.

zero/gkyl_dg_interpolate_priv.h
manauref: fixed

zero/dg_interpolate.c
manauref: There was a misleading/error in the comment ("in 2D the regions are"). We do 1D interpolation at a time, so we only need to consider stencils/regions in 1D. At any given stage the other coordinates are ignorable.

I tried to do genuine multi-D interpolation for many months, maybe years. But higher than 2D the kernels are too large.

zero/gkyl_dg_interpolate.h
JunoRavin: I guess I was thinking that this is the infrastructure that will be utilized when we extend @JonathanGorard's mesh refinement work to DG, and thus you'll want to do this refinement and coarsening on general DG fields (such as configuration space fields in Maxwell's equations). So we can leave the comment as is for now but there's nothing about this that is phase space specific.

JonathanGorard:
Oops, just saw @manauref's most recent commit, which apparently did exactly this. Apologies! Will test this out now to see whether it performs as expected.

manauref:
yeah I modified the updater so it works for conf-space and vlasov. I need to do more testing though, because there's something funky with p=2 Vlasov. One of the tests is coming out with a higher than expected error in the last DG coefficient of M2, which although potentially benign in this case it still seems suspicious and I'd like to look into it more.

manauref:
al'right, I performed the tests I wanted to perform and @JunoRavin helped me understand why M2 may not be conserved after interpolation with p=2 serendipity (see https://github.com/ammarhakim/gkylzero/pull/574/commits/13f453dc3f444f5c123987cd4c5fc5596e2f0fc5) . This work is ready for merge in my opinion, but happy to incorporate more feedback.

JonathanGorard (approved):

JunoRavin (approved):

Issue #573, opened 2025-01-17
[DR] omega_H restriction of dt in GK solver


manauref:
We have long been plagued by omega_H being unknowingly or knowingly unstable, causing undesirable oscillations and/or time step crashes in our simulations (see issue #347 ). The solution to this is to restrict the time step $\Delta t$ to satisfy the constraint $\max(\omega_H)\Delta t\leq1$ where $\omega_H$ is the frequency of this mode. In a simple slab, periodic, 1D box one can show that

$$\omega_H = k_\parallel v_{te} / k_\perp \rho_s$$.

A source of uncertainty, even in 1D, is choosing an appropriate $v_{te}$ and $\rho_s$.

Part of the challenge in enforcing an $\omega_H$ constraint is that most of our simulations are higher dimensional, in curvilinear coordinates and non-periodic. This means we don't know the true dispersion relation for $\omega_H$. We can however obtain a possibly reasonable, relatively general estimate by assuming a quasi-Fourier representation of the relevant quantities, keeping the real-space representation of metric coefficients, magnetic field, etc. The result of doing so is the expression:

$$\omega_H(t) = q_e\sqrt{\frac{J_x n_e(x,t)}{m_e}}\frac{\mathcal{C}}{J_x B_{\parallel}^*}\frac{k_\parallel}{\sqrt{k_x^2\epsilon_{xx}+k_xk_y\epsilon_{xy}+k_y^2\epsilon_{yy}}}$$

where the wavenumbers are evaluated in computational space (and only the k_x^2 term in the denominator contributes in 2D), and $\epsilon_{ij}$ is the polarization weight in the Poisson equation. In the slab 1D limit this becomes equation B9 of M. Francisquez, et al. Phys. Plasmas 30, 102504 (2023). One benefit of this approach is that it should provide an OK estimate in any geometry, and that we can reduce (i.e. find the max) of this expression over the whole grid with canned routines, rather than having to browse the grid in Cartesian/physical space space measuring the shortest $\lambda_\parallel$ and the longest $\lambda_\perp$ to obtain max and min $k_\parallel$ and $k_\perp$, respectively, in physical space.

With this new frequency estimate in mind, we propose to implement the $\max(\omega_H)\Delta t\leq1$ restriction as follows:
[...]

PR #575, opened 2025-01-22
Enforce an omega_H restriction on the GK time step

manauref:
Following the DR in issue #573 , we implement a restriction of the time step $\Delta t$ according to the criteria 
$\max(\omega_H)\Delta t\leq CFL_{\omega_H}$
where $\omega_H$ is the frequency of the mode and $CFL_{\omega_H}$ is an optional user input (the default value is 1.7 following Durran's Comp Methods for Fluid Dynamics figure 2.4). For the frequency we use the new expression
$\max(\omega_{Hs})(t) = q_s\max\left(\sqrt{\frac{J_x n_s(x,t)}{m_s}}\right)\cdot\max\left(\frac{\mathcal{C}}{J_x B_{\parallel}^*}\frac{k_\parallel}{\sqrt{k_x^2\epsilon_{xx}+k_xk_y\epsilon_{xy}+k_y^2\epsilon_{yy}}}\right)$

I tested this with 4 different regression tests/simulaitons. They all give nearly identical results; the biggest difference is seen in gk_step_out_2x2v_p1 and even then the difference is very tiny. The initial time step ratios are as follows (ratio of dt with omega_H restriction to dt without):

- rt_gk_sheath_2x2v_p1: 3.81306e-09/4.11872e-09 = 0.92.
- rt_gk_step_out_2x2v_p1: 5.04698e-09/1.18233e-08 = 0.43.
- rt_gk_mirror_kinetic_elc_1x2v_p1: 1 (dt=3.03505e-11).
- rt_gk_bgk_im_asdex_extend_2xIC_3x2v_p1: 1 (dt=1.00472e-10).
The latter is an LBO version of a simulation @dingyunl is running, started from a 2x2v sim.

Here we can see that the restriction doesn't behave equally for each type of simulation, as we would expect. It dropped the time step for STEP by 60%, but it kept the ASDEX and mirror time steps the same. It's possible that the default $CFL_{\omega_H}$ should be decreased, if we want to be conservative at the cost of some simulations running more slowly. I say this because:
1. In conservation tests elsewhere I see that energy balance time traces have fewer oscillations in rt_gk_sheath_2x2v_p1 when I use cfl_frac=0.5.
2. Initially I implemented $CFL_{\omega_H}=1$, and this gave a dt 2X smaller in rt_gk_sheath_2x2v_p1 and 4X smaller in rt_gk_step_out_2x2v_p1. It seems like a big hit, but actually the total number of time steps was not proportionally large: for STEP it was only 3X larger and there were fewer failed RK stages.

This branch also adds a new dg_array_reduce procedure which assumes a gkyl_array contains a DG field, evaluates the DG field at Gauss-Legendre nodes, and performs a reduction of those values over the cell and over the whole array. The existent array_reduce simply does component-wise reduction and doesn't evaluate the components as part of a DG expansion.

Additional points of interest:
- Technically we could consider the $\max(\omega_H)$ to be a ratio of the maximum real eigenvalue of the GK operator from parallel dynamics ($\lambda_{\parallel,\max}$) and the minimum eigenvalue of the Poisson operator ($\lambda_{\perp,\min}$). In the expression above we are approximating 
$$\sqrt{\lambda_{\parallel}} = q_s\sqrt{\frac{J_x n_s(x,t)}{m_s}}\cdot \frac{\mathcal{C}}{J_x B_{\parallel}^*}k_\parallel$$
$$\sqrt{\lambda_{\perp}} = \sqrt{k_x^2\epsilon_{xx}+k_xk_y\epsilon_{xy}+k_y^2\epsilon_{yy}}$$
so rather than evaluating the maxes like we do in the first expression at the top, we could instead do it like $\max(\omega_H) = \max(\sqrt{\lambda_{\parallel}})/\min(\sqrt{\lambda_{\perp}})$. I don't know what impact this would have, but it does require one extra weak multiplication.
- The maxes in the equation at the top are done using cell-center values only. I tried evaluating the max at Gauss-Legendre nodes and using the max of those. Such procedure gave a slightly more stringent constraint. We could switch to that if needed.
- The density under the square root should really be some equilibrium density, something like a flux surface and time average like $<n_s>_{y,z,t}$. But we don't really compute this yet, and instead of adding that here we simply use the max (sqrt) density. This will give a more stringent restriction in the presence of O(1) fluctuations, but we'll live with it for now until more people use this.


apps/gk_field.c
tnbernard, line 1: I'm assuming the weight-->epsilon changes account for the current conflict with gk_field

tnbernard (approved):
The changes to in app look fine to me. I feel less qualified to comment on the lower level changes in zero and will defer to Ammar for those.

JunoRavin (approved):

Issue #563, opened 2025-01-06
[DR] Full Specification of Gyrokinetic BCs in MB App


akashukla:
The current `gkyl_block_physical_bcs` struct used to specify BCs in the MB app does not have enough options to use certain gyrokinetic bc types such as `FIXED_FUNC` which requires a gyrokinetic  specific projection object. To enable  use of these bcs, we have added a new physical bc specification  struct to  `gkyl_gyrokinetic_multib.h`  : 

```
struct gkyl_gyrokinetic_block_physical_bcs {
  int bidx; // block index
  int dir;  // direction in which BC is specified
  enum gkyl_edge_loc edge; // which edge this BC is for
  enum gkyl_gyrokinetic_bc_type bc_type; // BC type flag.
  void (*aux_profile)(double t, const double *xn, double *fout, void *ctx); // Auxiliary function (e.g. wall potential).
  void *aux_ctx; // Context for aux_profile.
  double aux_parameter; // Parameter for aux_profile (maybe redundant).
  struct gkyl_gyrokinetic_projection projection; // Projection object input (e.g. for FIXED_FUNC).
};
```

In addition, we have united the gyrokinetic species and field BCs into a single enum `gkyl_gyrokinetic_bc_type` . There are helper functions in `gyrokinetic_multib.c` which translate the enum into the correct number to be passed to the field or species inputs.

The  only change to input files is in MB gyrokinetic input files where the enum prefixes are slightly different (`GKYL_BC_GK_` rather  than `GKYL_`) and the `gkyl_block_physical_bcs` structs are replaced with `gkyl_gyrokinetic_block_physical_bcs` structs.

A note: the field bcs for the MB gyrokinetic app are currently incorrectly passed (Current MB  input files work by coincidence since the `GKYL_POISSON_DIRICHLET` and `GKYL_FIELD_DIRICHLET` enums have the same number) . This issue is also repaired in this DR. 

The prototyping work is in the branch: https://github.com/ammarhakim/gkylzero/tree/gk-g0-app_gkbc_enhance_prototype.

JunoRavin:
DR implemented in https://github.com/ammarhakim/gkylzero/pull/580

PR #580, opened 2025-01-27
GK BC Enhancement

akashukla:
The current gyrokinetic multib app lacks the ability to fully specify gyrokinetic BCs. For example, BC_FIXED_FUNC cannot be used, and auxillary wall potentials cannot be specified. The full details are in the Design Review #563 . This PR enables full specification of the gyrokinetic species and field BCs in the gyrokinetic MB app.


manauref (changes requested):
Akash, quick change please:
can you rename regression tests so they are rt_gk_multib_... so we know they are GK tests?

manauref (approved):

JunoRavin (approved):

PR #582, opened 2025-01-28
Fix the size of the polarization weight array in gk_field for 1x

manauref:
Commit message:
The size of the polarization weight (times geo coeffs) epsilon in gk-field is the size of the global range in 1x sims, which interferes with the omega_H calculation in 1x when running in parallel. There was a confusing thing in that 1x epsilon is global but 2x/3x epsilon is local. Resolve this confusion by making epsilon also local in 1x and briefly creating a global epsilon which fem_parproj only needs during initialization. 1x and 2x sims produced expected results in parallel on my laptop (thanks to @akashukla for pointing this bug out).

### DIFF
--- a/apps/gk_field.c
+++ b/apps/gk_field.c
@@ -68,6 +68,7 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
     assert(app->cdim == 1); // Not yet implemented for cdim>1.
 
   f->epsilon = 0;
+  struct gkyl_array *epsilon_global = 0;
   f->kSq = 0;  // not currently used by fem_perp_poisson
   double polarization_weight = 0.0; 
   double es_energy_fac_1d_adiabatic = 0.0; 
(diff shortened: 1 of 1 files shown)


JunoRavin (approved):

apps/gk_field.c
```diff
@@ -81,6 +82,9 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
     }
   } else {
 
+    // Allocate array for the polarization weight times geometric coefficients.
+    f->epsilon = mkarr(app->use_gpu, (2*(app->cdim/3)+1)*app->confBasis.num_basis, app->local_ext.volume);
```
akashukla, line 86: Why isn't the second argument the same as it used to be for cdim>1 ? It was : (2*(app->cdim-1)-1)*app->confBasis.num_basis before

akashukla: Ok actually I see this is fine. if cdim = 3 we get 3 components, if cdim=2 we get 1 component because 2/3=0 here (correct?), and if cdim=1 we get  1 component because 1/0 = 0 here. This is correct.

akashukla (approved):

Issue #581, opened 2025-01-28
[DR] Restructure of write functions


tnbernard:
Move the write statements which currently live in the gyrokinetic app to their respective objects. See an example of this for the gk_species app write functions in the following branch: https://github.com/ammarhakim/gkylzero/tree/gk-g0-app-restructure-write-funcs

Whereas we currently have a write statement in the gyrokinetic app as: 
```
void
gkyl_gyrokinetic_app_write_species(gkyl_gyrokinetic_app* app, int sidx, double tm, int frame)
{
  struct gk_species *gk_s = &app->species[sidx];

  if (!gk_s->info.is_static || frame == 0) {
    struct timespec wst = gkyl_wall_clock();
    struct gkyl_msgpack_data *mt = gk_array_meta_new( (struct gyrokinetic_output_meta) {
        .frame = frame,
        .stime = tm,
        .poly_order = app->poly_order,
        .basis_type = app->basis.id
      }
    );

  ...
    
    gk_array_meta_release(mt);  
    app->stat.diag_tm += gkyl_time_diff_now_sec(wst);
    app->stat.ndiag += 1;
  }
}
```

The contents will now be defined within its respective object, gk_species in this case. It will appear instead as 

```
void
gkyl_gyrokinetic_app_write_species(gkyl_gyrokinetic_app* app, int sidx, double tm, int frame)
{
  struct gk_species *gks = &app->species[sidx];
  gk_species_write(app, gks, tm, frame);
}
```

JunoRavin:
DR implemented in https://github.com/ammarhakim/gkylzero/pull/585

PR #583, opened 2025-01-30
Static species functionality into gk-g0-app

tnbernard:
This is related to DR: https://github.com/ammarhakim/gkylzero/issues/450. 

It adds static functionality for gk_species, demonstrated in the rt_gk_static_3x2v_p1.c regression test. It also restructures logic related to static species, both gk and gk_neut. It removes (most of*) the is_static if statements from the gyrokinetic and rk3 update files. Function pointers are called within the species app, which differ depending whether the species is static or dynamic. These functions are defined privately to the species app and relate to the rhs, apply_bc, release, accumulate (now `step_f`), combine and copy_range methods. 

It addresses all comments in the now closed PR https://github.com/ammarhakim/gkylzero/pull/567

Several reg tests were tested with Valgrind on perlmutter in the last PR and were clean. Perlmutter is down the rest of the week, preventing additional Valgrind testing. The output from the rt_gk_sheath_3x2v_p1.c and rt_gk_step_out_2x2v_p1.c tests match the gk-g0-app branch exactly. 

*The other is_static if statements in gyrokinetic.c are in the write functions. A similar restructuring of the write statements for all objects will eliminate this type of logic. This is described in DR https://github.com/ammarhakim/gkylzero/issues/581.


JunoRavin (approved):

manauref (approved):
Approving, but we need to come back and run valgrind when Perlmutter or Ammar are back.

PR #584, opened 2025-01-31
Fixed Lua communicators, added neutron stars, plus added GK reactions, neutrals, and other small updates

JonathanGorard:
The main addition in this branch is the capability to initialize the appropriate MPI/NCCL communicators to run both multi-CPU and multi-GPU simulations through Lua. By default, if the `gkyl` executable is built with CUDA, MPI and/or NCCL, it is assumed that you intend to run a single-GPU, multi-CPU, and/or multi-GPU simulation, respectively, unless told otherwise. The flags `-S` and `-G` can be passed in to disable MPI/NCCL and to disable CUDA, respectively (if anyone has a better suggestion for which letter to use in the latter case, please let me know - I'm not a fan of the "G").

The other updates contained within this branch include the addition of neutron star metrics (already reviewed as part of PR #572), various fixes to the IO and logging triggers so that they now work correctly with restarts, and a few small updates to the Lua Moments and GK apps (including some work-in-progress on adding neutral species and reactions to Lua GK).


apps/gkyl_lw_priv.h
```diff
@@ -160,6 +168,30 @@ gkyl_register_gyrokinetic_radiation_types(lua_State *L);
 void
 gkyl_register_gyrokinetic_radiation_Te_types(lua_State *L);
 
+/**
+ * Add reaction type flags for gyrokinetic species initialization into Lua interpreter.
+ *
+ * @param L Lua state to use.
+ */
+void
+gkyl_register_gyrokinetic_reaction_types(lua_State *L);
```
JunoRavin, line 177: We may want to consider some different naming schemes going forward, something like

reaction_op_type (or operator)
reaction_ion_type
reaction_self_type

so that way someone doesn't wonder what the difference between reaction_type and reaction_self_type is, but I don't think that's a reason to delay this merge. Just making this note now for our records.

JunoRavin (approved):

manauref (approved):

PR #590, opened 2025-02-05
Fix zero-flux BCs in GK solver

manauref:
We didn't realize we messed up `is_zero_flux` passed to the collisionless solver in the static species PR #583. This commit fixes it.

:( we need automated testing.... :(


tnbernard:
Good catch. Was there a particular regression test where this became apparent?

manauref:
> Good catch. Was there a particular regression test where this became apparent?

I was running conservation tests in another branch and noticed it

tnbernard (approved):

PR #595, opened 2025-02-10
Updated mkdeps to build LuaJIT by default

JonathanGorard:
Thanks @JunoRavin and @Antoinehoff for pointing out that the recent configuration file updates to use Lua by default didn't also come with a corresponding change to the mkdeps files to build LuaJIT by default.

Eventually, we will want to modify the way that the Lua wrappers work to ensure that we can build using _only_ vanilla Lua 5.1 (right now this _almost_ works, but fails for some slightly obscure reason due to the way that we're packaging equation objects. It's likely an easy fix), in which case we will need to revisit this default building of LuaJIT. But for now this should work.


JunoRavin (approved):

ammarhakim (approved):
Ready to mereg

PR #597, opened 2025-02-11
Fix with the 1x polarization density inversion.

Maxwell-Rosen:
# Bug fix

Depends on #598 

## Summary

Description: There was an issue with the 1x kernels in the polarization density solve. Since gyrokinetics has a $\nabla_\perp^2$ in the Poisson equation, there is an issue 1x when computing the Poisson equation. Instead, we have $\nabla = k_\perp$. We use this in the kernels.

The prior version applied the polarization scaling to all ions. This isn't good because it applies to all species. This should only apply to one species. A parameter is added to the species input table to select which species should be scaled

Link to gkyl-cas PR: https://github.com/ammarhakim/gkylcas/pull/60

## Solution

This is fixed by noting that in 1x, $\epsilon = -  \epsilon_0 k_\perp^2 / J$, while in 2x,  $\epsilon = -  \epsilon_0 / J$. Thus, we must only multiply the potential by epsilon to get the polarization density. This is fixed by removing the gradients in the kernel files.

Non-uniform position space grids are considered. There is a present "issue" with the polarization solve. It's not really an issue, but an important thing to keep in mind. The projection operation of the potential we give the code must take input of computational coordinates, not physical coordinates. I added a few comments to emphasize this. This operation does not use the position map infrastructure, which a user should expect. Users should expect field-aligned coordinates, not physical coordinates.

Furthermore, the factor of $4/dx^2$ is removed because no gradients are being taken.

Impacted files: `maxima/g0/gk_pol_den/pol_density.mac`

## Community Standards

## Testing:

- [ ] I added a regression test to test for this bug.
- [ ] I added a unit test to test this bug.

## Additional Notes

Here is an example of it working. This is a 1x kinetic electron simulation initialized from a Boltzmann electron simulation. The ion density of the Boltzmann simulation is particle density, but the kinetic simulations have guiding center density, so the ion polarization density must be accounted for. Given that the electron polarization density is small, $n_{e,gc} = n_{e} = n_i$. We scale the ions to account for this.

image

image


manauref (approved):
Just remove the comments "Be careful ... " because you are about to fix that in another PR.

apps/gk_field.c
```diff
@@ -2,13 +2,21 @@
...
 #include <gkyl_dg_eqn.h>
+#include <gkyl_position_map.h>
 #include <gkyl_util.h>
 #include <gkyl_gyrokinetic_priv.h>
 
 #include <assert.h>
 #include <float.h>
 #include <time.h>
 
+void
```
manauref, line 13: this should be a static void

zero/eval_on_nodes.c
```diff
@@ -17,30 +17,66 @@ struct gkyl_eval_on_nodes {
...
+    up->c2p = inp->c2p_func;
+    up->c2p_ctx = inp->c2p_func_ctx;
+  }
 
   return up;
 }
 
 static inline void
-comp_to_phys(int ndim, const double *eta,
+log_to_comp(int ndim, const double *eta,
```
JunoRavin, line 79: what does log to comp mean? comp_to_phys makes sense, so I am a little confused about this name change?

Maxwell-Rosen: This is done to be consistent with proj_on_basis. This identical function is called both comp_to_phys and log_to_comp. when I read comp_to_phys, I think mapc2p or c2fa. Something like that. It's vague what is meant by "phys". How would you explain in long form what this function is doing?

I was going to change all occurrences of log_to_comp throughout the code base, but Mana suggested against that and having that be a separate PR. We aim for a small PR that does not touch too many parts of the code. If an inconsistency is noticed, it should be fixed immediately.

JunoRavin (approved):

Issue #593, opened 2025-02-10
[DR] Improve energy conservation in GK solver: Poisson in slabs, boundary diagnostics & smoothing BC


manauref:
While comprehensively reviewing our collisionless algorithm (see attached document, also shared in an email titled "Notes on GK equations" on Dec 2,, 2024),
[GkeyllGK-nonuniformV-model_20241202.pdf](https://github.com/user-attachments/files/18734938/GkeyllGK-nonuniformV-model_20241202.pdf)
 we found that solving the GK Poisson equation on planes breaks energy conservation in curvilinear coordinates. Thus we set out to prove that energy conservation is broken and that it'd be restored by solving Poisson on z-slabs. In order to do so, we implement:
1. An option to request Hamiltonian (M0, m*M1, H) integrated moments instead of integrated M0, M1, M2par and M2perp.
2. New diagnostics writing the integrated moments of the boundary fluxes.
3. New diagnostics writing the change in total integrated energy during a time step.
4. Changes to fem_poisson_perp so it works in 2x.
5. New diagnostics writing the time-integrated, integrated moments of the boundary fluxes.
6. New diagnostics writing the time-integrated, integrated moments of the source.

With these diagnostics we can examine the exact particle, momentum and energy balance in the GK solver in two ways:
a) Plot the sources, sinks and rate of changes over time. Perfect conservation would mean these add to zero.
b) Plot the total particle, momentum or energy over time, subtracting the amount added by sources and adding back the amount removed by sinks. Perfect conservation would mean this quantity remains constant at its initial value.

Here's an example of plot a) for particle balance (of rt_gk_step_out_2x2v_p1):
<img width="862" alt="Image" src="https://github.com/user-attachments/assets/f1ed6fe4-404f-414d-b3e9-b063b31d56b8" />
The purple line shows our error in particle conservation is qualitatively zero. If one zooms in, and does it for various values of `cfl_frac`, we demonstrate we conserve particles to machine precision. We can also see this in the type of plot b) for particle conservation, for example for gk_sheath_2x2v_p1:
<img width="752" alt="Image" src="https://github.com/user-attachments/assets/d47b4728-1224-4c51-9222-cf414433aa2d" />

We can make similar plots examining energy conservation. For example, here's a type a) plot for total energy balance in rt_gk_step_out_2x2v_p1:
[...]

PR #599, opened 2025-02-12
Fix energy conservation in GK solver when using curvilinear geometry

manauref:
Motivation
-----------

In this work we:
1. Implement the necessary diagnostics to perform a complete energy balance of the collisionless gyrokinetic algorithm.
2. Use such diagnostics to demonstrate that the way we are currently solving the Poisson equation on planes breaks energy conservation when using curvilinear geometry, and that solving it in z-slabs fixes energy conservation.

More details are in DR #593 .

Diagnostics & lack of energy conservation
------------------------------------------

Below we mention Hamiltonian moments. These are (M0, m*M1, 0.5*m*v^2+q*phi). The reason to include M0 and m*M1 this way is that it facilitates checking particle and momentum balance (or lack thereof) with the same infrastructure, python scripts, pgkyl comands, as we'd use to check energy balance. Also, since the integrated moments have by default been (n,upar,Tpar,Tperp), we implemented a species-table input file flag, `.integrated_hamiltonian_moments = true`, to instead requested integrated Hamiltonian moments.

The diagnostics implemented consist of:
- In order to measure the time rate of change of the total energy, we compute the time rate of change of the total energy $E = E_f + E_\phi$ as
$$\dot{E} = \left(E_f^{n+1}+E_\phi^{n+1} - E_f^{n}-E_\phi^{n}\right) / \Delta t$$
where $E_f$ is the integrated Hamiltonian moment of $f$, and $E_\phi$ is the field energy term. We compute the quantities at time $t^n$ at the beginning of the RK step, and those at $t^{n+1}$ at the end of the RK step. These are requested by passing `.time_rate_diagnostics = true` in the species and field tables of the input file.
- The time rate of change of the energy injected by the source is computed as the integrated Hamiltonian moment of the source, $\int\mathrm{d}x~\mathrm{d}v~H_s\mathcal{S}_s$. This requires passing the `.diagnostics` input in the source table of the input file.
- The time rate of change of the energy due to fluxes through the boundaries are computed by: a) collecting the phase-space fluxes through conf-space boundaries in their corresponding ghost cells, and b) computing the integrated Hamiltonian moment of those fluxes. These are requested by passing `.boundary_flux_diagnostics` in the species table of the input file.

The above diagnostics allow us to plot the energy balance of sources, sinks and rates of change. We can do this for particle (see the first plot in DR #593 ), momentum and energy rates of change.

Additional diagnostics are implemented to assess conservation by plotting total energy or particles as a function of time, subtracting the contribution from the sources and adding the contribution from boundary sinks. These are:
- Time-integrated, volume-integrated Hamiltonian moments of the source.
- Time-integrated, volume-integrated Hamiltonian moments of the boundary fluxes.

As shown in the 3rd plot of DR #593 , the energy errors in a curvilinear geo simulation (e.g. a STEP simulation) can be O(1) compared to the sources (granted that plot was made at low resolution, but the problem is concerning still because we'd like to resolve things coarsely along the field line, but geometry can have drastic variations in that direction like near the X-point or a mirror throat). Moreover, all the Gkeyll papers before have said we conserve energy, so it'd be good to maintain and demonstrate that property.

Restoring energy conservation with Poisson solves in z-slabs
-------------------------------------------------------------

Luckily the solution to restoring energy conservation, as far is analysis can tell, is very easy. Use the Poisson solver in slabs. The z-slab GK Poisson solver has been around for a long time (2-3 years?). Here we simply upgraded it so it works in 2x as well.

Making the move to fem_poisson_perp in the GK app simply involved switching 5-10 lines of code to simply call fem_poisson_perp instead of deflated_fem_poisson. Note that one must FEM project phi after using fem_poisson_perp as required by the energy theorem.

With this minor change the errors in energy balance drop dramatically, see the 4th plot in DR #593 .

Other perks
------------

In the course of this work we added some other functionality to help with the particle/energy balance calculation. These are:
a) Functions for writing out the time step. Because we needed these to be written out with the same frequency as the integrated moments, we added to functions:
`gkyl_gyrokinetic_app_save_dt`
`gkyl_gyrokinetic_app_write_dt`
the former is to be called whenever the integrated diagnostics are calculated, this appends a snapshot of dt to a DynVector, and the latter simply writes the dynvector and needs to be called whenever the integrated diags are written. Note that one has to be careful not to write dt at t=0 (I forget exactly why, this may need to be revised).
b) Functions for initializing the field from file or from a function, for the case of static fields. This helped in debugging, and I suspect it'll continue to be useful in the future (we've wanted this for debugging mirrors/nonuniform z grids too).
c) The user can specify which integrated diagnostics moments they want, so it doesn't have to always be n,upar,Tpar,Tperp. This was done to request Hamiltonian moments (M0,m*M1, 0.5\*m\*v^2+q\*phi).
d) We've learned that the parallel smoother is not self-adjoint with Dirichlet BCs, and makes the energy balance a bit worse. So we remove the fem_parbc option from input files and hardcode it under the hood.


ammarhakim (approved):
I am fine with this PR. A couple of questions, though, should be addressed (not top priority):

- Are these diagnostic quantities optional? Specially if phase-space arrays are needed in storing them then we should make them optional
- Do we need both rates and integrated diagnostics? I am fine keeping both, but if we can get away with just one then it reduces the amount of things we need to maintain.

JunoRavin:
I feel like a downside of calculating rates is rates require new phase space arrays (so that you can compute f_dot first and then take its moments) whereas if you only compute the integrated moments, no new phase space arrays are required. So the question is, what do we gain from supporting computing rates when we will likely not want to ever compute rates in production simulations (since we won't want to allocate extra phase space arrays)? 

I think we also need to discuss in our leadership meeting the order of PR merges. This is fantastic work on the discovery of the way we should be doing the Poisson solve in general, and I like the algorithmic approach a lot of solve Poisson with Jrho on the RHS (no need to smooth Jrho), then do the weighted smoother for phi. But we've got a number of pieces coming together between the FLR + biasing, @Antoinehoff's twist-shift branch, and this work, and I just want to make sure main works with the inner wall limited simulations as soon as possible. 

I think we know the procedure: with the revival of fem_perp_poisson and the fact that we smooth phi at the end, we just have to make sure phi is appropriately twisted and shifted after the solve right? And the biasing in the FLR+biasing branch will give us the radial continuity? I know we're all talking independently about this topic but it will be good to continue coordinating to get these various branches merged in the order that works best that also gets us doing limited TCV and DIII-D calculations with better energy conservation asap.

manauref:
FYI, I'm cleaning up the implementation of the diagnostics. See the DR on
boundary flux diagnostic API, for example. So don't merge this until I let
you know it's ready.

In response to @ammarhakim:
1. Yes, the diagnostics are optional. Also, they don't allocate phase-space
arrays, only conf-space arrays and dynVectors.
2. I'm not sure if we **need** both rates and diagnostics. However, it may
be useful to have them both. Part of the cleanup i'm doing is letting the
user choose which they want to get. Personally I both are equally useful.

manauref:
I'll add that if for some reason we want to ditch one of the diagnostics (rates vs. content), I'd prefer to ditch the content ones because to do a proper conservation balance with the content ones we need the time integrated boundary fluxes and sources, and those are more of a pain to keep track of than rates.

apps/gk_field.c
```diff
@@ -222,8 +243,9 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
 
   f->integ_energy = gkyl_dynvec_new(GKYL_DOUBLE, 1);
   f->is_first_energy_write_call = true;
+
   // Factors for ES energy. 
-  f->es_energy_fac = mkarr(app->use_gpu, app->basis.num_basis, app->local_ext.volume);
+  f->es_energy_fac = mkarr(app->use_gpu, (2*(app->cdim/3)+1)*app->basis.num_basis, app->local_ext.volume);
```
Antoinehoff, line 248: why this trick?

apps/gk_field.c
```diff
@@ -408,52 +443,133 @@ gk_field_rhs(gkyl_gyrokinetic_app *app, struct gk_field *field)
   // Compute the electrostatic potential
   struct timespec wst = gkyl_wall_clock();
   if (field->gkfield_id == GKYL_GK_FIELD_BOLTZMANN) { 
+    // Solve phi = phi_s + (Te/e)*ln(n_i/n_i,s).
```
Antoinehoff, line 446: we want to leave this?

Antoinehoff:
Wait but if we push this PR, main will not use deflated Poisson anymore and the discontinuity at the target corner will be back right?

JunoRavin:
I think @manauref is seeking feedback on the PR as is, but the order of operations is to merge your PR https://github.com/ammarhakim/gkylzero/pull/602

then add the necessary pieces to this PR to twist-shift phi in z and fix the corner target in fem_perp_poisson.

manauref:
Regarding @JunoRavin 's comment about the order of operations, we can do it either way @Antoinehoff . We could:
1. Merge this first, and notice that I left `deflated_fem_poisson` in the app, just commented out, so that when we try to merge the clopen PR we simply add an if-statement so that clopen continuous to use `deflated_fem_poisson` until we add the biasing to `fem_poisson_perp`, or
2. Merge the clopen PR first, and when resolving the conflicts we'll do add the if-statement mentioned above.
I don't feel strongly either way. Adding the biasing to fem_poisson_perp will happen in yet another PR.

apps/gk_field.c
manauref: yup, we have 1x sims that run with Boltzmann electrons. Unless you meant something else.

Antoinehoff:
ok maybe the clopen PR is closer to be done so it looks more logical to push it first, then add a hopefully temporary parameter to use the deflated poisson solver in this branch if we are in clopen sim. This will be then a good basis to implement the biasing in Poisson perp because we will be able to compare the two solvers within the same branch in production runs.

tnbernard (commented):
The gk_species_bflux needs to allow calculation of ion flux even if ions are static. The logic in this PR appears to only calculate it for non-periodic, non-zero flux BCs. But this assumes a future and not yet merged capability (recycling), so I guess we can fix that at a future time.

manauref:
@tnbernard only the diagnostic bfluxes are computed for non-periodic,
non-zero flux BCs. The "solver" bfluxes, which are used for recycling, are
computed for all dimensions at the moment (but I think as part of your
recycling branch we should change it so it's only computed for emitting
boundaries (and the sheath boundary for Boltzmann electrons)

apps/gk_species_bflux.c
```diff
@@ -42,13 +134,448 @@ gk_species_bflux_rhs(gkyl_gyrokinetic_app *app, const struct gk_species *species
...
+      bflux->bflux_rhs_func = gk_species_bflux_rhs_solver; 
+    else
+      bflux->bflux_rhs_func = gk_species_bflux_rhs_solver_init; 
+
+    for (int d=0; d<app->cdim; ++d) {
+      for (int e=0; e<2; ++e) {
+        // Allocate solver.
+        struct gkyl_range *skin_r = e==0? &gk_s->lower_skin[d] : &gk_s->upper_skin[d];
+        struct gkyl_range *ghost_r = e==0? &gk_s->lower_ghost[d] : &gk_s->upper_ghost[d];
+        bflux->flux_slvr[2*d+e] = gkyl_boundary_flux_new(d, e, &gk_s->grid,
```
tnbernard, line 527: Why not just initialize flux_slvr for all cases? How is this not an instance of code duplication? What is an example of a case where neither is initialized?

manauref:
A gk-sheath simulation WITHOUT recycling, Boltzmann electrons or boundary
flux diagnostics doesn't need either of the boundary flux operators (solver
or diag).

Because of the myriad of options we are allowing, this was the best logic I
could find.

apps/gk_field.c
Antoinehoff: no I mean the commented line

manauref: that's just a comment, not code.

apps/gk_species.c
```diff
@@ -1486,8 +1566,9 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
   }
 
   // Initialize boundary fluxes for diagnostics and/or neut recycle bcs 
-  gks->bflux = (struct gk_boundary_fluxes) { };
-  gk_species_bflux_init(app, gks, &gks->bflux);
+  // Initialize boundary fluxes for the ambipolar potential solve (after skin ghost ranges are created).
```
Antoinehoff, line 1569: I guess you forgot to remove the former comment line

apps/gk_species_source.c
```diff
@@ -184,16 +211,17 @@ gk_species_source_calc_integrated_mom(gkyl_gyrokinetic_app* app, struct gk_speci
 void
 gk_species_source_write_integrated_mom(gkyl_gyrokinetic_app* app, struct gk_species *gks)
 {
-  if (gks->src.source_id && gks->src.evolve) {
+  if (gks->src.source_id) { // && gks->src.evolve) { // MF 2025/02/04: commenting this out for some testing.
```
Antoinehoff, line 214: this is still in testing phase?

apps/gk_species.c
manauref: true

apps/gk_species_source.c
manauref: actually this change is not needed unless one wants to do in-depth energy balance. I'll change it back.

Antoinehoff:
In the first bullet point of the PR :

> These are requested by passing `fdot_diagnostics = true` in the GK app table of the input file.

Is this still true? It looks like `fdot_diagnostics` does not exist anymore.

Antoinehoff:
Also more important point, is it correct that the bflux diagnostics are not compatible with multi GPU? I try to include them in my baseline simulation but I get a seg fault that is resolved when I comment the `boundary_flux_diagnostics` structure...

I think it would be very useful to have them in production runs so would it be possible to make them multi GPU compatible?

manauref:
> In the first bullet point of the PR :
> 
> > These are requested by passing `fdot_diagnostics = true` in the GK app table of the input file.
> 
> Is this still true? It looks like `fdot_diagnostics` does not exist anymore.

Yeah I changed it. You have to see the regression tests to see how it's done now. But these are only needed if one wants to look at conservation using rates. If you only want boundary fluxes (which you can use to look at conservation through the particle/energy content) this isn't needed

manauref:
> Also more important point, is it correct that the bflux diagnostics are not compatible with multi GPU? I try to include them in my baseline simulation but I get a seg fault that is resolved when I comment the `boundary_flux_diagnostics` structure...
> 
> I think it would be very useful to have them in production runs so would it be possible to make them multi GPU compatible?

I don't see why it wouldn't work on mutiple GPUs, but I'll double check.

Antoinehoff:
> Yeah I changed it. You have to see the regression tests to see how it's done now. But these are only needed if one wants to look at conservation using rates. If you only want boundary fluxes (which you can use to look at conservation through the particle/energy content) this isn't needed

Which regression test are you talking about? `regression/rt_gk_step_out_2x2v_p1.c` this one?

manauref:
I think that should have it, yes. I'll change the PR message shortly

manauref:
> Also more important point, is it correct that the bflux diagnostics are not compatible with multi GPU? I try to include them in my baseline simulation but I get a seg fault that is resolved when I comment the `boundary_flux_diagnostics` structure...
> 
> I think it would be very useful to have them in production runs so would it be possible to make them multi GPU compatible?

Just pushed a fix. Should work on multiple GPUs now.

apps/gk_multib_field.c
```diff
@@ -221,51 +229,86 @@ gk_multib_field_new(const struct gkyl_gyrokinetic_multib *mbinp, struct gkyl_gyr
...
+//    }
+//    // Copy continuous potential back to apps.
+//    for (int bI=0; bI<mbf->num_local_blocks; ++bI) {
+//      struct gkyl_gyrokinetic_app *sbapp = mbapp->singleb_apps[bI];
+//      gkyl_array_copy_range_to_range(sbapp->field->rho_c_global_smooth,
+//        mbf->phi_multibz_smooth[bI], &sbapp->global, mbf->parent_subrangesz[bI]);
+//      // Copy from block-global to block local.
+//      gkyl_array_copy_range_to_range(sbapp->field->phi_smooth, sbapp->field->rho_c_global_smooth,
+//        &sbapp->local, &sbapp->field->global_sub_range);
+//    }
```
Antoinehoff, line 306: maybe this should be stored in a branch instead of being pushed in the main

manauref: many things in multiblock are works in progress, so I think it's ok to leave that there for now

apps/gyrokinetic.c
```diff
@@ -687,14 +733,34 @@ gkyl_gyrokinetic_app_apply_ic(gkyl_gyrokinetic_app* app, double t0)
 
         // Compute and store (in the ghost cell of of out) the boundary fluxes.
         // NOTE: this overwrites ghost cells that may be used for sourcing.
```
Antoinehoff, line 735: I have a question about that: does that mean that computing the bflux is incompatible with some other options? Is it like incompatible with the Maxwellian BC?

apps/gk_multib_field.c
Antoinehoff: ok

apps/gk_field.c
```diff
@@ -509,61 +548,147 @@ gk_field_calc_ambi_pot_sheath_vals(gkyl_gyrokinetic_app *app, struct gk_field *f
...
       }
       else {
-        gkyl_fem_parproj_set_rhs(field->fem_parproj, field->rho_c_global_dg, field->rho_c_global_dg);
-        gkyl_fem_parproj_solve(field->fem_parproj, field->rho_c_global_smooth);
+//        // This workflow solves Poisson on planes (doesn't conserve energy).
+//        gkyl_comm_array_allgather(app->comm, &app->local, &app->global, field->rho_c, field->rho_c_global_dg);
+//        gkyl_fem_parproj_set_rhs(field->fem_parproj, field->rho_c_global_dg, field->rho_c_global_dg);
+//        gkyl_fem_parproj_solve(field->fem_parproj, field->rho_c_global_smooth);
+//        gkyl_deflated_fem_poisson_advance(field->deflated_fem_poisson, field->rho_c_global_smooth,
+//          field->phi_bc, field->phi_smooth);
```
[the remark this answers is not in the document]
manauref: we still need to quickly switch between poisson on slabs and poisson on planes for multiblock testing, so I need to leave this here for now. Same reason I left the ability to quickly switch between the two in the multiblock app

apps/gyrokinetic.c
manauref: that's a very old comment, unrelated to this work

Antoinehoff: but I took this one as example... the comment:
`// NOTE: this overwrites ghost cells that may be used for sourcing.`
is present in a lot of places

manauref: yeah I think that's because a long time ago, in the early days of g0, we used to project some things into the ghost cells at t=0, and assumed they always stayed there, so this would've messed with this. But that's obsolete now.

apps/gyrokinetic.c
```diff
@@ -1380,8 +1488,11 @@ gyrokinetic_rhs(gkyl_gyrokinetic_app* app, double tcurr, double dt,
 
     // Compute and store (in the ghost cell of of out) the boundary fluxes.
     // NOTE: this overwrites ghost cells that may be used for sourcing.
-    if (app->field->update_field && app->field->gkfield_id == GKYL_GK_FIELD_BOLTZMANN)
-      gk_species_bflux_rhs(app, s, &s->bflux, fin[i], fout[i]);
+    gk_species_bflux_rhs(app, s, &s->bflux_solver, fin[i], fout[i], 0);
+
+    // Compute the boundary fluxes and their moments for diagnostics.
+    // NOTE: this overwrites ghost cells that may be used for sourcing.
```
Antoinehoff, line 1494: So linked to the comment above, I do not see the point of creating more obsolete lines of comments then. Why not use this PR to clean this as it modifies these locations?

manauref: I'm not creating obsolete lines. They were already there. They might've been moved. I can delete them if that's what you are asking

Antoinehoff: I think this is a copy paste here. Yes I think these comments are misleading, this PR is the good occasion to correct them.

manauref:
alright, will remove

Antoinehoff (approved):

JunoRavin (approved):

PR #601, opened 2025-02-13
Fix CX updater for CUDA

tnbernard:
The CX updater functions and methods of kernel selection are now optimized for CUDA to avoid memory leaks. Structure is now consistent with other cuda updaters in zero directory. Tests with CX are valgrind clean and results are consistent with main branch.

### DIFF
--- a/zero/dg_cx.c
+++ b/zero/dg_cx.c
@@ -55,7 +55,9 @@ gkyl_dg_cx_new(struct gkyl_dg_cx_inp *inp, bool use_gpu)
     up->b = 5.65e-20;
   }
 
-  up->react_rate = choose_kern(b_type, cdim, vdim_vl, poly_order);
+  int tblidx = cv_index[cdim].vdim[vdim_vl]
+  assert(tblidx != -1);
+  up->react_rate = choose_kern(b_type, tblidx, poly_order);
 
   up->flags = 0;
   GKYL_CLEAR_CU_ALLOC(up->flags);
--- a/zero/dg_cx.c
+++ b/zero/dg_cx.c
@@ -13,6 +13,11 @@
 gkyl_dg_cx*
 gkyl_dg_cx_new(struct gkyl_dg_cx_inp *inp, bool use_gpu)
 {
+#ifdef GKYL_HAVE_CUDA
+  if(use_gpu) {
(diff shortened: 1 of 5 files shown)


zero/dg_cx.c
```diff
@@ -95,7 +90,7 @@ void gkyl_dg_cx_coll(const struct gkyl_dg_cx *up,
 
     double *coef_cx_d = gkyl_array_fetch(coef_cx, loc);
     
-    double cflr = up->kernels->react_rate(up->a, up->b, up->vt_sq_ion_min, up->vt_sq_neut_min,
+    double cflr = up->react_rate(up->a, up->b, up->vt_sq_ion_min, up->vt_sq_neut_min,
```
manauref, line 93: can we delete the commented code below this?

zero/dg_cx_cu.cu
```diff
@@ -71,7 +48,63 @@ void gkyl_dg_cx_coll_cu(const struct gkyl_dg_cx *up,
...
+  }
+  else if (up->type_ion == GKYL_ION_HE) {
+    up->a = 6.484e-19;
+    up->b = 4.350e-20;
+  } 
+  else if (up->type_ion == GKYL_ION_NE) {
+    up->a = 7.95e-19;
+    up->b = 5.65e-20;
+  }
+
```
manauref, line 95: can you put these parameters in a place common to both the .c and .cu files (probably the private header) please? otherwise someone may come here to adjust one of these numbers but do it for CPUs but not GPUs, or viceversa.
Also, a comment on where these numbers came from would be good.

manauref (changes requested):

zero/dg_cx_cu.cu
tnbernard: Fixed in last two commits

JunoRavin:
@tnbernard this looks great. My only request is a small one about names. Since we use cu_dev_new in a bunch of other places in the code, let's also do cu_dev_new here (instead of cu_new). 

We may change these names uniformly at some point when we switch to HIP and add AMD GPU support, but it will make the search and replace easier if this name is also cu_dev_new.

JunoRavin (approved):

manauref (approved):

Issue #605, opened 2025-02-14
[DR] Global (in z) Normalization for Tokamak Geometry


akashukla:
### Problem to be Addressed
This DR suggests a change that only affects multi-bblock tokamak simulations. Originally, for multi-block simulations, each section of a z-connected region (for example the 3 SOL blocks or the 2 core blocks), had normalization factor on the z coordinate which was the poloidal length of the flux surface in that region. This resulted in a Jacobian that was discontinuous along z. This has 4 disadvantages:

1. Writing functions for initial conditions and sources is very  hard, because every block has different coordinates
2. command line postgkyl can't be used to plot quantities along the field line because the coordinate system changes along the field line
3. Because J is discontinuous along z, the procedure for the smoothing and poisson solve cannot be identical to the procedure used in the single block app. Using an identical procedure is preferable
4. It is not possible to create identical single and multi-block simulations for a 1:1 comparison.

### Proposed Solution
In this DR, we suggest using a normalization factor just like single-block simulations: the poloidal length of the whole flux surface from divertor plate to divertor plate (in the SOL) or the all the way around in the core.

In the SOL, previously, cuts between blocks were horizontal lines at the cylindrical Z value of the X-point. Now, the separatrix is traced, and the fraction of the poloidal arc length contained below, between, and above the X-points is calculated. Based on this, the computational grid extents are set. So, if  for example 30% of the poloidal length is below the lower X-point, 40% is between the X-points, and 30% is above the upper X-point on the separatrix, all other flux surfaces in the region will have this same of arc length split between blocks. For the core, if at the separatrix the split it 60% on the right half and 40% on the left half, then all flux surfaces in the core will share that split.

This gets rid of all 4 disadvantages listed above. This does not change the single block app or geometry  at all. An example grid using this method is shown below.

![Image](https://github.com/user-attachments/assets/e23c7717-edea-4dcd-9b46-22fc9cda57d4)

![Image](https://github.com/user-attachments/assets/44240e33-03ae-412f-8ec9-26047cec176d)

### Code Changes:
The branch `gk-g0-app_global_tokamak_norm_gkbc_enhance` contains the code for this change.

There are changes at the low level in zero to redefine how the normalization factor is set for multi-block block types.
At the app level, there is one additional function added to block geom
```
void
gkyl_block_geom_reset_block_extents(struct gkyl_block_geom *bgeom, int bidx, double *lower, double *upper)
{
[...]

PR #606, opened 2025-02-14
Global (in z) Tokamak geometry normalization

akashukla:
This PR is explained in DR #605.

Update in response to a comment by @JunoRavin : 

There are 3 changes to the low-level code in zero/tok_geo.c and zero/tok_geo_utils.c:
1. The find_endpoints method and associated low level tracing methods have been updated to reflect the new definition of the normalization factor.
2. There is a new  gkyl_tok_geo_set_extent method that just traces the separatrix and sets grid extents based on the fraction of 2pi occupied by each block of a z-connected region at the separatrix
3. Many if statements required to enforce consistency between z-connected blocks in the old geometry generation method are now unnecessary  and have been removed.

There is one change to the app level code:
1. gyrokinetic_multiub.c now calls gkyl_gk_geometry_tok_set_grid_extents to determine the z-extent of the grid for each block.


JunoRavin:
Two questions:
1. If I understand the DR, you are now saying with this procedure we can also have Jacobians that are continuous in radius too if you are matching the amount of poloidal projection on either side of the radial blocks? Or is the Jacobian still going to be discontinuous in radius? This is probably not a big deal, but the multi-block grid you posted in the DR looks very nice and I'm wondering what, if any, disadvantages this approach has (it seems like it has none relative to what we were thinking before?). 
2. Can you include some comments in the PR for the movement of code? It looks like identifying the end points has moved out of tok_geo.c and into the utils object? I am curious what the thought process for this refactoring is and how the code may have changed in this PR since there are a lot of deletions in tok_geo.c and a lot of additions in tok_geo_utils.c

akashukla:
> Two questions:
> 
> 1. If I understand the DR, you are now saying with this procedure we can also have Jacobians that are continuous in radius too if you are matching the amount of poloidal projection on either side of the radial blocks? Or is the Jacobian still going to be discontinuous in radius? This is probably not a big deal, but the multi-block grid you posted in the DR looks very nice and I'm wondering what, if any, disadvantages this approach has (it seems like it has none relative to what we were thinking before?).
> 2. Can you include some comments in the PR for the movement of code? It looks like identifying the end points has moved out of tok_geo.c and into the utils object? I am curious what the thought process for this refactoring is and how the code may have changed in this PR since there are a lot of deletions in tok_geo.c and a lot of additions in tok_geo_utils.c

1. The jacobian is going to be discontinuous in the radial coordinate. The normalization factor between core and SOL will be different. The jacobian will only be continuous along z-connected regions, but will have a jump due to the normalization factor across radial boundaries. @manauref and I are optimistic that the poisson solve we will be able to  handle this in the poisson solve. This discontinuity in radius of J is the only real disadvantage I see in this method of geometry generation.

2.  Yes I can do that. The endpoint function hasn't actually been moved, just modified to account for the different normalization factor definitions. A set_extents function has also been added to tok_geo_utils.c - this function is similar to find_endpoints but it actually sets the limit of the computational grid based on tracing the separatrix. This new function ends up being called at the app level.

All the deletions in tok_geo.c are related to another disadvantage of the old method: When the coordinate system changed along z (due to different normalization factors), one had to be very  careful that the endpoints of one z-connected block and another matched exactly. There were a lot of if statements in tok_geo.c that were used to enforce this. Now that the z-connected regions share a coordinate system, the endpoints automatically coincide and those if statements are unnecessary.

regression/rt_gk_multib_sheath_1x2v_p1.c
```diff
@@ -0,0 +1,852 @@
...
+   x  
+   ^  
+   |
+   1  +------------------+------------------+
+   |  |b0                |b1                |
+   |  |lower SOL         |upper             |
+   |  |                  |                  |
+   0  +------------------+------------------+
+      0 -----------------1------------------2----> z
+
```
manauref, line 428: this diagram only has 2 blocks, but this input files has 3. Update the comment/diagram if that's correct.

regression/rt_gk_multib_sheath_1x2v_p1.c
```diff
@@ -0,0 +1,852 @@
...
+  // Initial and final simulation times.
+  int frame_curr = 0;
+  double t_curr = 0.0, t_end = ctx.t_end;
+  // Initialize simulation.
+  if (app_args.is_restart) {
+    struct gkyl_app_restart_status status = gkyl_gyrokinetic_multib_app_read_from_frame(app, app_args.restart_frame);
+
+    if (status.io_status != GKYL_ARRAY_RIO_SUCCESS) {
+      gkyl_gyrokinetic_multib_app_cout(app, stderr, "*** Failed to read restart file! (%s)\n",
+        gkyl_array_rio_status_msg(status.io_status));
```
manauref, line 748: fix indent

manauref (changes requested):
Can you check that this branch is valgrind clean please?

also, aren't the z-extents in the MB input files meaningless now? should we add a comment about it in the input files or just make them absurd values (e.g. like 1e10 and mention in the comment that they are set under the hood?

manauref (approved):

JunoRavin (approved):

Issue #604, opened 2025-02-14
[DR] Boundary flux diagnostics API


manauref:
As part of the work in DR #593 and PR #599 , we implemented boundary flux integrated diagnostics (well, 1 integrated diagnostic to be precise, the integrated Hamiltonian moments). I wasn't totally happy with the public API for those and I'd like to redo it as described here.

First consider the species integrated moments. Presently the GK solver outputs integrated BiMaxwellian moments of each by default. But to perform particle/momentum/energy balance calculations the Hamiltonian moments (M0, m*M1, H) are needed. So instead of tacitly allowing the code always and only output integrated BiMaxwellian moments, I think the public API (in the species table of the input file) should be:
```
    .num_integrated_diag_moments = 1,
    .integrated_diag_moments = { "BiMaxwellianMoments" },
```
analogous to how diag_moments are specified. This more naturally allows us to use
```
    .num_integrated_diag_moments = 1,
    .integrated_diag_moments = { "Hamiltonian" },
```
as needed for energy balance calcs.

Now for boundary flux diagnostics I propose that we use
```
    .boundary_flux_diagnostics = {
      .num_diag_moments = #,
      .diag_moments = {...},
      .num_integrated_diag_moments = 1,
      .integrated_diag_moments = { "Hamiltonian" },
    }
```
analogous to how the speces diag moms and integrated diag moms are specified. This will also allow us to request diag moments of the boundary flux (e.g. to compute EXACTLY what the particle, momentum, energy flux profile to the boundary is).

There are three other reasons I like this bflux diag API:
1. When one wants to perform particle/momentum/energy balance calcs by looking a total content instead of rates of change, we need the time integrated boundary flux moments (or int moments). So for this type of analysis we'd like to pass an extra flag requesting time integrated int moms, and rather than polluting the species table, we can use the consolidation of bflux diag inputs to request time integration as
```
    .boundary_flux_diagnostics = {
      ...
      .time_integrated = true,
    }
```
2. In clopen simulations where the limiter sticks into our domain we'll want the boundary flux diags along $z$ to be computed over a restricted radial range only, and at the input file level the consolidation of bflux_diag inputs into a struct will also allow us to specify the LCFS
```
    .boundary_flux_diagnostics = {
      ...
      .xLCFS = 0.04,
    }
```
so that the bflux diags are not computed in the core (assuming that whatever comes out one end enters into the other in the core).
[...]

PR #608, opened 2025-02-16
Boundary flux diagnostic API

manauref:
This is the implementation of DR #604 , merging into the gk-g0-app_energy_conservation-production branch.

### DIFF
--- a/apps/gk_species_bflux.c
+++ b/apps/gk_species_bflux.c
@@ -244,7 +244,7 @@ gk_species_bflux_calc_integrated_mom_dynamic(gkyl_gyrokinetic_app* app,
 }
 
 static void
-gk_species_bflux_calc_integrated_mom_time_integrated_dynamic(gkyl_gyrokinetic_app* app,
+gk_species_bflux_calc_voltime_integrated_mom_dynamic(gkyl_gyrokinetic_app* app,
   const struct gk_species *gk_s, struct gk_boundary_fluxes *bflux, double tm)
 {
   int vdim = app->vdim;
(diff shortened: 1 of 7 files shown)


apps/gk_species_bflux.c
```diff
@@ -237,11 +237,24 @@ gk_species_bflux_calc_boundary_flux_integrated_mom_dynamic(gkyl_gyrokinetic_app*
...
       gkyl_dynvec_append(bflux->intmom[2*d+e], tm, avals_global);
+    }
+  } 
+}
 
-      // Cummulative (in time) integrated moment of the boundary flux.
-      // Here we assume the final bflux (not divided by dt) is in f1.
-      gkyl_array_integrate_advance(bflux->integ_op, bflux->f1[2*d+e], 1., 0,
+static void
+gk_species_bflux_calc_boundary_flux_integrated_mom_time_integrated_dynamic(gkyl_gyrokinetic_app* app,
```
JunoRavin, line 247: I love a long function name, but these function names are a lot. Why are we including "calc_boundary_flux" in the name when these are already bflux functions? Why did you abandon the naming convention you seemed to have had with cumm or cumulative in favor or time_integrated (integrated_mom_time_integrated is a particular mouthful when it seemed like you had a good name when discussing the cumulative integrated moments).

apps/gk_species_bflux.c
```diff
@@ -356,12 +392,19 @@ gk_species_bflux_init(struct gkyl_gyrokinetic_app *app, struct gk_species *gk_s,
     bflux->bflux_combine_range_func = gk_species_bflux_combine_range_dynamic;
     bflux->bflux_copy_range_func = gk_species_bflux_copy_range_dynamic;
     bflux->bflux_rhs_func = gk_species_bflux_rhs_diag; 
-    bflux->calc_boundary_flux_integrated_mom_func = gk_species_bflux_calc_boundary_flux_integrated_mom_dynamic;
+    if (gk_s->info.boundary_flux_diagnostics.time_integrated) {
+      bflux->calc_boundary_flux_integrated_mom_func = gk_species_bflux_append_boundary_flux_integrated_mom;
+      bflux->calc_bflux_int_mom_time_integrate_func = gk_species_bflux_calc_boundary_flux_integrated_mom_time_integrated_dynamic;
+    }
+    else
```
JunoRavin, line 399: If the if statement has braces, the else should also have braces, even if it's just one line.

regression/rt_gk_step_out_2x2v_p1.c
```diff
@@ -313,12 +313,12 @@ create_ctx(void)
 void
 calc_integrated_diagnostics(struct gkyl_tm_trigger* iot, gkyl_gyrokinetic_app* app, double t_curr, double dt, bool force_calc)
 {
-  if (gkyl_tm_trigger_check_and_bump(iot, t_curr) || force_calc) {
+//  if (gkyl_tm_trigger_check_and_bump(iot, t_curr) || force_calc) {
```
JunoRavin, line 316: Did you mean to leave these lines commented in this PR?

JunoRavin:
I like the function pointer construction for bflux here. This structure is quite nice. My only question is related to the single comment about some function names. Some of them are a lot and in various places bflux and boundary_flux are used, or gk_species_bflux_calc_boundary_flux, and I'm wondering if we can streamline some of these names.

apps/gk_species_bflux.c
manauref: hmm true we could remove the "boundary_flux" part of the name.

Re cummulative: well because the word cummulative by itself doesn't tell us that it is time integrated (i.e. cummulative in what sense?) True that the name integrated_mom_time_integrated is a mouthful, but I didn't have any better options.

regression/rt_gk_step_out_2x2v_p1.c
manauref: don't worry about the reg tests, there's all sorts of changes and hacks in them that I will clean up before merging energy_conservation into main

manauref:
> I like the function pointer construction for bflux here. This structure is quite nice. Also the input file struct is good and I approve of that design too.
> 
> My only question is related to the single comment about some function names. Some of them are a lot and in various places bflux and boundary_flux are used, or gk_species_bflux_calc_boundary_flux, and I'm wondering if we can streamline some of these names.

will think about names

Issue #609, opened 2025-02-17
Refactored vlasov_fluid_species and new equations utilizing existing canonical Poisson bracket infrastructure.


JunoRavin:
This design review encompasses a refactored vlasov_fluid_species struct, following the successful refactor of other parts of the app in https://github.com/ammarhakim/gkylzero/issues/450 and https://github.com/ammarhakim/gkylzero/issues/581 to set function pointers in the init method for these structs and simplify how the system switches between different models, and extends vlasov_fluid_species to solve new fluid equation systems utilizing the canonical Poisson bracket infrastructure spearheaded by @johnson452. Work for this branch can be found in https://github.com/ammarhakim/gkylzero/tree/gk-g0-app-can-pb-fluid

In particular, we now support canonical Poisson bracket systems which @ammarhakim previously demo-ed all the way back in Gkeyll 1.0, including incompressible Euler (https://ammar-hakim.org/sj/je/je13/je13-incomp-euler-2d.html) and Hasegawa-Mima, Hasegawa-Wakatani, and the modified variant of Hasegawa-Wakatani which subtracts off the zonal components of the potential and density (https://ammar-hakim.org/sj/je/je17/je17-hasegawa-wakatani.html), the latter of which utilizes the work implemented in https://github.com/ammarhakim/gkylzero/issues/557 by @Antoinehoff to compute the zonal components of the potential and density. 

All of these extensions are transparently handled within the app by setting a few different function pointers, such as one for what we call the "primitive variables" (or more generally the auxiliary variables), which in the case of Euler computes the flow velocity, u, and pressure, p, but for the canonical Poisson bracket systems solves a Poisson equation from, e.g., the vorticity in incompressible Euler. We also have function pointers for the integrated quantities (which for example include the energy and enstrophy for the canonical Poisson bracket systems as described in https://ammar-hakim.org/sj/je/je13/je13-incomp-euler-2d.html, or the integrated mass, momentum, kinetic energy, and internal energy for Euler's equations), a write method which can write out the potential for canonical Poisson bracket systems and primitive variables for Euler's equations, and a separate release method for releasing equation specific information. 

So for each equation we have a specific initialization routine:
<code> // Setup equation-specific memory and equation type/number of equations based on input table
  f->eqn_type = f->info.equation->type;
  f->num_equations = f->info.equation->num_equations;
  f->equation = gkyl_wv_eqn_acquire(f->info.equation);
[...]

Issue #485, opened 2024-09-29
DG fluid regression test seg faults


manauref:
In gk-g0-app the regression test `rt_advect_1x.c` seg faults. I think it is because `gkyl_wv_advect_new` sets
```
  advect->eqn.type = GKYL_EQN_MAXWELL;
```
but when the flow goes into `gkyl_dg_updater_fluid_new` and sees this chunk
```
  up->eqn_id = wv_eqn->type;
  if (up->eqn_id == GKYL_EQN_ADVECTION) {
    up->eqn_fluid = gkyl_dg_advection_new(cbasis, conf_range, up->use_gpu);
    struct gkyl_dg_advection_auxfields *adv_in = aux_inp;
    gkyl_advection_set_auxfields(up->eqn_fluid, *adv_in);
  }
  else if (up->eqn_id == GKYL_EQN_EULER) {
    up->eqn_fluid = gkyl_dg_euler_new(cbasis, conf_range, wv_eqn, geom, up->use_gpu);
    struct gkyl_dg_euler_auxfields *euler_inp = aux_inp;
    gkyl_euler_set_auxfields(up->eqn_fluid, *euler_inp);
  }
```
the equation object is not initialized because there's no option for `GKYL_EQN_MAXWELL`. Then it tries to create a hyper_dg without an equation object and it seg faults.

JunoRavin:
Bug fixed in https://github.com/ammarhakim/gkylzero/pull/610

PR #610, opened 2025-02-17
Refactored vlasov_fluid_species and new equations utilizing existing canonical Poisson bracket infrastructure

JunoRavin:
This PR goes with DR https://github.com/ammarhakim/gkylzero/issues/609

We have refactored the vlasov_fluid_species as discussed in this DR to make more extensive use of function pointers to cleanly handle the different computations performed by different fluid species we might simulate (including Euler's equations, Advections equations, and the new capabilities to simulate systems described by a Poisson bracket such as incompressible Euler, Hasegawa-Mima, Hasegawa-Wakatani, and the modified Hasegawa-Wakatani system). 

Utilizing the same upwind procedure as the current kinetic canonical Poisson bracket formulation, fluid systems such as incompressible Euler conserve energy and decay the enstrophy monotonically. See attached vortex-waltz with 64^2, p=2 (integrated quantities are vorticity, enstrophy, and energy). All test cases are based on simulations on @ammarhakim's website, including the incompressible Euler tests (https://ammar-hakim.org/sj/je/je13/je13-incomp-euler-2d.html) and Hasegawa-Wakatani tests (https://ammar-hakim.org/sj/je/je17/je17-hasegawa-wakatani.html). All input files are driven via Lua and can be found in regression/lua. 

Some notes: 
1. The Poisson solver is the dominant cost, taking 90+ percent of the time on resolutions up to 256^2. 
2. It is highly recommended that you utilize @manauref's work on cuDSS if you want to run these simulations on GPUs, as cuDSS is upwards of 2-3 times faster than the cuSparse implementation which does not work on cuda 12.1+ (also cuda 12.4 compiles faster, so that's nice). 
3. There is currently some limitation to the fem_poisson solver that if you try to do these simulations on 512^2 grids with p=2, the constant offset factor from applying periodic boundary conditions becomes very large (normally it's O(1), but it becomes much larger, 10^10 in incompressible Euler), causing Hasegawa-Wakatani simulations to crash. Obviously that is a highly resolved grid for any GK simulation, so we might not have been affected by such an issue before, but it's worth noting in case anyone tries to do highly resolved Poisson calculations. 
4. I am currently limiting these fluid systems to running with periodic boundary conditions since the Poisson solve is handled internally to the fluid species. Eventually we will likely want to utilize the Poisson solve infrastructure being used in Vlasov-Poisson calculations, but for these fluid test cases where there are only single fluid systems, I thought this was fine to make this self-contained this way to start. 
5. This PR also addresses issue https://github.com/ammarhakim/gkylzero/issues/485 which was not the only reason the advection equations were failing, but was a big one. Now the advection tests run through C, and with the extension of the Lua wrappers, Lua, and produce the expected behavior (in fact the behavior is better due to realizing that the GAUSS_LOBATTO specification does not actually make a continuous function, and one needs to use eval_on_nodes to get a continuous advection velocity). 

Screenshot 2025-02-16 at 8 52 30 PM
Screenshot 2025-02-16 at 8 52 56 PM


johnson452 (approved):

JunoRavin:
With the fix to array_reduce sum (which is utilized by fem_poisson in the case of doubly periodic domains to subtract off the average RHS value) by first clearing the double on device before entering the kernel, we can now do high resolution (512^2 and bigger) simulations on GPUs. Here is the incompressible Euler vortex-waltz. 

This bug in array_reduce sum also fixes other issues anywhere this routine was being utilized. In effect, if your grid was big enough and you were utilizing array_reduce sum, there was a race condition between the blockReduce CUB call and the zero-ing out of the array (not sure why this was grid dependent except maybe some limitation of how CUB handles the blockReduce; if your grid is bigger than 2048x32 = 256^2 in terms of thread count, which is I think the limit on how many blocks are in a grid, then maybe two different grids were touching out at the same time even as the blockReduce started its work). 

Anyway, fixed now thank goodness and this merge is ready. 
Screenshot 2025-02-27 at 10 16 33 AM

PR #618, opened 2025-02-21
Make the step regression test more reasonable. 

akashukla:
1) move plates to reasonable location 2) move the domain out from the separatrix 3) increase the density floor on the ICs and 4 Remove Argon. Otherwise the STEP test was not stable for more than 1 microsecond.

During conservation testing, @manauref noticed that this test was poorly behaved when run to times longer than the old rather short end time in the regression test (5e-7). These changes fix this so the STEP test is now more realistic and stable.

### DIFF
--- a/regression/rt_gk_step_out_2x2v_p1.c
+++ b/regression/rt_gk_step_out_2x2v_p1.c
@@ -67,8 +67,8 @@ eval_density(double t, const double * GKYL_RESTRICT xn, double* GKYL_RESTRICT fo
   double cz = app->cz;
   double xcenter = 1.2014;
   double n = n0*exp(-(x-xcenter)*(x-xcenter)/2/cx/cx) * exp(-z*z/2/cz/cz);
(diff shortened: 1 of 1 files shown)


manauref:
absorb instead of zero-flux BCs along x?

manauref:
Here are the particle balance for this new version of the test:

Screenshot 2025-02-21 at 10 16 57 AM

Screenshot 2025-02-21 at 10 16 06 AM

and here's the energy balance

Screenshot 2025-02-21 at 10 19 55 AM

Screenshot 2025-02-21 at 10 22 54 AM

and yes, changing cfl_frac from 0.5 to 0.25 didn't change the late error at all, I checked twice

akashukla:
> absorb instead of zero-flux BCs along x?

This is not a super important change, but  this is how production simulations were run. I would be fine with reverting this if you prefer zero flux

manauref (approved):

JunoRavin (approved):

Issue #607, opened 2025-02-16
GR Free-streaming Kinetics with Can-pb


johnson452:
This design review encompasses the code requirements needed to extend the canonical bracket solver to handle free-streaming kinetics in static spacetimes. The bracket solver will remain largely un-modified. Only the input Hamiltonian, moments, and boundary conditions require extensions or modification. 

The main changes will be:
- Adding in a new gkyl_model_id named GKYL_MODEL_CANONICAL_PB_GR.
- Modifying Can-pb Moments: A replacement set of moments for M0/ M1i as worked out in the attached document. These moments are valid for non-rel. can-pb, SR can-pb and GR can-pb. Therefore, the M1i moment will replace the M1i moment for all of can-pb. (This is expected to very slightly change the results of all simulations using can-pb moments like unit tests, @tnbernard.) Essentially we are switching from using covariant to contravariant momentum: nu_i = integral ( p_i f dp ) to nu^i = integral ( \alpha^i f dp) where \alpha is the characteristic velocity.
- Modifying the temperature moment for the nonrelativistic case to be in terms of nu^i.  JndT = 2*E - (Jnu^i) (Jnu^j)h_{ij}/(Jn).
- Adding in covariant metric components h_ij to compute the temperature.
- Restrict the moments so only M0, M1i moments will be initially available and only for diagnostic purposes after this DR/PR. (More to come in a future DR/PR) 
- Adding Gauge variables to can-pb: Linking in the GR lapse, shift and spatial metric and projecting these onto the spatial DG basis. 
- Add fields for alpha and beta to the can-pb union of data, so it’s treated the same to how h_ij_inv and det_h are currently.
- Add Boundary Conditions: Additional option for the the GR-kinetic-can-pb for excision boundary conditions. 

Tests: Compare with analytic solutions: Geodesics.
- Chandrashekar (chapter 3) Orbits of the Schwarzschild blackhole.
- Chandrashekar (chapter 7) Orbits of the Kerr blackhole.

This PR will not modify f_lte and will require more theory development to correctly develop a local equilibrium. This will come in a separate DR once fully worked out.

See the attached [work-in-progress] notes for the documented equations (Particularly subsection 2.2 summarizes the neutral evolution, and section 3 the moments).

[WIP_Notes_GR_Kinetics.pdf](https://github.com/user-attachments/files/18813419/WIP_Notes_GR_Kinetics.pdf)

johnson452:
[...]

PR #623, opened 2025-02-23
Can pb momentum moment from hamil

johnson452:
This branch addresses the first component of [DR#607](https://github.com/ammarhakim/gkylzero/issues/607) and adds a new moment to can-pb, M1i_from_H. This new moment uses the discretely projected Hamiltonian and the resulting equations of motion in the spatial directions to compute the bulk momentum. 

This branch is varified to:
[X] Build on GPU
[X] Run through lua
[X] Valgrind clean
[X] Passes unit tests

Summary:
- LTE moments is unaffected, as well as pressure, which will continue to use entirely covariant components. This branch compute the contravariant momentum under the hood, and converts the momentum to the covariant components via the spatial metric. This requires us to have a new geometric variable, h_ij, (the covariant components of the spatial metric) input as an extra geometric quantity to use the LTE moments/f_lte/bgk collisions.
- Changing the momentum calculation has largely left p2 velocity space calculations unchanged, and primarily affects p1 momentum space simulations. (See comments on [DR#607](https://github.com/ammarhakim/gkylzero/issues/607))
- Fixed a bug in p2 tensor pressure moments to now call the correct dg_binops

Additional notes:
- There are no guarantees that the discreetly projected h_ij and h^{ij} when multiplied obey: h_{ij}h^{jk} = \detla_i^k. A potentially addressing of this is handle these quantities at nodal points rather than convert via dg_binop_mul.
- This change is largely intended to lay the foundation for moments needed in GR.


JunoRavin (approved):
This looks good to me, though we should start to think about some naming conventions going forward. I think at this point if we have to add more ORs to our checks for the type of system we're doing, we're doing something wrong (and arguably we're doing something inefficient now). 

You, me, @JonathanGorard, and @ammarhakim will probably want to do a bigger design review on once and for all folding in the Hamiltonian infrastructure to Vlasov so that every Vlasov object expects a Hamiltonian and we don't have to distinguish between canonical PB neutrals and Vlasov neutrals (and ways internally to distinguish the different overall sparsity of the Hamiltonian, whether it's something like 1/2 p^2, sqrt(1 + p^2) (for flat spacetime relativity), or 1/2 g^ij p_i p_j).

Further, having a more generic moment type organization for "these moment types come from a Hamiltonian" and "these moment types come from the grid" is the right dichotomy, and then generically there is a pressure computation of some kind depending on the system, since right now the M1i_from_H moment is just the same thing as the M1i moment in vlasov_sr, and the MEnergy moment is the same as the M2 moment in vlasov_sr.

ammarhakim (approved):

Issue #622, opened 2025-02-23
[DR] Refactor code to prep for modularization into sub-directories.


manauref:
These are sets of tasks/design notes on refactoring the code so we can modularize into sub-dirs. As we work on directory reorg, see #562, we are discovering some improperly arranged code. We need to refactor so ensure our modularization efforts yield a sensible split of code into various modules.

Right now zero/array_ops has reductions, and reductions that are DG-aware (and thus depend on basis). Also, `unit/ctest_array.c` tests reductions and array_ops. We propose to split these things a bit so there's less overlap between objects/modules. Specifically:

1. Remove reductions and DG-aware reductions from `zero/array.c`. Move reductions to `zero/array_reduce.c` (we already have `zero/array_reduce_cu.cu`) and DG-aware reductions to `zero/array_dg_reduce.c`.
2. Move the reduction tests from `unit/ctest_array.c` into `unit/ctest_array_reduce.c` and `unit/ctest_array_dg_reduce.c`.
3. Move the array_ops tests from `unit/ctest_array.c` into a new `unit/ctest_array_ops.c`.

ammarhakim:
There is also a bug in the array IO code. A serious one at that! We have:

```
uint64_t file_type_u64;
frr = fread(&file_type, sizeof(uint64_t), 1, fp);
```

(See https://github.com/ammarhakim/gkylzero/blob/main/zero/array_rio.c Line 105-106). This needs to be replaced by:

```
    uint64_t file_type_u64;
    frr = fread(&file_type_u64, sizeof(uint64_t), 1, fp);
```

This error is insidious: seems that the unit test for this is "passing" with clang but not with gcc.

ammarhakim:
Why is gkyl_eqn_type.h being included in gkyl_proj_on_basis.h? All we need is the quad type from that file. I recommend we split the header up into modular bits, splitting the various eqn-specific enums in into eqn-specific headers 

Otherwise we will not be able to properly modularize the system. How can we have a moments-only app when the eqn_type header has info for all equation?

(One could argue this is fine, though. We need to think more carefully. I err on side of splitting into multiple headers)

manauref:
> Why is gkyl_eqn_type.h being included in gkyl_proj_on_basis.h? All we need is the quad type from that file. I recommend we split the header up into modular bits, splitting the various eqn-specific enums in into eqn-specific headers
> 
[...]

PR #624, opened 2025-02-24
Reorganize array ops, array reduce and their unit tests.

manauref:
See DR #622 

We move reductions out of array_ops and into array_reduce, and we move the DG-aware reductions into array_dg_reduce. Made corresponding changes in unit tests.

Also fixed a bug in array_rio and a bug in gkyl_array_clone.

Unit tests pass on CPU and GPU.


JunoRavin:
Do we use gkyl_array_clone anywhere in the code? I think we always do things with acquires and releases instead of creating quasi-deep copies of arrays. Still awesome to fix bugs.

ammarhakim (approved):

PR #626, opened 2025-02-25
The gkyl_gk struct has become too HUGE for storing on the stack!!! So

ammarhakim:
The gkyl_gk struct has become too **HUGE** for storing on the stack!!! So we can no longer have C regressions the way we used to. We need to allocat the gkyl_gk struct on the heap, and set it manually. This sucks, but we must do this from now on. I have fixed one regression test regression/rt_gk_d3d_iwl_2x2v_p1.c and fixed a memory leak. This test is now valgrind clean.

@manauref @manauref @akashukla please go over EVERY SINGLE GK test/file and make the same fixes. This is a serious bug. More below.

Basically, the gkyl_gk struct is too big to keep on the stack. **This is a silent failure**. Some sims may seem to run but may randomly fail. Hence, it is very bad as this will happen at random. So we must fix every single input file in the manner I have indicated.

### DIFF
--- a/regression/rt_gk_d3d_iwl_2x2v_p1.c
+++ b/regression/rt_gk_d3d_iwl_2x2v_p1.c
@@ -9,6 +9,7 @@
 #include <gkyl_fem_poisson_bctype.h>
 #include <gkyl_gyrokinetic.h>
 #include <gkyl_math.h>
+#include <gkyl_alloc.h>
 
 #include <rt_arg_parse.h>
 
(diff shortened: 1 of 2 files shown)


JunoRavin (approved):

regression/rt_gk_d3d_iwl_2x2v_p1.c
```diff
@@ -9,6 +9,7 @@
 #include <gkyl_fem_poisson_bctype.h>
 #include <gkyl_gyrokinetic.h>
 #include <gkyl_math.h>
+#include <gkyl_alloc.h>
```
tnbernard, line 12: Looks like this appears twice. Here and at line 5.

manauref:
PR #712 should help alleviate this

manauref:
PR 712 should've alleviated this. We can close this PR until it pops up again.

PR #627, opened 2025-02-25
Fixing bflux seg fault in static species after new boundary flux features were added

tnbernard:
The bflux integrated moment function for static species wasn't initialized leading to a seg fault when the following was called in gyrokinetic.c: 
```
void
gkyl_gyrokinetic_app_calc_species_boundary_flux_integrated_mom(gkyl_gyrokinetic_app* app, int sidx, double tm)
{
  struct gk_species *gks = &app->species[sidx];
  gk_species_bflux_calc_integrated_mom(app, gks, &gks->bflux_diag, tm);
}
```
Now fixed and rt_gk_static test runs.

### DIFF
--- a/apps/gk_species.c
+++ b/apps/gk_species.c
@@ -1069,7 +1069,6 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
   }
 
   // Create boundary flux diagnostics if requested.
-  gks->bflux_diag = (struct gk_boundary_fluxes) { };
   gk_species_bflux_init(app, gks, &gks->bflux_diag, true);
 
   if (app->enforce_positivity) {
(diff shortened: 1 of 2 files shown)


JunoRavin (approved):

manauref (changes requested):
If I'm reading this correctly dynamic species will call gk_species_bflux_init twice, so if you check a reg test with blux diagnostics I bet it is not valgrind clean.

Also, please move the comment in line 1071 also.

JunoRavin:
Is there a reason we cannot just set the bflux diagnostic init to be true? Or is this a statement that the integrated moment calculator is not correctly setting the appropriate function pointer when running with a static species?

manauref:
@tnbernard you also need to move the release of bflux_diag.

PR #629, opened 2025-02-27
Corrected kernels for Hamiltonian moment and other fixes

manauref:
see commit message. This also goes with [PR 65 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/65)


manauref:
the actual phase range over which the updater operates is the one passed in
the _advanced method. These are only used for computing the linear index
given a multidimensional index that comes from the update range (the latter
is usually local), and such operation works equally with the local or
local_ext, as long as local is a sub of local_ext. I made this change
though because I wanted to index some conf-space fields in the ghost cell
in computing boundary fluxes. I made a similar change for moments in the
energy PR, where the creation of the updater takes the extended range, but
the _advance takes whatever update range is desired (local for most things,
the ghost range for boundary flux moments).

An example: this phase range passed in _new becomes eqn->phase_range. And
this range is used to index the phase space advection speeds, which we also
define in the upper ghost cell. So my understanding is that to properly
index alpha at the upper ghost cell we should pass local_ext. So I don't
understand how it worked before, but it worked. It'll also work as it is
here.

JunoRavin:
It worked before because local is a subrange of local_ext. It's the same reason you can index the ghost cells even though the update range is only a local range (because in hyper_dg the idxL and idxR are the increment of the idx in the desired direction, which pushes you into the extended range if you are at a skin cell). 

So this indexing here in the surface update: https://github.com/ammarhakim/gkylzero/blob/main/zero/gkyl_dg_gyrokinetic_priv.h#L607

was fetching the surface alpha stored in the ghost cell in the same way that pidxR indexes the distribution function in that ghost cell. 

I don't think this is right that these should be extended ranges. They should be local ranges and you can always index into the ghost cells with the needed relative shifts. Otherwise you may run into a situation where you apply a relative shift to an extended range and you will access memory out of bounds.

manauref:
If we are going to index into ghost cells we should use the extended
range, otherwise it is confusing or misleading (or harder to understand).
If you feel strongly about the opposite I can change it back.

JunoRavin:
It depends on what we're doing. If you mean that in some parts of boundary fluxes, you are looping over the ghost range and then accessing the surface characteristics in the skin cell by indexing into the domain from the extended range (basically shifting to the right if you are at the far left ghost cells and vice versa), but I'm not sure this is the right way to think about it anyway? 

But we index into the ghost cells all the time looping over the local range, and this is by design. We don't want to pass the extended range to things like hyper_dg. We want the relative index to just handle that you need stuff from the ghost cells. And because these same indices are passed to the surf function pointers (the general function signature for a surface update requires idxL, idxC, and idxR), we can always access information such as the surface characteristics in the ghost cells in the same way we access the distribution function in the ghost cells.

Antoinehoff:
I see both of your points. IMO, a typical finite difference scheme should be called on a loop that is local and the ghost would be accessed relatively by naively writing something like
```
loop(local); 
   FD(f[i-1],f[i],f[i+1]);
```
so in that sense I think it's better to keep the local ranges to be sure that the `local` in the loop does not become `local_ext`. I guess this is Jimmy's main arguments but maybe I did not understand the other arguments here.

manauref:
I hold that if we have a grid from 0 to 11, and the local range is [1,10] but we want to access the idx=11 cell, we should use the extended range. But if the original design Ammar had for ranges, extended ranges and sub ranges at the beginning of gkylzero does not agree with this I can change it back.Again, hyper_dg uses the local range passed _advance method, so I see no problem with passing an extended range to _new.On Feb 27, 2025, at 8:05 AM, Antoine Hoffmann ***@***.***> wrote:﻿
  I see both of your points. IMO, a typical finite difference scheme should be called on a loop that is local and the ghost would be accessed relatively by naively writing something like
loop(local); 
   FD(f[i-1],f[i],f[i+1]);

so in that sense I think it's better to keep the local ranges to be sure that the local in the loop does not become local_ext. I guess this is Jimmy's main arguments but maybe I did not understand the other arguments here.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



Antoinehoff left a comment (ammarhakim/gkylzero#629)
I see both of your points. IMO, a typical finite difference scheme should be called on a loop that is local and the ghost would be accessed relatively by naively writing something like
loop(local); 
   FD(f[i-1],f[i],f[i+1]);

so in that sense I think it's better to keep the local ranges to be sure that the local in the loop does not become local_ext. I guess this is Jimmy's main arguments but maybe I did not understand the other arguments here.

ammarhakim:
In general I don't want range mods in updaters to be done with deeper discussion. The design is deliberate and I need to understand why we are changing things like this. I know one wants to index in ghost cells. But that may need more thought to do properly.

I agree with Antoine and Jimmy for now. But we should think carefully and submit a DR for looping over ghost cells for diagnostics.

manauref:
Ok i'll switch it back.

JunoRavin (approved):

Antoinehoff (approved):

PR #630, opened 2025-02-27
Write out an error message if eqdsk file is not found

manauref:


ammarhakim:
Don't use printf in the code. Please use log so we don't write 1000s of messages in parallel.

ammarhakim:
Please also check if you can use the 'with_file' macro for opening files. Don't fopen directly. There are other functions to check if files exist. Those are safer to use. Please look at utils header.

manauref:
I switched it to fprintf.

Using `with file` will require a more careful examination of the code, because the file pointer is used over a big chunk of code and I don't know what's local (to this function) or not. So we'll leave that for later.

ammarhakim:
Also you did not address the issue of what happens in parallel. Plus as I pointed out you need to use the proper method to check for file existence that are in the utils.

ammarhakim:
One more point related to the above. The return value of fopen can't be used to check if file exists. It only tells you that file can't be opened, even if it exists (for example due to permissions). So the correct method to use is the one I pointed above.

manauref:
addressed in PR #787

PR #635, opened 2025-03-05
Bug fix: Boltzmann electron sheath values were half

Maxwell-Rosen:
# Bug fix

## Summary

Description: In calculating the sheath values for the Boltzmann electron model, there is an apparent average that the app does. The code reads like this. This is exclusive to 1x.

First, sheath_vals[0] calculates the density and sheath potential on the left edge. Then, sheath_vals[1] calculates density and sheath potential on the right edge. The rest of the arrays are zero. We want one array with all these quantities, so they are accumulated (out = out + 1 * inp).

```
gkyl_array_accumulate(field->sheath_vals[0], 1., field->sheath_vals[1]);
gkyl_array_scale(field->sheath_vals[0], 0.5);
```

If all the sheath values were 1, the arrays would look like this:
sheath_vals[0] = {1,0,0, ... , 0,0,0}
sheath_vals[1] = {0,0,0, ... , 0,0,1}

We want a result:
sheath_vals[0] = {1,0,0, ..., 0,0,1}

This scale should not be here because it would effectively average the sheath_vals with zero. This results in a 1/2 factor in the ion density at the sheath and the sheath potential, which should not be there.

Issue link: No issue opened

## Solution

The bug is fixed by removing the gkyl_array_scale. 

This was found through writing a unit test. There was no unit test for the functions in `zero/ambi_bolt_potential.c`. All files in `/zero` should have unit tests.

A unit test is added to ensure we understand how these functions behave. The unit test copies the features in the app directly with a known analytic solution.

Impacted files: app/gk_field.c

## Community Standards

- [ ] I added a regression test to test for this bug.
- [x] I added a unit test to test this bug.

## Additional Notes

This is a part of an inclusion of 2x Boltzmann electrons into Gkeyll. I wanted to PR this bug fix separately to make the PRs easy to address. There is some renaming of variables as well, and refactoring is required. It does not impact anything relevant to the 1x work. I want to add 2x unit tests before I PR that side of the project. I think what I implemented here regarding that should mean this works in 2x, but it's not tested. This is a separate bug-related PR and does not mean to be a 2x/3x Boltzmann electrons PR

Here, I run rt_gk_mirror_boltz_elc_1x2v_p1.c before and after the bug. It's interesting that the sheath potential is not doubled. Perhaps this interacts with the density.

image

### DIFF
--- a/apps/gk_field.c
+++ b/apps/gk_field.c
@@ -223,7 +223,7 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
     f->ambi_pot = gkyl_ambi_bolt_potential_new(&app->grid, &app->basis, 
       f->info.electron_mass, f->info.electron_charge, f->info.electron_temp, app->use_gpu);
     for (int j=0; j<app->cdim; ++j) {
-      f->sheath_vals[2*j] = mkarr(app->use_gpu, 2*app->basis.num_basis, app->local_ext.volume);
+      f->sheath_vals[2*j]   = mkarr(app->use_gpu, 2*app->basis.num_basis, app->local_ext.volume);
       f->sheath_vals[2*j+1] = mkarr(app->use_gpu, 2*app->basis.num_basis, app->local_ext.volume);
     }
   } else {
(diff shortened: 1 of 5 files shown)


Maxwell-Rosen:
I am really not sure what relevant numbers are necessary. I can print out the sheath potential before and after it gets halved, but it will just show that the array_scale is halving the density and potential. It's circular reasoning.

This bug halved the density, which is a log(2) additive difference in the potential because phi = phi_s + ln(ni/ni_sheath), which drives the sheath potential up, but then the sheath potential is lowered, so they are counter-directional effects.

Without bug: Sheath vals = 
out_p[0] = 1406773872409082368.000000 // density when you divide by sqrt(2)
out_p[1] = 0.000000
out_p[2] = 1890.990373 // sheath potential when you divide by sqrt(2)
out_p[3] = 0.000000

With bug: Sheath vals: 
703294069229407616.000000 
0.000000 
945.407151
0.000000

The difference between the two plots above is a uniform 17 volts.

ammarhakim:
I would like to understand this better. Printing things will not help here. We need to design unit tests that do automated testing, not be printing things out for a human to see. Plus, if this code is used in a regression test then that means there is a regression test for this already.

One strange thing here is that ion density and phi expansions are in the same array. I do not understand this. Why are two disparate physical quantities stored in the same array? This is a sign of bad design. Perhaps the explanation here is that we have 1 broadcast vs 2. but I do not think that is sufficiently good reason. 

It would be good for @Maxwell-Rosen to take ownership of the Boltzmann elc model and submit a DR to fix the design and also add a doc to that DR describing the equations so we understand this issue properly.

manauref (approved):

Maxwell-Rosen:
I approve Mana's changes. Thanks for implementing the average as intended!

ammarhakim (approved):
Ready to merge.

PR #641, opened 2025-03-08
Port the total M0 preserving source from VM to GK with enhanced M0 preservation in the long term

Maxwell-Rosen:
# Feature

## Summary

Purpose? For steady-state simulations, it can be helpful to preserve the total M0 in the system. This feature scales the source flux to the boundary flux of the problem to ensure total M0 is conserved.

I worked with @kltrbrds to bring this functionality over. There were some intricacies to deal with over Jacobians and factors of $dz/2$ in the boundary fluxes. This should work with the non-uniform grids because we are taking a ratio of input flux to output flux. This is a little different than how it is handled in the Vlasov-Maxwell modules, but it is more robust.

I use a spring-like restoring force to the total M0 and source strength. This term is important because machine precision errors will compound over millions of timesteps and grow to something big. This term ensures were are oscillating around the initial value (target).

$$\Gamma_{source}^\prime = \frac{\Gamma_{boundary} }{ \Gamma_{source} }* (1 + F),$$

$$ F = - k * \frac{M0_{current} - M0_{target}}{M0_{target} } $$

where $k$ is M0_feedback_strength. I found $k=10^3$ effective in this unit test, shown at the bottom, but that's not guaranteed for everyone. I made it public in the input file. Set $k=0$ to remove this feature. It is zero by default.

## Implementation Details

Key changes: Key changes include some new terms being initialized in gk_species_source.c

Added a new kind of source that preserves M0

Dependencies: None

DR: #643

Automated testing: I added a regression test that uses this feature. It's a 1x gk_sheath regression test

## Example Use

In this example $\Gamma_{bflux} = \Gamma_{ion}$.

```   
 .source = {
      .source_id = GKYL_BFLUX_SOURCE,
      .source_species = "ion", // The species of which the fluxes are used to support the total M0 of this species
      .evolve = true, // This must be here
      .M0_feedback_strength = 1e3,

      .num_sources = 1,
      .projection[0] = {
        .proj_id = GKYL_PROJ_MAXWELLIAN_PRIM,
        .density = evalElcSourceDensityInit, // Can have any amplitude
        .ctx_density = &ctx,
        .temp = evalElcSourceTempInit,
        .ctx_temp = &ctx,
        .upar = evalElcSourceUparInit,
        .ctx_upar = &ctx,
      },
      .diagnostics = { // You need at least the M0 diagnostic on the source for this to work because the source reads M0.
        .num_diag_moments = 1,
        .diag_moments = { "M0"},
      },
    },

    .boundary_flux_diagnostics = { // Needed to read the boundary fluxes
      .num_integrated_diag_moments = 1,
      .integrated_diag_moments = { "FourMoments" },
    },
```

## Community Standards

- [x] Changes to `/zero` should have a unit test. None here. Only app-level changes.

## Testing:

- [x] Verified that when M0_feedback_strength that $k=0$

## Additional Notes

Here is gk_sheath_1x2v_p1_bflux_source
image

Here is rt_gk_sheath_3x2v_p1
image

Production level WHAM 1x test
image


manauref:
Can you submit a DR for this work please? We should schedule a DR discussion on adaptive sources including @Antoinehoff too, so we can incorporate this and the stuff he's working on in a coherent manner.

ammarhakim:
A DR for this should be easy: just copy parts of this PR and turn that into a DR. Then @Antoinehoff can add his stuff to that DR and we can review in one shot.

PR #642, opened 2025-03-09
The adiabatic coupling array needs to be initialized to 0 

JunoRavin:
so that the logic that I copied from other DG equations works correctly for not indexing an unallocated array. I was accidentally being saved by certain compilers being okay with this while others were not.


ammarhakim (approved):
Ready to merge

JonathanGorard (approved):

PR #645, opened 2025-03-12
Disable the linux build for the GH actions

Maxwell-Rosen:
The linux build was failing, so we should just not use it at all. I commented it out. Someone in the future can fix it.


JonathanGorard (approved):
Nice! Should have added `make regression` to the GitHub actions long ago.

At some point we should try adding @ammarhakim's new Valgrind unit test checks (falling back to using Leaks on Mac OS) to the automatic CI system too. But for now this is already a great improvement over what is done already.

JunoRavin (approved):

PR #657, opened 2025-03-21
First pass at provably-correct solvers

JonathanGorard:
This PR accompanies the `gkylcas` PR https://github.com/ammarhakim/gkylcas/pull/70, which contains the associated Racket code required to generate both the C implementations themselves, as well as their formal proofs of correctness.

In this branch, we switch out the `wv_eqn` objects for the linear advection, inviscid Burgers', Maxwell's and isothermal Euler equations, and replace them with automatically-generated, formally-verified equivalents from `gkylcas`. Each equation object includes both a Lax-Friedrichs and a Roe solver, with Roe used by default. The header/private header/source file format is identical across advection, Burgers' and isothermal Euler; for Maxwell's things are a little different, since it needs to be possible to instantiate the Maxwell equation object on GPUs (as part of the Vlasov app), and so all of the function implementations are contained within the private header directly, rather than as part of the corresponding source file. It's possible that, as part of our forthcoming work on GPU-acceleration of the fluid solvers as part of the GPU hackathon this summer, we may transition towards all of our `wv_eqn` objects employing this format (or something equivalent to it).

We also switch out the minmod, monotonized-centered, superbee and van Leer flux limiters in `wave_prop` for automatically-generated, formally-verified equivalents from `gkylcas`, in much the same way.

All proof certificates are contained within a new `/proofs` directory, with a `finite_volume` subdirectory only for now (although eventually the plan would be to have equivalent subdirectories for source solvers, DG schemes, etc.). The core prover code is also enclosed (duplicated from the version in `gkylcas`), so that each proof certificate can be executed as a standalone piece of code. Again, we may end up devising a better infrastructure for this at some point. Paths to the relevant proof certificates are given as part of the comments within the header/source files for each equation object. Eventually, our plan is for these to be documented in some systematic way, so that (for instance, from the command line) it becomes possible to see exactly which verifiable properties hold for each equation type and each solver, and automatically obtain links to the corresponding Racket files.


ammarhakim (approved):
Ready to merge.

PR #658, opened 2025-03-21
Axisymmetric Neutrals for 2x3v tokamaks

akashukla:
This branch makes it so for tokamak geometry, 2x3v axisymmetric neutrals use a truly axisymmetric coordinate system with coordinate alpha= phi.


tnbernard (approved):
Looks good.

Issue #667, opened 2025-03-27
Fix sheath potential in Boltzmann elc model with curvilinear/nonuniform coordinates


manauref:
I think I understand what's wrong and how to fix the sheath potential calculation in the Boltzmann electron model when using curvilinear/nonuniform coordinates. The current model calculates the sheath potential using
$$\phi_s = \frac{T_e}{e}\ln\left(\frac{n_{i,s}v_{te}}{\sqrt{2\pi}\Gamma_{i,s}}\right)$$
where the subscript $s$ means evaluated at the sheath entrance, and $\Gamma_i$ is the ion particle flux. This equation follows from computing the electron particle flux, say at the upper boundary, using
$$\Gamma_e = \frac{2\pi}{m} \int_0^{v_{\parallel\max}}\mathrm{d}v_\parallel\int_0^{\mu_{\max}}\mathrm{d}\mu ~ v_{\parallel}~ B f_{Me}$$
where $f_{Me}$ is a Maxwellian electron distribution, and equaling it to the ion particle flux.

The problem with the last equation is that it assumes that the advecting speed along the magnetic field is $v_{\parallel}$, but in field-aligned coordinates it is actually $[\mathcal{C}/(J_x ~ B)]v_\parallel$ (this is assuming axisymmetry or $b_2=0$). So my hypothesis is that we should change the first equation above by
$$\phi_s = \frac{T_e}{e}\ln\left(\frac{(\mathcal{C}/(J_x B)))n_{i,s}v_{te}}{\sqrt{2\pi}\Gamma_{i,s}}\right)$$

I suspect that in 1D mirror simulations this will make little to no change, because $J_x=1/B$ and $\mathcal{C}$ is 1 or some constant close to it. But this may fix the issues when using nonuniform z, @Maxwell-Rosen will test in the [gk-boltz_elc_phis_fix](https://github.com/ammarhakim/gkylzero/tree/gk-boltz_elc_phis_fix) branch.

Maxwell-Rosen:
I've tested and compared a uniform to non-uniform WHAM production level calculation, and once this is accounted for, we get the same sheath potential in both cases. The above correction gives the expected results when changing the grid.

[...]

PR #668, opened 2025-03-28
Change the definition of the sheath potential in the GK Boltzmann electron model

manauref:
This addresses issue #667 

The Boltzmann electron model has been using the following definition of the sheath potential
$$\phi_s = \frac{T_e}{e}\ln\left(\frac{n_{i,s}v_{te}}{\sqrt{2\pi}\Gamma_{i,s}}\right)$$
but this arises from assuming that the advecting velocity along the field line is $v_\parallel$. In field-aligned coordinates, neglecting perpendicular drifts and using $b_2=0$ the electron kinetic equation is
$\frac{\partial(J_x B f_e)}{\partial t} + \frac{\partial}{\partial z} \frac{\mathcal{C}}{J_x B}v_\parallel J_x B f_e + \frac{\partial}{\partial v_\parallel}\dot{v_\parallel}J_x B f_e = 0$
Hence, assuming a non-drifting Maxwellian $f_e$ with density $n_i$ (assuming quasineutrality) and temperature $T_e=mv_{te}^2$ , the electron particle flux (multiplied by $J_x$) out of the z boundary is
$J_x\Gamma = \frac{\mathcal{C}}{J_x B}\frac{n_iv_{te}}{\sqrt{2\pi}}$
So our Boltzmann elc model needs to use the following for the sheath potential:
$$\phi_s = \frac{T_e}{e}\ln\left(\frac{\mathcal{C}}{J_x B}\frac{n_{i,s}v_{te}}{\sqrt{2\pi}\Gamma_{i,s}}\right)$$

Note that in the mirror simulations we published in M. Francisquez et al, PoP 30 102504 (2023) using simulations produced with g2 and a Lorentzian analytic equilibrium $\mathcal{C}=1$ and $J_xB=1$, so the new formula effects no change, as shown in this comparison with a new simulation produced with gkylzero:

Screenshot 2025-03-28 at 9 09 04 AM

But with nonuniform $z$ grids this can have an effect because $J_xB$ is no longer 1 at the boundary, nor constant along the field.


Maxwell-Rosen (approved):

Antoinehoff:
Since this PR changes kernels and affect a ctest, it would be good to show that it was valgrind checked. Then I the rest is fine I guess

manauref:
@Antoinehoff the ctest_ambi_bolt_potential unit test is valgrind clean on Perlmutter, and so is the rt_gk_ltx_boltz_elc_1x2v_p1 reg test which uses these models

Antoinehoff (approved):

Issue #660, opened 2025-03-24
[DR] Perpendicular field solve for multiblock tokamak GK sims


manauref:
Multiblock gyrokinetic simulations of tokamaks need a solution to the quasineutrality equation

$$ - \nabla\cdot\epsilon \nabla_\perp\phi = \rho$$

across multiple blocks. In this stage of the project we propose to extend the current approach used in single block sims to multiple blocks, that is:
1. Smooth $\rho$ along the field line.
2. Solve the quasineutrality problem in z-slabs.
3. Smooth $\phi$ along the field line.

We provide 2 stages of implementation here. The first one uses solvers we already have, the second one requires new solvers to enforce continuity and allow for more flexibility.

Stage 1:
======

Smoothing the charge density
------------------------------

This is presently done with the `fem_parproj` updater, or could be done with its multiblock counterpart `fem_parproj_multib` (in [gk-g0-app_fem_parproj_mb](https://github.com/ammarhakim/gkylzero/tree/gk-g0-app_fem_parproj_mb)). These solvers project $\rho$ onto an FEM basis that is local (to a cell) in x-y and nonlocal along z. Smoothing the charge density can be done independently in all the regions connected by the magnetic field. Consider the following sketch of a lower single null poloidal cross section:

![Image](https://github.com/user-attachments/assets/a42dc298-5594-4c5f-82f1-61b063135a20)

We can do 3 independent smoothings (red, green, purple), each of which of which only couples 1-3 blocks that are connected by B. If using `fem_parproj` (as we do now), we simply gather $\rho$ across the connected blocks, apply `fem_parproj` and then take the piece of $rho$ (from the z-multiblock $\rho$) that is local to the block or MPI rank in question.

Aside: There's a question of whether this smoothing should be done in physical space, which introduces a Jacobian $J_x$ weight to the integral in computational coordinates, or whether it should be done in computational space, in which case no such $J_x$ appears. The current updaters allow specifying such weight or not. I'll note that to my knowledge proofs of conservation only require $\phi$ to be continuous in computational space, although this partly depends on how one defines self-adjointness. It may be possible to write down another consistent conservation proof which uses a smoother with a J_x.

Solving quasineutrality in z-slabs
---------------------------------
[...]

PR #669, opened 2025-03-28
Perpendicular GK field solve for multiblocks

manauref:
This is the implementation of stage 1 of the perpendicular field solve for multiblock gyrokinetic simulations, described in DR #660

Here we simply apply fem_poisson_perp out of the box. That is:
1. Smooth the charge density $\rho$ across magnetically blocks, producing $\bar{\rho}$.
2. Gather $J_x ~ g^{ij} \epsilon_\perp$ across radially connected blocks, so each block has a radially global version of $J_x ~ g^{ij}\epsilon_\perp$.
3. Gather $\bar{\rho}$ across radially connected blocks, so each block has a radially global version of $\bar{\rho}$
4. Each block solves the radially global perpendicular Poisson problem: $-\nabla \cdot J_x ~ g^{ij}\epsilon_\perp\nabla_\perp\phi = \bar{\rho}$.
5. Smooth the potential $\phi$ across magnetically blocks, producing $\bar{\phi}$.

Steps/operations 1 and 5 were already in `main`, here we simply implement steps 2-4. Step 4 is done in z-slabs using fem_poisson_perp out of the box. This simply works, but there are two caveats:
- The coefficients $J_x ~ g^{ij}$ are discontinuous across radially connected blocks. Despite this, the FEM Poisson solver behaves fine, i.e. the potential that comes out of it is smooth and continuous. After all FEMs have been used with discontinuities for a long time, but it'd be good to explore the implications of this a little more when we have time.
- Because we use fem_poisson_perp out of the box we are limited to grids that have the same $\Delta x$ in radially connected blocks, but because fem_parproj (for the parallel smoothing) requires magnetically connected blocks to have the same $\Delta x$, this means every block in the sim needs to have the same $\Delta x$. One solution to this restriction is to create a new updater (aka fem_poisson_perp_multib) which allows for each block to have its own $\Delta x$, similar to fem_parproj_multib in the  [gk-g0-app_fem_parproj_mb](https://github.com/ammarhakim/gkylzero/tree/gk-g0-app_fem_parproj_mb) branch.

Despite these restrictions, it seems that multiblock tokamak simulations run and produce sensible results (in the quadgeo branch(es)).


akashukla (approved):

akashukla:
Here is an example radial profile of the potential from a double-null multi-block tokamak simulation. In cases I have checked, the relative difference between phi at the radial block boundaries is quite small (1e-3 or less). 

image

Antoinehoff:
since this branch seems to declare new variables, I recommend to test if it is valgrind clean before merging. Otherwise I have nothing much to say, the limitations about the $\Delta x$ may be a bit annoying but I guess it can be considered later.

manauref:
@Antoinehoff I just checked a multiblock regression test and the perp field solve is valgrind clean

manauref:
on Perlmutter CPU that is

Antoinehoff (approved):

PR #671, opened 2025-03-28
Numerical IWL geometry

akashukla:
Add inner wall limited geometry to our numerical tokamak equilibrium types. Example test with an LTX efit file (generated using miller parameters) produces good results. Below I show the nodes, the configuration space jacobian, and the magnetic field magnitude.
image
image


tnbernard:
This feature is great! Can you add an example ltx regression test that uses it?

PR #672, opened 2025-03-28
Some regression tests about a leaky bag

Maxwell-Rosen:
# Feature

## Summary

Purpose: We expand our regression tests using a test from a leaky simulation. Through this, we have an analytic solution for the fluxes out of the simulation. We can use this to diagnose issues with the Boltzmann electron model and the boundary fluxes related to the geometry.

Issue link: #667 #668

Modled after https://ammar-hakim.org/sj/je/je20/je20-vlasov-bounded.html

## Implementation Details

Key changes: Added some regression tests

The sheath regression tests fail and this is addressed by issue #667 

_Describe any new components, classes, or modules introduced._

Automated testing: I included checks of the analytic values in the regression test. Tolerance is much worse than I would like. 

## Community Standards

Valgrind sanitized on perlmutter

## Additional Notes

Future work could explore why the tolerance of these regression tests is so poor

### DIFF
--- a/regression/rt_gk_leaky_bag_1x2v_p1.c
+++ b/regression/rt_gk_leaky_bag_1x2v_p1.c
@@ -1,0 +1,599 @@
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+#include <gkyl_alloc.h>
+#include <gkyl_const.h>
+#include <gkyl_dynvec.h>
+#include <gkyl_gyrokinetic.h>
+#include <gkyl_util.h>
+
+#include <gkyl_null_comm.h>
+
+#ifdef GKYL_HAVE_MPI
+#include <mpi.h>
+#include <gkyl_mpi_comm.h>
+#ifdef GKYL_HAVE_NCCL
+#include <gkyl_nccl_comm.h>
+#endif
+#endif
+
+#include <rt_arg_parse.h>
+
+struct boundary_ctx
+{ 
+  // Physical constants (using non-normalized physical units).
+  double epsilon0; // Permittivity of free space.
+  double mass_ion; // Proton mass.
+  double charge_ion; // Proton charge.
+
+  double Ti; // Ion temperature.
+  double n0; // Reference number density (1 / m^3).
+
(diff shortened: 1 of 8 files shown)


manauref:
@Maxwell-Rosen can you make a list of the tests and the purpose of each? I suspect we don't need so many. Keep in mind that it's hard to maintain so many tests, so if some of them can be consolidated that'd be good

Maxwell-Rosen:
The differences in the regression tests have to do with the geometry. I also include tests to verify the sheath potential from the Boltzmann electron model.

_rt_gk_leaky_bag_1x2v_p1_
This one is uniform geometry with an identity mapc2p. This is the base-case to compare to

_rt_gk_leaky_bag_1x2v_p1_ux_stretch_
This one has a uniform geometry but mapc2p has z_c = 5*z_p, so the computational grid is stretched by 5x to be the physical grid. In every other way, this is the same as rt_gk_leaky_bag_1x2v_p1. The domain length is also divided by 5 to compensate. This is useful in debugging issues with the geometry compression.

_rt_gk_leaky_bag_1x2v_p1_nux_center_
This test uses the non-uniform infrastructure to deform the grid and make sure that the loss rates are what we would expect. Cells are concentrated near the center of the domain.

_rt_gk_leaky_bag_1x2v_p1_nux_edge_
This test also uses the non-uniform infrastructure, but concentrates cells at the edge of the domain.

The following tests only run 1 time step to check the sheath potential.
_rt_gk_leaky_bag_boltz_sheath_1x2v_p1_
Uniform geometry. Base case

_rt_gk_leaky_bag_boltz_sheath_1x2v_p1_ux_stretch_
5x stretched in mapc2p to check the uniform grids.

_rt_gk_leaky_bag_boltz_sheath_1x2v_p1_nux_center_
Cells are concentrated at the center of the domain. Checks the sheath potential

_rt_gk_leaky_bag_boltz_sheath_1x2v_p1_nux_edge_
Cells concentrated at the edge of the domain to check the sheath potential

manauref (approved):

JunoRavin (approved):

PR #677, opened 2025-04-03
--c2p option works with deflated mc2p and mc2nu_pos arrays

Maxwell-Rosen:
# Bug fix

## Summary

Description: So for a long time, I've been trying to use the `--c2p` option in postgkyl on my mirror simulations. In a 1x simulation, I'd expect the x-axis and coordinate to by cylindrical Z. Instead it would give an error that the interpolation wasn't working. This has puzzled me for a long time and I finally got around to fixing this issue.

Issue link: I was bad and didn't make an issue for this. In the postgkyl repo, it's here https://github.com/ammarhakim/postgkyl/issues/157

## Solution

The issue is that mapc2p is always written as a 3 component array. It always has R,Z,phi. They can either be a function of 1,2, or 3 components, but there are always 3 of them. This is packaged inside mapc2p. Postgkyl expects an array that has the number of components as the dimension of the array. So in 2x, it would expect only to have R,Z or Z,phi. We are giving it too many parts when using --c2p <>-mapc2p.gkyl. The solution is to create a file <>-mapc2p_reduced.gkyl which has selected components removed. Furthermore, we also do this with <>-mc2nu_pos_reduced.gkyl. **This is only intended to be used for plotting purposes**. 

I discussed with @akashukla and it seems reasonable that a 1x simulation would want cylindrical Z and a 2x simulation would look for the R,Z plane. This makes a lot of sense for tokamaks and mirrors, but not so much for mapc2p geometries. We will not consider those because they can be setup in any orientation the user wants. In mapc2p geometry, it will plot y in 1x,  and x,y in 2x. For the non-uniform grids and position maps, <>-mc2nu_pos_reduced.gkyl will plot z in 1x, and x,z in 2x.

Rather than changing postgkyl to work with GK, it is much safer to change GK to fit postgkyl and create plotting-only outputs.

In doing this, I needed to make a new array operation that assigns components of one array to another. The existing operation applies all components > input, but I want to select each component individually. I implemented this on GPU and CPU. Tested and verified with a unit test. I added some comments to the files to clarify what these functions are doing.

This change is very important for plotting non-uniform grids.

Impacted files: 
gyrokinetic.c
gk_geometry.c
array_ops.c
array_ops_cu.cu
ctest_array_ops.c

Automated testing: Unit test is added for the array operation added

## Community Standards

## Testing:

- [x] I added a unit test for this feature.
- [ ] I ran a few regression tests to ensure no apparent errors.

## Additional Notes

Here is an example of the 2x2v mirror regression test being plotted. 
image

### DIFF
--- a/apps/gyrokinetic.c
+++ b/apps/gyrokinetic.c
@@ -835,13 +835,16 @@ gkyl_gyrokinetic_app_write_geometry(gkyl_gyrokinetic_app* app)
 
   // Gather geo into a global array
   struct gkyl_array* arr_ho1 = mkarr(false,   app->basis.num_basis, app->local_ext.volume);
+  struct gkyl_array* arr_hocdim = mkarr(false, app->cdim*app->basis.num_basis, app->local_ext.volume);
   struct gkyl_array* arr_ho3 = mkarr(false, 3*app->basis.num_basis, app->local_ext.volume);
   struct gkyl_array* arr_ho6 = mkarr(false, 6*app->basis.num_basis, app->local_ext.volume);
   struct gkyl_array* arr_ho9 = mkarr(false, 9*app->basis.num_basis, app->local_ext.volume);
 
   struct timespec wtm = gkyl_wall_clock();
   gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2p        , arr_ho3, "mapc2p", mt);
-  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos        , arr_ho3, "mc2nu_pos", mt);
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2p_reduced, arr_hocdim, "mapc2p_reduced", mt);  
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos   , arr_ho3, "mc2nu_pos", mt);
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos_reduced, arr_hocdim, "mc2nu_pos_reduced", mt);  
   gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->bmag        , arr_ho1, "bmag", mt);
   gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->g_ij        , arr_ho6, "g_ij", mt);
   gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->dxdz        , arr_ho9, "dxdz", mt);
--- a/zero/array_ops.c
+++ b/zero/array_ops.c
@@ -126,6 +126,30 @@ gkyl_array_set_offset(struct gkyl_array* out, double a,
 }
 
 struct gkyl_array*
+gkyl_array_set_offset_comp(struct gkyl_array* out, double a,
+  const struct gkyl_array* inp, int component_offset_out, 
+  int component_offset_inp, int num_basis)
+{
+  assert(out->type == GKYL_DOUBLE);
+  assert(out->size == inp->size);
+
+#ifdef GKYL_HAVE_CUDA
+  assert(gkyl_array_is_cu_dev(out)==gkyl_array_is_cu_dev(inp));
+  if (gkyl_array_is_cu_dev(out)) { gkyl_array_set_offset_comp_cu(out, a, inp, component_offset_out, component_offset_inp, num_basis); return out; }
+#endif
+
+  int offset_out = component_offset_out * num_basis;
+  int offset_inp = component_offset_inp * num_basis;
+
+  double *out_d = out->data;
+  const double *inp_d = inp->data;
+  for (size_t i=0; i<out->size; ++i)
+    for (size_t d=0; d<num_basis; ++d)
+      out_d[i*NCOM(out)+d+offset_out] = a*inp_d[i*NCOM(inp)+d+offset_inp];
+  return out;
+}
+
+struct gkyl_array*
 gkyl_array_scale(struct gkyl_array* out, double a)
 {
 #ifdef GKYL_HAVE_CUDA
--- a/zero/gk_geometry.c
+++ b/zero/gk_geometry.c
@@ -32,7 +32,9 @@ gkyl_gk_geometry_new(struct gk_geometry* geo_host, struct gkyl_gk_geometry_inp *
 
   // bmag, metrics and derived geo quantities
   up->mc2p = gkyl_array_new(GKYL_DOUBLE, 3*up->basis.num_basis, up->local_ext.volume);
+  up->mc2p_reduced = gkyl_array_new(GKYL_DOUBLE, up->grid.ndim*up->basis.num_basis, up->local_ext.volume);
   up->mc2nu_pos = gkyl_array_new(GKYL_DOUBLE, 3*up->basis.num_basis, up->local_ext.volume);
+  up->mc2nu_pos_reduced = gkyl_array_new(GKYL_DOUBLE, up->grid.ndim*up->basis.num_basis, up->local_ext.volume);
   up->bmag = gkyl_array_new(GKYL_DOUBLE, up->basis.num_basis, up->local_ext.volume);
   up->g_ij = gkyl_array_new(GKYL_DOUBLE, 6*up->basis.num_basis, up->local_ext.volume);
   up->dxdz = gkyl_array_new(GKYL_DOUBLE, 9*up->basis.num_basis, up->local_ext.volume);
@@ -241,7 +243,9 @@ gkyl_gk_geometry_deflate(const struct gk_geometry* up_3d, struct gkyl_gk_geometr
 
   // bmag, metrics and derived geo quantities
   up->mc2p = gkyl_array_new(GKYL_DOUBLE, 3*up->basis.num_basis, up->local_ext.volume);
+  up->mc2p_reduced = gkyl_array_new(GKYL_DOUBLE, up->grid.ndim*up->basis.num_basis, up->local_ext.volume);
   up->mc2nu_pos = gkyl_array_new(GKYL_DOUBLE, 3*up->basis.num_basis, up->local_ext.volume);
+  up->mc2nu_pos_reduced = gkyl_array_new(GKYL_DOUBLE, up->grid.ndim*up->basis.num_basis, up->local_ext.volume);
   up->bmag = gkyl_array_new(GKYL_DOUBLE, up->basis.num_basis, up->local_ext.volume);
   up->g_ij = gkyl_array_new(GKYL_DOUBLE, 6*up->basis.num_basis, up->local_ext.volume);
   up->dxdz = gkyl_array_new(GKYL_DOUBLE, 9*up->basis.num_basis, up->local_ext.volume);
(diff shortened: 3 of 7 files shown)


zero/array_ops.c
```diff
@@ -125,6 +125,30 @@ gkyl_array_set_offset(struct gkyl_array* out, double a,
...
+  assert(out->type == GKYL_DOUBLE);
+  assert(out->size == inp->size);
+
+#ifdef GKYL_HAVE_CUDA
+  assert(gkyl_array_is_cu_dev(out)==gkyl_array_is_cu_dev(inp));
+  if (gkyl_array_is_cu_dev(out)) { gkyl_array_set_offset_comp_cu(out, a, inp, component_offset_out, component_offset_inp, num_basis); return out; }
+#endif
+
+  int offset_out = component_offset_out * num_basis;
+  int offset_inp = component_offset_inp * num_basis;
```
JunoRavin, line 142: It's fine to add this array method, but keep in mind that array methods are not DG specific and thus should not be coded up as DG specific. You can have a component_offset_out and component_offset_inp, and if someone is doing something with DG, these components will need to be multiplied by num_basis. 

So don't take num_basis as an input and appropriately adjust your use of this method in your unit tests and gk_geometry.

zero/gk_geometry.c
```diff
@@ -301,6 +305,21 @@ gkyl_gk_geometry_deflate(const struct gk_geometry* up_3d, struct gkyl_gk_geometr
   // Done deflating
   gkyl_deflate_geo_release(deflator);
 
+  if (up->grid.ndim==1) {
```
JunoRavin, line 308: I would add comments here for these offset assignments because as is I'm going cross-eyed trying to figure out how these offsets make sense for the assignment here.

zero/gkyl_gk_geometry.h
```diff
@@ -27,7 +27,9 @@ struct gk_geometry {
   // GK Equation and Poisson Equation and to apply certain BC's
   // The first 20 are defined on the configuration space domain. The last is a single element.
   struct gkyl_array* mc2p; // 3 components. Cartesian X,Y, and Z
+  struct gkyl_array* mc2p_reduced; // cdim components. Component removed (Z in 1x, R,Z in 2x, R,Z,phi in 3x)
   struct gkyl_array* mc2nu_pos; // 3 components. Uniform computational space to non-uniform computational space mapping
+  struct gkyl_array* mc2nu_pos_reduced; // cdim components. Uniform computational space to non-uniform computational space mapping
```
JunoRavin, line 32: Why do you need both of these? We only need one mapping for the IO correct? You should be writing out a final composite mapping no?

Maxwell-Rosen: I assign the array to this struct so that it can be deflated with the rest of the geometry when that happens. We could allocate this array during the geometry write, but then we would have to do a deflation inside the geometry write. I put it here because it is a part of the geometry, although it's redundant information

manauref (changes requested):
Line 144 of array_ops.c seems suspect to me. Why do you have to loop over components? Same thing in GPU code

also, can you describe why this method is needed in the PR? I’m wondering if the offset methods that we have are sufficient

Maxwell-Rosen:
This method is neccisary because of line 327 in gk_geometry.c and 319. In 2x, we want to get R,Z which are components 0 and 2 of mapc2p. Skipping over the middle one means it's neccisary to implement this new method. The method set_offset would restart back at index zero when it is called twice. I also think this may cause an issue and bug right here https://github.com/ammarhakim/gkylzero/blob/phase-cell-update-turn-off/apps/gk_field.c#L449

zero/gkyl_gk_geometry.h
manauref, line 32: I think these _reduced should be renamed _deflated. That's the terminology we typically use in the geo code for these sorts of things.

apps/gyrokinetic.c
```diff
@@ -875,13 +875,16 @@ gkyl_gyrokinetic_app_write_geometry(gkyl_gyrokinetic_app* app)
...
   struct gkyl_array* arr_ho3 = mkarr(false, 3*app->basis.num_basis, app->local_ext.volume);
   struct gkyl_array* arr_ho6 = mkarr(false, 6*app->basis.num_basis, app->local_ext.volume);
   struct gkyl_array* arr_ho9 = mkarr(false, 9*app->basis.num_basis, app->local_ext.volume);
 
   struct timespec wtm = gkyl_wall_clock();
   gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2p        , arr_ho3, "mapc2p", mt);
-  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos        , arr_ho3, "mc2nu_pos", mt);
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2p_reduced, arr_hocdim, "mapc2p_reduced", mt);  
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos   , arr_ho3, "mc2nu_pos", mt);
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos_reduced, arr_hocdim, "mc2nu_pos_reduced", mt);
```
manauref, line 887: Why are you writing 2 new things? I would think for plotting you only need one of these

zero/gkyl_array_ops.h
```diff
@@ -93,6 +93,10 @@ struct gkyl_array* gkyl_array_set(struct gkyl_array *out,
  * Set out = a*inp[coff] where coff is a component-offset if
  * out->ncomp < inp->ncomp, or out[coff] = a*inp if
  * out->ncomp > inp->ncomp. Returns out.
+ * 
+ * Sets all components greater than the coeff offset to the output
+ * e.g. out[0,1,2, ...] = a*inp[2,3,4,...]
+ * where the middle indicies are the different components
```
manauref, line 99: this doesn't sound right to me, and I think I implemented these offset methods.

Maxwell-Rosen:
I just explained that set_offset is specifically not sufficient in 2D. We could use set_offset for 1D since it's only picking out a single component, but because in 2D we need to jump over component 1. We need a new array method

zero/array_ops.c
```diff
@@ -125,6 +125,27 @@ gkyl_array_set_offset(struct gkyl_array* out, double a,
...
+#ifdef GKYL_HAVE_CUDA
+  assert(gkyl_array_is_cu_dev(out)==gkyl_array_is_cu_dev(inp));
+  if (gkyl_array_is_cu_dev(out)) { gkyl_array_set_offset_comp_cu(out, a, inp, offset_out, ffset_inp); return out; }
+#endif
+
+  double *out_d = out->data;
+  const double *inp_d = inp->data;
+  for (size_t i=0; i<out->size; ++i)
+    for (size_t d=0; d<NCOM(out); ++d)
+      out_d[i*NCOM(out)+d+offset_out] = a*inp_d[i*NCOM(inp)+d+offset_inp];
```
manauref, line 145: I think this method is not needed, but if it's kept, it should be modified because the other offset methods handle the case in which NCOM(inp) > NCOM(out) or viceversa. This one only handles the NCOM(inp) > NCOM(out) case (the other case doesn't seem memory safe right now)

manauref (changes requested):

apps/gyrokinetic.c
Maxwell-Rosen: We want to deflate the non-uniform map too. mc2nu_pos is always output in 3x, just like mapc2p

zero/gkyl_array_ops.h
Maxwell-Rosen: Yes you did. I'm just illustrating how the offset works. Maybe the notation is strange. It'll set index 0 to 2, 1 to 3, 2 to 4, etc.

zero/array_ops.c
Maxwell-Rosen: This is a memory safe operation because it's only assigning 1 component of inp to 1 component of out. I'll put in a modification to make sure the second for loop is over the lower NCOM

zero/gkyl_array_ops.h
Maxwell-Rosen: I updated this. Can you check the new comment is okay?

Maxwell-Rosen:
I'm testing and everything looks like it works with the new changes

zero/gkyl_array_ops.h
```diff
@@ -93,6 +93,12 @@ struct gkyl_array* gkyl_array_set(struct gkyl_array *out,
  * Set out = a*inp[coff] where coff is a component-offset if
  * out->ncomp < inp->ncomp, or out[coff] = a*inp if
  * out->ncomp > inp->ncomp. Returns out.
+ * 
+ * Sets all components greater than the coeff offset to the output
+ * e.g. coff = 2;
+ *      if out has less components out[0,1,2, ...] = a*inp[2,3,4,...]
+ *      if out has more components out[2,3,4, ...] = a*inp[0,1,2,...]
+ * where the middle indicies are the different components
```
manauref, line 101: This isn't accurate. It's not just "all components greater than". Please rephrase in terms of the number of components of the smaller array, and add a finishing index after ... . Otherwise it is wrong.

Also, the correct English is "fewer components" not "less components".

Or just delete this comment.

manauref (changes requested):
another thing: the currently written deflated arrays don't make sense for low dim mapc2p. I think it's better to not write them out, otherwise they will mislead people. So please put an if-statement using the geo ID where the deflated mc2p deflated are populated and written out.

Maxwell-Rosen:
> another thing: the currently written deflated arrays don't make sense for low dim mapc2p. I think it's better to not write them out, otherwise they will mislead people. So please put an if-statement using the geo ID where the deflated mc2p deflated are populated and written out.

Addressed in the latest commit. We decided together that we should allocate these deflated arrays in all geoemtries so that the free method is universal. However, when writing geometry, we want to only write out the mapc2p array in 2x or 1x if we are using a mirror or tokamak geometry

zero/gkyl_array_ops.h
Maxwell-Rosen: I deleted this comment

apps/gyrokinetic.c
```diff
@@ -875,13 +875,19 @@ gkyl_gyrokinetic_app_write_geometry(gkyl_gyrokinetic_app* app)
...
   struct gkyl_array* arr_ho9 = mkarr(false, 9*app->basis.num_basis, app->local_ext.volume);
 
   struct timespec wtm = gkyl_wall_clock();
   gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2p        , arr_ho3, "mapc2p", mt);
-  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos        , arr_ho3, "mc2nu_pos", mt);
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos   , arr_ho3, "mc2nu_pos", mt);
+  if (app->cdim < 3) {
+    if (geometry_inp->geometry_id == GKYL_MIRROR || geometry_inp->geometry_id == GKYL_TOKAMAK)
+      gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2p_deflated, arr_hocdim, "mapc2p_deflated", mt);  
+    gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->mc2nu_pos_deflated, arr_hocdim, "mc2nu_pos_deflated", mt);
```
akashukla, line 889: Is this really supposed to write the deflated position map even for mapc2p geometry?

Maxwell-Rosen: Yes. The position map has the same deflation process. Since the deflation process always isolates the third computational coordinate, that is the one mc2nu_pos describes.

akashukla (approved):

PR #683, opened 2025-04-08
Fix broken wv_maxwell equation object on GPUs

JonathanGorard:
As pointed out by @manauref, the changes I made in PR #657 broke the GPU build, essentially because of the use of the host-side `gkyl_malloc` and `gkyl_free` functions within the `wv_maxwell` equation object, which needs to be instantiatable on _both_ host and device because of the way the Vlasov app works. Sorry! I've now restructured the equation object to allocate all arrays on the stack and to avoid the use of these functions entirely. The GPU build now works on Stellar-AMD, although it would be good to verify this independently. Apologies for the inconvenience!


manauref (approved):

JunoRavin (approved):
I've also confirmed this branch compiles on Stellar-AMD with cuda 12.4

PR #684, opened 2025-04-11
Simplify the GK-neutral recycling API in input files

manauref:
this way the user has less room for error

@tnbernard I couldn't make the appropriate change in the rt_gk_neut_recycle_3x3v test because it still has the old way of specifying recycling. Could you update that please?


tnbernard:
Why couldn't we put the projection into the recycling app instead of species, which already has a ton of extra functions? We could do it more directly instead of by the fixed function bc approach, which was just an intermediate step to begin with.

manauref:
we could, i just didn't know you wanted to move the projection into the
recycling app, but yes I agree that that's where it should go

Issue #688, opened 2025-04-16
Machine build is not working anymore (on MacOS)


Antoinehoff:
I encounter two issues (and report a third one) when calling the `mkdeps.macos.sh` script to install from scratch gkylzero on my laptop (M3 Pro, Sonoma 14.7.2).
This is linked to the PR #689.

### LuaJIT git clone
It looks like the cloning of the LuaJIT git repository is made with an outdated link:
```
Building Luajit
Downloading LuaJIT ..
Cloning into 'luajit'...
fatal: unable to access 'https://luajit.org/git/luajit.git/': SSL certificate problem: self signed certificate
./build-luajit.sh: line 19: cd: luajit: No such file or directory
Building LuaJIT ..
./build-luajit.sh: line 25: cd: luajit: No such file or directory
make: *** No rule to make target `install'.  Stop.
```

This is resolved by replacing
```
git clone https://luajit.org/git/luajit.git
```
by 
```
git clone https://github.com/LuaJIT/LuaJIT.git
```
in `install-deps/build-luajit.sh`.

### LuaJIT deployment target variable for MaxOSX
Additionally, LuaJIT installation requires now to define the following Makefile variable
```
export MACOSX_DEPLOYMENT_TARGET=14.7
```
which I add in `machines/mkdeps.macos.sh`.

---
#### Side comment on SuperLU with cmake 4.0.0
The build of SuperLU stops with the current state of the build script and cmake 4.0.0 yields:

```
Building SUPERLU
Downloading SuperLU ..
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 2423k    0 2423k    0     0  3638k      0 --:--:-- --:--:-- --:--:-- 3638k
Building SuperLU ..
x superlu-5.2.2/
x superlu-5.2.2/.gitignore
x superlu-5.2.2/CBLAS/
x superlu-5.2.2/CBLAS/CMakeLists.txt

[...]

x superlu-5.2.2/cmake/XSDKDefaults.cmake
x superlu-5.2.2/make.inc.in
x superlu-5.2.2/superlu.pc.in
CMake Error at CMakeLists.txt:8 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
[...]

PR #689, opened 2025-04-16
Build fix (for MacOS and more)

Antoinehoff:
This PR is related to DR #688 and aims to fix the machine build.

### DIFF
--- a/install-deps/build-luajit.sh
+++ b/install-deps/build-luajit.sh
@@ -14,9 +14,9 @@ if [ "$DOWNLOAD_PKGS" = "yes" ]
 then
     echo "Downloading LuaJIT .."
     # delete old checkout and builds
-    rm -rf luajit
+    rm -rf LuaJIT
     git clone https://github.com/LuaJIT/LuaJIT.git
-    cd luajit
+    cd LuaJIT
 fi
 
 if [ "$BUILD_PKGS" = "yes" ]
--- a/install-deps/build-luajit.sh
+++ b/install-deps/build-luajit.sh
@@ -15,7 +15,7 @@ then
     echo "Downloading LuaJIT .."
     # delete old checkout and builds
     rm -rf luajit
-    git clone https://luajit.org/git/luajit.git
+    git clone https://github.com/LuaJIT/LuaJIT.git
     cd luajit
 fi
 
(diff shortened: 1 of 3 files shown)


Antoinehoff:
The regression test `regression/rt_gk_sheath_3x2v_p1.c` works fine after the modifications

manauref:
Antoine, regarding the SuperLU build, I think your Cmake version is just too old.
Can you do `cmake --version`?

Regarding the luajit issue, I just tried and agree with your issue. But your fixes were not enough to fix it. It seemed to me some directory names need to be changed in the build-luajit script still.

Antoinehoff:
Yeah I am still working on it worry I'll reopen it asap

Antoinehoff:
I had some naming issue with "luajit" vs "LuaJIT". The fix should work now.

manauref:
instead of the 3 changes you made in the build-luajit script, I think it's safer to only change L18 to
```
git clone https://github.com/LuaJIT/LuaJIT.git luajit
```

manauref:
> instead of the 3 changes you made in the build-luajit script, I think it's safer to only change L18 to
> 
> ```
> git clone https://github.com/LuaJIT/LuaJIT.git luajit
> ```

also, I think L19 is not needed, and throws an unnecessary error message

JunoRavin:
Since we're modifying builds, we should actually just take this opportunity to try building with a newer version of superLU and make sure that these fixes still work (or if they're needed).

manauref (approved):

Antoinehoff:
> > instead of the 3 changes you made in the build-luajit script, I think it's safer to only change L18 to
> > ```
> > git clone https://github.com/LuaJIT/LuaJIT.git luajit
> > ```
> 
> also, I think L19 is not needed, and throws an unnecessary error message

Is this comment still valid?

manauref:
> > > instead of the 3 changes you made in the build-luajit script, I think it's safer to only change L18 to
> > > ```
> > > git clone https://github.com/LuaJIT/LuaJIT.git luajit
> > > ```
> > 
> > 
> > also, I think L19 is not needed, and throws an unnecessary error message
> 
> Is this comment still valid?

yes, just read the script logic and you'll see. I think it's not needed

JunoRavin (approved):

PR #691, opened 2025-04-16
Add a moving average to the position map

Maxwell-Rosen:
# Feature

## Summary

Purpose: I encountered these errors where the position map was suddenly changing, so discontinuous in C1. At those locations, there seemed to be a weak division error, noticed in examining the non-uniform calculations with Boltzmann electrons. This had to do with a weak division operation that happened inside the #667 PR. This was addressed. I think this has to do with discontinuities in C1 of the position map, so I implemented a moving average that will reduce these errors

Issue link: #668 

DR link: #693 

## Implementation Details

Key changes: Moving average of the position map for numerical geometries. Really only affects the position map code.

Automated testing: None. Maybe this is bad practice. I want to merge before the gkyl split so it's accounted for

The edge of the simulation is handled by changing the width to have the edge of the simulation on the outer side. If the width is zero, or we are on the edge, return a value without the moving average.

## Example Use

```
      .position_map_info = {
        .id = GKYL_PMAP_CONSTANT_DB_NUMERIC,
        .moving_average_width = 0.2, // field line length of the window for the moving average. 0.1 on the left, 0.1 on the right
      },
```

## Community Standards

- [ ] Changes to `/zero` should have a unit test.

## Additional notes

Here is an example of the use of the moving average. The edge is handled to take a width that goes until the edge of the simulation. Here, red has the moving average, blue has no moving average, and yellow is uniform. This uses a moving average width of 0.5 radians, which is larger than I think we should use. It's exaggerated for a bigger effect than is necessary. In a production run, 0.2 is a reasonable value.

image


manauref:
This hasn't gone through the DR stage. Please initiate the work there

Maxwell-Rosen:
DR submitted at #691

zero/gkyl_position_map_priv.h
```diff
@@ -358,6 +354,12 @@ find_B_field_extrema(struct gkyl_position_map *gpm)
       }
     }
   }
+  
+  // I had a memory issue with doing this before the for loop on my local laptop. Why?
```
JunoRavin, line 358: Please do not leave comments like this in the repo. If you think something is memory unsafe, run it through valgrind. I remember that we talked about this issue and that it was solved by moving the final extrema set to after the loop, but in that case we should (assuming everything is valgrind clean) just add a note that we are setting the final extrema after the loop with your initials and the date to signify within the code not to move this.

zero/gkyl_position_map_priv.h
```diff
@@ -493,6 +495,17 @@ refine_B_field_extrema(struct gkyl_position_map *gpm)
...
   gpm->constB_ctx->dB_cell = B_total_change / (gpm->constB_ctx->N_theta_boundaries);
+
+  // Check that all extrema are increasing
+  for (int i = 1; i < gpm->constB_ctx->num_extrema; i++)
+  {
+    if (gpm->constB_ctx->theta_extrema[i] < gpm->constB_ctx->theta_extrema[i-1])
+    {
+      printf("Numerical position map magnetic field extrema finding failed\n");
+      printf("theta_extrema[%d] = %f\n", i, gpm->constB_ctx->theta_extrema[i]);
+      printf("theta_extrema[%d] = %f\n", i-1, gpm->constB_ctx->theta_extrema[i-1]);
```
JunoRavin, line 506: We don't want to leave print statements in the code like this.

manauref (changes requested):
@Maxwell-Rosen I think the DR and this PR need a little more motivation. There's no public or documented need for this code (although I suspect that it'll be useful to have). Can you show what errors you are referring to?

Aside from that, the biggest thing I'm concerned with is that there is a possibility, IMO, that there's an error somewhere and that by simply smoothing the mapping we are shoving it under the rug. An example of this is the error we found and fixed in PR #668 , where how one handles Jacobians in the Boltzmann elc model has a local impact on the solutions. If such an error exists elsewhere, we should do the careful analysis we did in PR #668 here and fix it, instead of smoothing things.

Also, you link issue #667 and PR #668 but I don't they have anything to do with this, or at least you haven't shown it to.

Maxwell-Rosen:
> @Maxwell-Rosen I think the DR and this PR need a little more motivation. There's no public or documented need for this code (although I suspect that it'll be useful to have). Can you show what errors you are referring to?
> 
> Aside from that, the biggest thing I'm concerned with is that there is a possibility, IMO, that there's an error somewhere and that by simply smoothing the mapping we are shoving it under the rug. An example of this is the error we found and fixed in PR #668 , where how one handles Jacobians in the Boltzmann elc model has a local impact on the solutions. If such an error exists elsewhere, we should do the careful analysis we did in PR #668 here and fix it, instead of smoothing things.
> 
> Also, you link issue #667 and PR #668 but I don't they have anything to do with this, or at least you haven't shown it to.

I added issue #703 to detail why I implemented this, including the figures we have exchanged. I agree that this demands a deeper investigation into other kernels that are dividing by Jacobians, but it seems this is done in the GK kernels too, and those are very intricate.

manauref:
We should close this PR and dig into the code/simulation to understand why these features are appearing.

Maxwell-Rosen:
I disagree. We can also include this feature without necessarily having a justification beyond the fact that we want a smooth non-uniform mapping. It's a low-pass filter.

I just memory-sanitized this branch. It's fine.

Issue #687, opened 2025-04-15
[DR] Diagnostics for examining how the time step is set


Maxwell-Rosen:
# Design Review

## Overview

Description: In many simulations, we are concerned with identifying which phenomena are setting the time step. It is valuable to look at the array of the CFL condition in each cell, as well as the omegaH mode. This diagnostic will output these two arrays.

Motivation: Diagnosing time step restrictions

Impact: A boolean flag in the input file `.output_dt_diagnostics = true` and writing the appropriate arrays in the code. The arrays are there; they need to be written out.

## Design Review Checklist 

Approval Criteria:

- [x] Is the design coherent and consistent with Gkeyll's existing architecture?
- [x] Does the design align with user-facing API standards?
- [x] Do functionalities overlap with existing features?
- [x] Are the algorithms robust and appropriate for the intended purpose?

Issue #686, opened 2025-04-15
[DR] Turning off updates of cells with small values of distribution function


Maxwell-Rosen:
# Design Review

## Overview

Description: I aim to implement a specific input file level flag that turns off cell updates in cells with a distribution function smaller than that value. This should help with positivity issues and save on the DT set by regions of small distribution function.

Motivation: This feature is used in the SR/GR parts of the code we want to use in GK. It's valuable in the mirror, where many cells have a tiny distribution function.

Impact: The GK kernels must be modified with some appropriate flags. There will be statements inside them that look like the following:
```
  if (fabs(fin[0]) < input_min_val) {
    return 0.0;
  }
```
The kernels which must be modified are gyrokinetic_boundary_surfvpar_*, gyrokinetic_boundary_surfx_* (y,z), gyrokinetic_surf* (vpar, x,y,z), gyrokinetic_vol_*

## Proposed Design

Key Design Elements: At the input file, the gk_species object will have a flag for the lowest value of f[0], which is the 0th moment of the cell. This can be different for each species. Since it's phase space information, it should go in the species object. This value gets fed through to the kernels.

Prototyping Efforts (if applicable): branch phase-cell-update-turn-off

## Design Review Checklist 

Approval Criteria:

- [x] Is the design coherent and consistent with Gkeyll's existing architecture?
- [x] Does the design align with user-facing API standards?
- [x] Do functionalities overlap with existing features?
- [x] Are the algorithms robust and appropriate for the intended purpose?

## Additional Notes

To test whether this changes the dynamics of the problem, I implemented it crudely in 1x. I took a steady-state Boltzmann electron enhanced collision WHAM simulation and turned on this feature with cell updates. The steady-state did not change, so this feature will not dramatically impact the dynamics while reducing computational cost. This ran for 5 ion collision times, 500 microseconds.

![Image](https://github.com/user-attachments/assets/2dd5f049-1fbf-4052-bfe3-27d7b92bc06e)

Along with this direction, I'm interested in how we can reduce the total number of cell updates and how this impacts performance. I didn't notice my simulations getting faster due to this change. It's still a good feature to include for stability.

PR #694, opened 2025-04-17
Turn off cell updates when f[0]<val

Maxwell-Rosen:
# Feature

## Summary

Purpose: Turn off updates in cells with small density. This relaxes the CFL constraint without affecting dynamics dramatically

DR link: #686 #687 

## Implementation Details

Key changes: Only affects the dg_gyrokinetic modules. The flag is passed through the struct. Each species can set a different flag

I combined a CFL diagnostic implementation into this PR because it's necessary to make the plots below.

Automated testing: There is already a unit test for these files

## Example Use

```
  .skip_cell_if_smaller_than = 1e-16,
}
```

## Community Standards

## Testing:

This unit test takes forever
build/unit/ctest_fem_poisson_perp
Test test_2x_p1_periodicx...    

## Additional Notes

Here is an example use case from the WHAM 1x2v regression test

image


Maxwell-Rosen:
Here is an example where I'm skipping cells for the electrons, but not for the ions. I removed the const from that parameter. In principle, one might be interested in changing it through the course of the simulation, but I wanted to make sure the cell updates aren't changing this

image

Maxwell-Rosen:
It's working on 1 GPU on stellar as well

image

apps/gk_species.c
```diff
@@ -1156,6 +1189,10 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
   else
     gks->apply_pos_shift_func = gk_species_apply_pos_shift_disabled;
   gks->write_func = gk_species_write_dynamic;
+  if (gks->info.cfl_dt_diangostic)
```
JunoRavin, line 1192: typo in the name of the diagnostic.

JunoRavin (approved):

zero/dg_gyrokinetic.c
```diff
@@ -48,13 +48,13 @@ gkyl_gyrokinetic_set_auxfields(const struct gkyl_dg_eqn *eqn, struct gkyl_dg_gyr
 struct gkyl_dg_eqn*
 gkyl_dg_gyrokinetic_new(const struct gkyl_basis *cbasis, const struct gkyl_basis *pbasis,
   const struct gkyl_range *conf_range, const struct gkyl_range *phase_range, 
-  const double charge, const double mass, enum gkyl_gkmodel_id gkmodel_id, const struct gk_geometry *gk_geom,
-  const struct gkyl_velocity_map *vel_map, bool use_gpu)
+  const double charge, const double mass, double skip_if_smaller_than, enum gkyl_gkmodel_id gkmodel_id,
```
JonathanGorard, line 51: I'm not a fan of the variable name `skip_if_smaller_than`, particularly given that this quantity subsequently appears as part of logical comparisons such as `skip_if_smaller_than > 0.0` (which feels needlessly difficult to parse). `skip_threshold`, `skip_minimum`, `skip_floor` (or, in fact, those should probably be `evolve_minimum` or `evolve_floor`) would all feel less cumbersome, and more informative, here.

JunoRavin: I like skip_threshold. That's a good name because I'm using threshold now elsewhere in the code for defining partial moments.

Maxwell-Rosen: Sure. Threshold sounds like a good name for this

Maxwell-Rosen: My thinking was that in the input file, it reads kind of like a sentence. Skip_cell_update_if_f0_smaller_than = 1e-16

apps/gkyl_gyrokinetic.h
```diff
@@ -228,6 +228,7 @@ struct gkyl_gyrokinetic_species {
 
   enum gkyl_gkmodel_id gkmodel_id;
   double charge, mass; // Charge and mass.
+  double skip_cell_threshold; // Skip updates over cells where 0th DG component is smaller than this value
```
manauref, line 231: The user doesn't know the DG normalization, so this shouldn't be a DG component threshold, but a f value threshold. Please update the comment and code accordingly.

apps/gkyl_gyrokinetic.h
```diff
@@ -260,6 +261,7 @@ struct gkyl_gyrokinetic_species {
   int num_integrated_diag_moments; // Number of integrated diagnostic moments.
   char integrated_diag_moments[24][24]; // List of integrated diagnostic moments.
   bool time_rate_diagnostics; // Whether to ouput df/dt diagnostics.
+  bool cfl_dt_diagnostic; // Whether to ouput dt diagnostics such as CFL and omegaH.
```
manauref, line 264: omegaH ? I see no changes related to omegaH in this PR. If it's just for writing omega_cfl, then maybe just call it `write_omega_cfl`

regression/rt_gk_wham_1x2v_p1.c
```diff
@@ -607,6 +607,7 @@ int main(int argc, char **argv)
     .name = "elc",
     .charge = ctx.qe,
     .mass = ctx.me,
+    .skip_cell_threshold = 1e-16,
```
manauref, line 610: Add a comment describing what this is here

apps/gk_species.c
```diff
@@ -837,6 +869,7 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
 
   // Allocate cflrate (scalar array).
   gks->cflrate = mkarr(app->use_gpu, 1, gks->local_ext.volume);
+  gks->cflrate_ho = mkarr(false, 1, gks->local_ext.volume);
```
manauref, line 872: this should depend on whether the user requested the diagnostic, don't always allocate it

also, whenever you are allocating host (or device) twins of another array it is best to get the number of components and size from its twin, like
`gks->cflrate_ho = mkarr(false, gks->cflrate->ncomp, gks->cflrate->size);`

apps/gk_species.c
```diff
@@ -1156,6 +1189,10 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
   else
     gks->apply_pos_shift_func = gk_species_apply_pos_shift_disabled;
   gks->write_func = gk_species_write_dynamic;
+  if (gks->info.cfl_dt_diagnostic)
+    gks->write_cfl_func = gk_species_write_cfl_dynamic;
+  else 
+    gks->write_cfl_func = gk_species_write_cfl_static;
```
manauref, line 1195: instead of _dynamic and _static, can you use _enabled and _disabled ? I want to start moving to this language to reserve dynamic and static specifically for the distinction between dynamic and static species (or fields)

apps/gk_species.c
```diff
@@ -1893,8 +1931,10 @@ gk_species_write(gkyl_gyrokinetic_app* app, struct gk_species *gks, double tm, i
   if (frame == 0) {
     gk_species_write_dynamic(app, gks, tm, frame);
   }
-  else
+  else {
     gks->write_func(app, gks, tm, frame);
+    gks->write_cfl_func(app, gks, tm, frame);
```
manauref, line 1936: please put this inside write_func

zero/gkyl_dg_gyrokinetic_priv.h
```diff
@@ -78,8 +80,13 @@ static double
 kernel_gyrokinetic_vol_1x1v_ser_p1(const struct gkyl_dg_eqn *eqn, const double* xc, const double* dx, 
   const int* idx, const double* qIn, double* GKYL_RESTRICT qRhsOut)
 {
+  
   struct dg_gyrokinetic *gyrokinetic = container_of(eqn, struct dg_gyrokinetic, eqn);
 
+  if (gyrokinetic->skip_cells && fabs(qIn[0]) < gyrokinetic->skip_cells_mag) {
```
manauref, line 86: @JunoRavin rather than checking these two conditions, isn't better to:
1. remove `skip_cells`.
2. set the default value of `skip_cells_mag` to -DBL_MIN, or change it to the user-provided value if they give one?

unit/ctest_dg_gyrokinetic.c
```diff
@@ -105,7 +105,7 @@ test_dg_gyrokinetic()
     phaseRange, phaseRange_ext, velLocal, velLocal_ext, false);
 
   struct gkyl_dg_eqn* eqn = gkyl_dg_gyrokinetic_new(&confBasis, &basis, &confRange, &phaseRange, 
-    charge, mass, 0, gk_geom, gvm, false);
+    charge, mass, 0, 0, gk_geom, gvm, false);
```
manauref, line 108: should this be -DBL_MIN instead of 0? because before we were updating cells even if f<0, and I believe that, although not great, it is needed for conservation.

manauref (changes requested):
This is great Max. Aside from the small things I pointed out, there are three things that would be good for you to do:
1. After your last change, check unit tests and (several) GK tests are valgrind clean. Also check that a couple of the reg tests you didn't touch produce the same results as before, I suggest gk_step_out_2x2v, gk_sheath_3x2v, and gk_ltx_boltz_elc_1x2v.
2. Ensure that f<0 cells are still updated if the user didn't provide a threshold.
3. I think you should do some more quantification of what this capability is doing. You show some mirror profiles in the DR, but we should at the least do some conservation tests to understand (and maybe quantify) how badly this is breaking conservation, and perhaps how it depends on the v-extents (presumably there's a balance between large v-extents and the threshold f). I can share some conservation balance scripts.

There are 2 other things that come to mind not addressed here:
1. LBO collisions: these use integrals of f around the v-space boundaries. But if some cells are not updated out there, should those cells also be skipped in the boundary correction calculation?
2. Same thing for sources.
3. Same thing for boundary flux (I think this is handled by the if-statement in the boundary_surf function, but it'd be good to check, and it'll probably show up in conservation balance calculations).

Thinking about this further, some of my above comments lead to the following question: when f<threshold, which parts of the GK equation do we want to turn off, and why?

unit/ctest_dg_gyrokinetic.c
JunoRavin: No I think this is fine because when you create the equation object you check whether this input parameter is defined, so we can always pass 0 here.

zero/gkyl_dg_gyrokinetic_priv.h
JunoRavin: Yea this is true, we could set skip_cells_mag to some reference value (even just -1.0 since we're checking if fabs(f) < skip_cells_mag). 

So basically when the new method is called we check if skip_cells_threshold > 0.0, if it is, we set it to that, and if it's not, we set it to -1.0. We could set it to -DBL_MIN, but I was noticing recently that some files that need to include float.h to have this defined are not doing so and this is affecting newer CUDA builds like on my workstation. I don't think we need to use -DBL_MIN and I worry about missing a float.h inclusion until we have a better handle on why some systems seem to require float.h to be included and others don't (since we're not going to go and just include float.h everywhere for now).

apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Specifically, the code outputs Jf in the -ion_#.gkyl files. We compare the 0th component of this expansion of Jf to the `skip_cell_threshold`. Are you suggesting that we change how it is used? How would you explain this to a user?

apps/gkyl_gyrokinetic.h
Maxwell-Rosen: I had difficulty in writing omega_cfl because it's not stored as an array. It's a double which is incrimented every time something is larger than it. Outputing a single double in a whole file seems excessive, so I didn't do that

zero/gkyl_dg_gyrokinetic_priv.h
Maxwell-Rosen: Jimmy's above comment mentioned wanting a boolean to turn this feature on and off. I agree that it's redundant and we can store this information inside the double skip_cells_mag. Doubles are initialized to zero, but this could be machine dependent, so we can initialize this to -1 when it's not specified for clarity

Maxwell-Rosen:
Valgrind sanitized
```
(pgkyl) (phase-cell-update-turn-off) gkylzero: valgrind ./build/regression/rt_gk_wham_1x2v_p1 -s1
==1171076== Memcheck, a memory error detector
==1171076== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==1171076== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==1171076== Command: ./build/regression/rt_gk_wham_1x2v_p1 -s1
==1171076== 
Taking time-step at t = 0 ... dt = 1.70428e-11

Number of update calls 1
Number of forward-Euler calls 5
Number of RK stage-2 failures 1
Max rel dt diff for RK stage-2 failures 1.388
Min rel dt diff for RK stage-2 failures 1.388
Species RHS calc took 2.63978 secs
Species collisions RHS calc took 12.6372 secs
Field RHS calc took 0.0506605 secs
Species collisional moments took 0.69549 secs
Updates took 17.2747 secs
Number of write calls 6,
IO time took 0.0197107 secs 
==1171076== 
==1171076== HEAP SUMMARY:
==1171076==     in use at exit: 0 bytes in 0 blocks
==1171076==   total heap usage: 2,488 allocs, 2,488 frees, 73,006,195 bytes allocated
==1171076== 
==1171076== All heap blocks were freed -- no leaks are possible
==1171076== 
==1171076== For lists of detected and suppressed errors, rerun with: -s
==1171076== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

Maxwell-Rosen:
To examine the effect of this feature on energy and particle conservation, in these figures, I examine the difference between simulations without these features and simulations when turning off phase space cell updates. We use `gk_sheath_1x2v_p1` as a test case. First, I ran `pgkyl gk_sheath_1x2v_p1 interp info` to look for the maximum value of the distribution function at t=0 for each species. Using this value, I multiplied it by a threshold (1e-3, 1e-5, 1e-8, 1e-10), the same value for both electrons and ions. To examine the effect of this feature, we take the difference between the simulation without these effects and the simulation with these effects. Turning off phase space cell updates affects the time step, so we linearly interpolate the integrated diagnostics from the simulation without this feature onto the time grid of each simulation with this feature. We take the absolute fractional difference between the compared simulations in all these plots.

Examining these figures, we see reasonable agreement in $M_0$ for all threshold values, but there are significant deviations in $M_2$ for low threshold values. $M_0$ tends to oscillate around the value of the simulation without this feature. This is because turning off phase space cell updates effectively pulls the distribution function advection in regions of small f. However, there is significant drift in electron $M_2$ for the threshold of 1e-3. 

Doubt may be cast on the threshold values because the maximum of the distribution function changes in time. Since this is set in the input file based on t=0, this is not a representative value for all times. Overall, M0 is decreasing rapidly at the beginning of the simulation.

Overall, the cyan line of 1e-5 demonstrates reasonable agreement over long periods of time with respect to simulations without this feature. This is a reasonable value to run this test. Future investigation will look at the appropriate threshold for mirror simulations.

particle_balance_m0_wrt_main_elc
particle_balance_m0_wrt_main_ion
particle_balance_m2_wrt_main_elc
particle_balance_m2_wrt_main_ion

Maxwell-Rosen:
I ran a similar test with a WHAM production test. I took an adiabatic electron simulation with enhanced collision frequency (2,000x) run to steady state. We use this initial condition and turn off phase space cell updates in regions where f is small. I ran this for 10 microseconds, which is 1-2 ion transit times and we look for if the integrated M0 or M2 change dramatically. The only difference between these simulations is changing the threshold which updates are turned off under. We see that the difference with respect to the normal simulation approaches a steady state value. They seem like very small differences, even in the case of 1e-3, which is a large threshold.

particle_balance_m0_wrt_main_ion
particle_balance_m2_wrt_main_ion

Maxwell-Rosen:
To confirm that this is real, this is a plot comparing main (blue) to the 1e-3 threshold. Moments are top left M0, top right M1, bottom left M2par, bottom right M2perp. The simulation hasn't reached steady state, but there is an offset it seems to be approaching.

image

Maxwell-Rosen:
In rt_gk_sheath, we also are interested in how well the simulation conserves energy. Although it seems like particles and energy are accounted for in all cases, this feature raises the noise level for errors. The errors oscillate around zero, suggesting that energy conservation holds, but has a pulsed nature.

energy_balance_Energy_change
energy_balance_thresh-1e-3_energy
energy_balance_thresh-1e-5_energy
energy_balance_thresh-1e-8_energy
energy_balance_thresh-1e-10_energy
energy_balance_on_main_energy
particle_balance_particle_change
particle_balance_thresh-1e-3_particle
particle_balance_thresh-1e-5_particle
particle_balance_thresh-1e-8_particle
particle_balance_thresh-1e-10_particle
particle_balance_on_main_particle

zero/gkyl_dg_gyrokinetic.h
```diff
@@ -36,7 +36,7 @@ struct gkyl_dg_gyrokinetic_auxfields {
  */
 struct gkyl_dg_eqn* gkyl_dg_gyrokinetic_new(const struct gkyl_basis *cbasis, const struct gkyl_basis *pbasis, 
   const struct gkyl_range *conf_range, const struct gkyl_range *phase_range, 
-  const double charge, const double mass, enum gkyl_gkmodel_id gkmodel_id,
+  const double charge, const double mass, const double skip_cell_threshold, enum gkyl_gkmodel_id gkmodel_id,
```
manauref, line 39: add to comment with description of this new input. Also, remove `const`

zero/gkyl_dg_lbo_gyrokinetic_diff.h
```diff
@@ -23,13 +23,15 @@ struct gkyl_dg_lbo_gyrokinetic_diff_auxfields {
  * @param conf_range Configuration space range for use in indexing primitive moments
  * @param pgrid Phase-space grid object.
  * @param mass Species mass.
+ * @param skip_cell_threshold Threshold which under skips cells.
```
manauref, line 26: shouid be "Threshold under which to skips cells." ?

zero/gkyl_dg_lbo_gyrokinetic_drag.h
```diff
@@ -23,13 +23,14 @@ struct gkyl_dg_lbo_gyrokinetic_drag_auxfields {
  * @param conf_range Configuration space range for use in indexing primitive moments
  * @param pgrid Phase-space grid object.
  * @param mass Species mass
+ * @param skip_cell_threshold Threshold which under skips cells
```
manauref, line 26: "Threshold under which to skips cells" ?

zero/gkyl_dg_updater_gyrokinetic.h
```diff
@@ -38,7 +38,7 @@ struct gkyl_dg_updater_gyrokinetic_tm {
 gkyl_dg_updater_gyrokinetic* gkyl_dg_updater_gyrokinetic_new(const struct gkyl_rect_grid *grid, 
   const struct gkyl_basis *cbasis, const struct gkyl_basis *pbasis, 
   const struct gkyl_range *conf_range, const struct gkyl_range *phase_range,
-  const bool *is_zero_flux_bc, double charge, double mass, enum gkyl_gkmodel_id gkmodel_id,
+  const bool *is_zero_flux_bc, double charge, double mass, const double skip_cell_threshold, enum gkyl_gkmodel_id gkmodel_id,
```
manauref, line 41: add comment describing new input. Also, remove `const`

zero/gkyl_dg_updater_lbo_gyrokinetic.h
```diff
@@ -37,7 +37,7 @@ struct gkyl_dg_updater_collisions*
 gkyl_dg_updater_lbo_gyrokinetic_new(const struct gkyl_rect_grid *phase_grid,
   const struct gkyl_basis *conf_basis, const struct gkyl_basis *phase_basis, const struct gkyl_range *conf_range,
   struct gkyl_dg_lbo_gyrokinetic_drag_auxfields *drag_inp, struct gkyl_dg_lbo_gyrokinetic_diff_auxfields *diff_inp, 
-  double mass, const struct gk_geometry *gk_geom, const struct gkyl_velocity_map *vel_map,
+  double mass, double skip_cell_threshold, const struct gk_geometry *gk_geom, const struct gkyl_velocity_map *vel_map,
```
manauref, line 40: add to comment with description of new input

zero/boundary_flux.c
```diff
@@ -67,15 +72,21 @@ gkyl_boundary_flux_advance(gkyl_boundary_flux *up,
     long linidx_g = gkyl_range_idx(&up->ghost_r, idx_g); 
     long linidx_s = gkyl_range_idx(&up->skin_r, idx_s);
 
+    const double *fIn_s = gkyl_array_cfetch(fIn, linidx_s);
+    const double *fIn_g = gkyl_array_cfetch(fIn, linidx_g);
+
+    if (fabs(fIn_s[0]) < up->skip_cell_threshold && fabs(fIn_g[0]) < up->skip_cell_threshold)
+      continue;
```
manauref, line 79: why is this, and its GPU counterpart, a continue?

unit/ctest_position_map.c
```diff
@@ -441,7 +441,7 @@ test_position_polynomial_map_optimize_1x()
   pos_map->bmag_ctx->crange_global = &localRange;
   pos_map->bmag_ctx->cbasis = &basis;
   pos_map->bmag_ctx->cgrid = &grid;
-  gkyl_array_copy(pos_map->bmag_ctx->bmag, bmag_global);
+  gkyl_array_set(pos_map->bmag_ctx->bmag, 1.0, bmag_global);
```
manauref, line 444: i think this is tied up in another PR, so is this what you still want to do here or do you want to correct it here too?

apps/gkyl_gyrokinetic_priv.h
```diff
@@ -627,6 +627,7 @@ struct gk_species {
 
   struct gkyl_array *f, *f1, *fnew; // arrays for updates
   struct gkyl_array *cflrate; // CFL rate in each cell
+  struct gkyl_array *cflrate_ho; // host copy for use in IO and initialization
```
manauref, line 630: i don't think this has anything to do with initialization.

apps/gkyl_gyrokinetic.h
```diff
@@ -228,6 +228,7 @@ struct gkyl_gyrokinetic_species {
 
   enum gkyl_gkmodel_id gkmodel_id;
   double charge, mass; // Charge and mass.
+  double skip_cell_threshold; // Skip updates over cells where 0th DG component of Jf is smaller than this value. Jf is what is output in the -species_#.gkyl files
```
manauref, line 231: I think I mentioned this elsewhere, or maybe I mentioned it in Slack, but I think this comment, and the meaning of `skip_cell_threshold` should be changed so that it doesn't involve the 0th DG coefficient and instead the cell-averaged Jf. The former requires the user to know the relationship between the 0th coefficient and the cell-av Jf (i.e. the normalization), but in general they don't. So it's best to say it'll skip cells when Jf_cell-av < skip_cell_threshold, and under the hood rescale `skip_cell_threshold` by the appropriate normalization factor.

apps/gk_species.c
```diff
@@ -702,6 +735,9 @@ gk_species_release_dynamic(const gkyl_gyrokinetic_app* app, const struct gk_spec
   gkyl_array_release(s->bc_buffer_lo_fixed);
   gkyl_array_release(s->bc_buffer_up_fixed);
 
+  if (s->write_cfl_func) {
```
manauref, line 738: use `s->info.write_omega_cfl` instead

manauref (changes requested):
just a few minor mods requested.

Also, can you please do this to dg_diffusion_gyrokinetic? it seems like it's needed there too/

zero/boundary_flux.c
Maxwell-Rosen: It's because this call is inside a while loop, so to skip these cells, we continue the loop. You make a good point that I didn't realize though. There should be something that is set to zero here.

Maxwell-Rosen:
Code compiles without errors on Stellar-amd GPU build

apps/gkyl_gyrokinetic.h
manauref: change comment since you changed the normalization

PR #702, opened 2025-04-24
Install scripts for Snellius supercomputer

SjoerdPR:
Added mkdeps and configuration scripts and an example jobscript for the Snellius supercomputer in the Netherlands.

PS This is my first time contributing to a project on Github, so if something went wrong or I'm not doing things by the book, please let me know so I do not make the same mistakes in the future.


JunoRavin (approved):
Thanks for making these scripts! This all looks good to me. Glad Gkeyll can be useful to you!

Antoinehoff (approved):

PR #706, opened 2025-05-02
Fix memory issue in position map

Maxwell-Rosen:
On my laptops, the position map doesn't detect the extrema appropriately. This fixes that by changing where the outer limits of the array are assigned.

### DIFF
--- a/apps/gyrokinetic.c
+++ b/apps/gyrokinetic.c
@@ -388,7 +388,7 @@ gkyl_gyrokinetic_app_new_geom(struct gkyl_gk *gk)
 
   app->bmag_ref = (bmag_max_global + bmag_min_global)/2.0;
 
-  gkyl_position_map_set(app->position_map, app->gk_geom->mc2nu_pos);
+  gkyl_position_map_set_mc2nu(app->position_map, app->gk_geom->mc2nu_pos);
 
   // If we are on the gpu, copy from host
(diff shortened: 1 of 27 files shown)


manauref:
did you tested on the cluster and/or GPUs to ensure it behaves as you expect?

manauref (approved):

ammarhakim:
Do not remove the const keyword. Why is this being removed? Please fix the underlying issue before you do that. Basis functions should be passed as const to prevent the object from being changed.

ammarhakim:
I am closing this PR till the issue of const is properly resolved. I don't understand why changes are made to core APIs to accommodate test cases. First thing to ask is: is the const in the struct incorrectly specified? Should it be there or not? Once this is resolved we can reopen the PR.

Maxwell-Rosen:
I discussed the structure of this object with @ammarhakim, and we agreed that the original design on main is not true to the functional programming paradigm that we aim for. I updated the objects to have a more "functional programming" design, where object attributes are not manipulated outside the structure. We will revisit this design in the future because it is not clear why we need to do the entire geometry in order to get just the magnetic field.

ammarhakim (approved):
This is much better. We will need to do another refactor once the geometry is properly generated but for now this is ok to merge. Further debugging can occur then.

PR #707, opened 2025-05-02
Update the normalization of u_par error.

dingyunl:
Normalize the error in u_par by thermal velocity when u_par is smaller than thermal velocity.
Fix a bug in computing the relative u_par error when u_par is larger than thermal velocity: the denominator should be the absolute value of u_par instead of u_par itself since it can be negative.


JunoRavin (approved):
This looks good @dingyunl. I realize we still need better safe guards external to this updater to insure the target moments like density and temperature are not negative anywhere (especially now that you're taking a square root of the target temperature), but I think we are checking this elsewhere and if the target temperature is negative, this updater can't correct anything anyway because the Maxwellian isn't projected correctly.

manauref (approved):

Issue #708, opened 2025-05-05
Boundary Flux Diagnostics Fail 2x Sims in x-direction


akashukla:
The translate dim operation required for boundary flux diagnostics in 2x simulations throws an assertion failure when trying to use diagnostics at the x-boundaries. This can be seen by replacing the GKYL_SPECIES_ZERO_FLUX boundary conditions with GKYL_SPECIES_ABSORB in regression/rt_gk_sheath_2x2v_p1.c.

`rt_gk_sheath_2x2v_p1: zero/translate_dim.c:48: void gkyl_translate_dim_advance(gkyl_translate_dim *, const struct gkyl_range *, const struct gkyl_range *, const struct gkyl_array *restrict, int, struct gkyl_array *restrict): Assertion rng_do->upper[d] == rng_tar->upper[d] failed.`

With ZERO_FLUX BC's in the x-direction, the diagnostics work fine because in that case the diagnostics are only called for the z direction.

manauref:
not an issue anymore

PR #710, opened 2025-05-07
Fix assert check in translate_dim updater to address issue #708.

manauref:
see issue #708

### DIFF
--- a/zero/translate_dim.c
+++ b/zero/translate_dim.c
@@ -43,9 +43,22 @@ gkyl_translate_dim_advance(gkyl_translate_dim* up,
   struct gkyl_array *GKYL_RESTRICT ftar)
 {
   // Perform some basic checks:
-  for (int d=0; d<up->cdim_do-1; d++) {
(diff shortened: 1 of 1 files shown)


Antoinehoff:
Can you explain in a few lines what the new code does?

Maxwell-Rosen:
Does this bug appear in the regression test? Can you make sure that the regression test is updated so that we test for this bug in the future?

manauref:
@Antoinehoff 
the assert that was there before was old, and didn't correctly check for consistency between ranges, including support for all these potential translations: https://github.com/ammarhakim/gkylzero/blob/6648f3eb6d47d337e42e340e22722bebf284deea/zero/gkyl_translate_dim.h#L15-L27
Specifically, it failed for translating a 2D field down to 1D.

@Maxwell-Rosen 
We didn't have a 2D regression test that incurred in this feature. We'll have one soon when I standardize conservation checks for all sims, but that's going to happen in a separate PR.

JunoRavin (approved):

akashukla (approved):

PR #713, opened 2025-05-13
Change size of inflated grid in the ignored direction so that 2x mapc…

manauref:
…2p geo comes out correct.


JunoRavin (approved):

akashukla:
I observed that this fixed plotting of the nodes for D3D mapc2p geo. To be clear, this only affects the output mapc2p and nodes for plotting, not the metric coefficients.

manauref:
Yes that's correct.

Issue #703, opened 2025-04-24
Bug when dividing by Jacobians inside of Boltzmann electron field kernels, but also seems to be somewhere else too


Maxwell-Rosen:
In #668, we modified the way that the Jacobian is divided out in the sheath potential and ambipolar potential evaluation by using `gkyl_dg_div_op_range` in `gk_field.c` rather than doing the division inside the phi calculation. The reason for this is showcased in the plot below.

![Image](https://github.com/user-attachments/assets/5d1581e3-6715-4de7-b7c1-5e555863bca0)

In the image above, we see that there are little ears on phi which do not appear in the density. Since $\phi = \phi_{sheath} + \ln(\frac{n_i}{n_{i,sheath}})$, how can this be? It's because there are some limiters in the mapc2nu_pos.

![Image](https://github.com/user-attachments/assets/9643990c-b301-4d01-909b-ebc7b5524825)

Near z=0, the midplane of the mirror, we don't want the cells to get too small, so we put a limiter there. This applies a straight line through the non-uniform grid. However, this introduces a discontinuity in the first derivative near the transition. This transition is carried forward into the Jacobian, causing numerical errors in the division.

By moving the division by J outside the potential calculation and into `gk_field.c`, we removed the bumps that plagued our fields.

![Image](https://github.com/user-attachments/assets/ae84fa48-cb8b-4835-b925-a896a5852bcc)

This teaches us that there is something wrong with doing the division inside our kernels, and we should use the official `gkyl_dg_div_op_range` when dividing Jacobians out of quantities.

### Where else might this be causing problems?

If one looks inside the gyrokinetic kernels, there are all sorts of divisions by Jacobians that are going on. This issue could be in more places than just the potential kernels, but has only been brought up by the non-uniform grid work.

I ran a 1x2v mirror problem for 1 microsecond, and we see signs of this, even after fixing the division by Jacobian issue.

Field
![Image](https://github.com/user-attachments/assets/d4e22a51-0d43-4e80-ba11-8efcbb9a5434)
Ion density
![Image](https://github.com/user-attachments/assets/4a6daebf-f28c-445a-a7ca-33fe64a881dd)

In the figures above, we observe errors in both the ion density and potential, indicating that this issue likely originates from a bug elsewhere in the code, perhaps in the gyrokinetic equations themselves.

### Lesson

We must be careful when dividing DG fields so as not to introduce numerical errors.

### Solution

[...]

PR #717, opened 2025-05-16
Jacobtot was not being weak divided in boltzmann electron sheath calculations

Maxwell-Rosen:
# Bug fix

## Summary

Description: in #668, we noticed that the sheath potential was wrong and needed this factor of $C/JB$. In the same PR, we noticed issues that occurred due to weak division of $JM_0 / J \neq M_0$. We also introduced another $1/J$ and still used the jacobtot_inv variable even though we knew the issues with that. This PR performs a weak division of cmag and jacobgeo_tot instead of doing this inside the kernels.

Issue link: #703 

## Solution

Geometry is divided in the app, then passed through the boltzmann electron struct to the sheath calculation

Impacted files: gk_field.c, boltzmann electron kernels, ambibolt_potential.c

Automated testing: Leaky_bag tests are instrumental in verifying the sheath potential. These tests pass, so the sheath potential is very close to the analytic value it should have.

Lesson: Be very cautious when doing mul_op in kernels with inverse quantities. Make sure that you keep in mind whether these are divided at nodal values or weak divided.

## Community Standards

## Testing:

- [x] I added a unit test for this feature.

## Additional Notes

Associated gkylcas PR:


manauref:
@Maxwell-Rosen what is this fixing or improving? can you provide some evidence?

Maxwell-Rosen:
It does not improve the quality of the simulation. It's improving the clarity to readers of the code that jacobgeo_inv is not to be used outside of the gk kernels because it is not weakly equal to 1/jacobgeo. In this case, since the sheath is evaluated at the node, this is exact and it's okay to do this here, but in general using jacobgeo_inv is a bad idea.

JunoRavin:
@Maxwell-Rosen This distinction will matter if this updater's higher dimensional generalizations are ever utilized, as then one would need to be careful to insure the desired weak equality properties along the surface.

manauref:
If there's no evidence that this fixes or improves anything, I'm inclined to close this PR without merging.

Furthermore: the code you are changing here is currently consistently with how the gyrokinetic equation is discretized (for both electrons and ions). If you wanted to change this:
1. It should be well supported because it's foundational
2. I would encourage you to investigate its repercussions in the actual kinetic equation not just the sheath potential, e.g. if you make the same change in the kinetic equation, does anything change/improve/get worse?

JunoRavin:
The fact that we do this in the gyrokinetic equation is technically wrong and the motivation for things like the quadgeo branch and different volume vs. surface expansions. We are only forced to do this in the gyrokinetic equation because continuity of things like the inverse Jacobian (using the current Gauss-Lobatto nodal evaluations) is more important to our current implementation in main than the aliasing error caused by 1/J*J =/= 1 from the nodal evaluations. 

Max's point is this sort of aliasing can and does matter or you would not have changed the sheath potential definition in the previous DR Max is pointing to. 

In general we might not expect this sort of change to have a significant impact because one hopes that this C/JB factor is 1 for the way we currently construct geometry and the differences from 1.0 produced by nodal evaluations vs. alias-free weak operations, but should we ever switch to coordinate systems where C is not 1.0, then this difference between a nodal operation and weak division may matter, especially in higher dimensions.

manauref:
I'm not saying this sort of change doesn't matter. I'm saying:
1) there has to be a stronger basis for merging it, in the form of data
evidence or mathematical arguments (with equations) to support it.
2) because this is based on how the GK equation is discretized (and there's
actually a reason for using jacobgeotot_inv rather than cmag/jacongeotot),
it should be investigated in the GK equation before changing it here.

Issue #711, opened 2025-05-07
Centralization of the last closed flux surface position and index through the gk app


Antoinehoff:
Many places in the gk app use the position or index of the cell that abuts the last closed flux surface (LCFS) from below using :
```
  double xLCFS = gk_s->lower_bc[d].aux_parameter;
  int idxLCFS = (xLCFS-1e-8 - app->grid.lower[0])/app->grid.dx[0]+1;
```
or
```
  double xLCFS = f->info.xLCFS;
  int idxLCFS_m = (xLCFS-1e-8 - app->grid.lower[0])/app->grid.dx[0]+1;
```
From the examples above, we see that the LCFS informations are stored through the position along x, `xLCFS`, which is an attribute of many structures, here the boundary structure and in the field parameters.

This issue wants to redesign this, storing the LCFS position in only one place, either the `app` structure or the `geometry`, as well as the below LCFS cell index `idxLCFS` to avoid inconsistency and code duplication.

This will yield a PR soon. The changes are minors but multiple and the user interface will have to change in many input parameters...

manauref:
For context, historically this parameter was put in a place like gk_species because it was only used by the twis-shift BC updater only. As the model becomes more sophisticated it has began using it in more places, so it makes sense to centralize its definition (for safety / ease of maintainability).

PR #712, opened 2025-05-08
Replace strings specifying moment (of a distribution function) types with enum options 

manauref:
This goes with [PR 78 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/78), and addresses issue #652 .

In this work we replace the specification of diagnostic moments that was previously done with strings, so that now it is done with enum options. The enum (and corresponding strings) are the following (located in zero/gkyl_mom_type.h):
```
enum gkyl_distribution_moments {
  GKYL_F_MOMENT_M0 = 0, // Number density.
  GKYL_F_MOMENT_M1, // Momentum density.
  GKYL_F_MOMENT_M2, // Kinetic energy density.
  GKYL_F_MOMENT_M2PAR, // Parallel kinetic energy density.
  GKYL_F_MOMENT_M2PERP, // Perpendicular kinetic energy density.
  GKYL_F_MOMENT_M2IJ, // Kinetic energy tensor..
  GKYL_F_MOMENT_M3, // Heat flux.
  GKYL_F_MOMENT_M3PAR, // Parallel energy flux.
  GKYL_F_MOMENT_M3PERP, // Perpendicular energy flux.
  GKYL_F_MOMENT_M3IJK, // Heat flux in lab frame.
  GKYL_F_MOMENT_MAXWELLIAN, // M0, drift speed, T/m.
  GKYL_F_MOMENT_BIMAXWELLIAN, // M0, drift speed, Tpar/m, Tperp/m.
  GKYL_F_MOMENT_LTE, // Maxwellian or Maxwell-Juttner moments.
  GKYL_F_MOMENT_M0M1M2,  // M0, M1, M2.
  GKYL_F_MOMENT_M0M1M2PARM2PERP,  // M0, M1, M2par, M2perp.
  GKYL_F_MOMENT_HAMILTONIAN,  // M0, mass*M1, H moments.
  GKYL_F_MOMENT_M1_FROM_H, // dH/dv / m moment.
  GKYL_F_MOMENT_ENERGY, // H moment.
  GKYL_F_MOMENT_M0ENERGYM3, // M0, Energy (H) and M3 moments.
  GKYL_F_MOMENT_NI, // M0, M1i for-vector.
  GKYL_F_MOMENT_TIJ, // Stress-energy tensor.
};
```

I ran make unit and everything passes. Also, make regression compiles all regression tests successfully. Lastly, I visually checked the final moments and integrated moments of 
```
[...]

PR #718, opened 2025-05-16
Correct and centralize creation of core/SOL ranges in IWL GK sims

manauref:
Some of the ranges in GK IWL clopen simulations were wrong (off by 1). For sure the global core range is wrong, and it's likely the extended core and sol ranges were also wrong.

Here we correct that.

Also, to avoid numerous definitions of LCFS related quantities (i.e. the index of the cell to its left, and the core and sol ranges) throughout the app (i.e. in GK app, in species, field, sources, bflux), and thus having an increased risk for making a mistake, we centralize the location of the LCFS parameters (x_LCFS and idx_LCFS_lo) to gk_geometry, and the ranges are only created in one place (in gyrokinetic for conf- and in species for phase-space). This way other modules can simply fetch them from these places rather than creating them again. This addreses issue #711 .

Here's proof that the ranges are correct now, done with rt_gk_d3d_iwl_3x2v_p1:
```
x_LCFS was not at a cell boundary. Moved to: 9.375000000e-02

conf grid: 0:0.15,-0.0707482:0.0707482,-3.14159:3.14159 | cells= 8, 4, 8

x_LCFS = 0.09375 | idx_LCFS_lo =  5

app->global      =  1: 8, 1: 4, 1: 8
app->global_core =  1: 5, 1: 4, 1: 8
app->global_sol  =  6: 8, 1: 4, 1: 8

app->global_ext      =  0: 9, 0: 5, 0: 9
app->global_ext_core =  0: 5, 0: 5, 0: 9
app->global_ext_sol  =  6: 9, 0: 5, 0: 9

app->lower_skin_par_core  =  1: 5, 1: 4, 1: 1
app->lower_skin_par_sol   =  6: 8, 1: 4, 1: 1
app->lower_ghost_par_core =  1: 5, 1: 4, 0: 0
app->lower_ghost_par_sol  =  6: 8, 1: 4, 0: 0

app->upper_skin_par_core  =  1: 5, 1: 4, 8: 8
app->upper_skin_par_sol   =  6: 8, 1: 4, 8: 8
app->upper_ghost_par_core =  1: 5, 1: 4, 9: 9
app->upper_ghost_par_sol  =  6: 8, 1: 4, 9: 9

app->local_par_ext_core =  1: 5, 1: 4, 0: 9
```

### DIFF
--- a/apps/gk_species.c
+++ b/apps/gk_species.c
@@ -928,13 +928,6 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
       // applying sheath BCs in the SOL.
       gks->periodic_dirs[gks->num_periodic_dir] = app->cdim-1; // The last direction is the parallel one.
       gks->num_periodic_dir += 1;
-      // Check that the LCFS is the same on both BCs and that it's on a cell boundary within our grid.
-      double xLCFS = app->gk_geom->x_LCFS;
-      assert(fabs(xLCFS-gks->upper_bc[dir].aux_parameter) < 1e-14);
-      // Check the split happens within the domain and at a cell boundary.
-      assert((app->grid.lower[0]<xLCFS) && (xLCFS<app->grid.upper[0]));
-      double needint = (xLCFS-app->grid.lower[0])/app->grid.dx[0];
-      assert(floor(fabs(needint-floor(needint))) < 1.);
     }
   }
   
@@ -1015,7 +1008,7 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
     else if (gks->lower_bc[d].type == GKYL_SPECIES_GK_IWL) {
       double xLCFS = app->gk_geom->x_LCFS;
       // Index of the cell that abuts the xLCFS from below.
-      int idxLCFS_m = (xLCFS-1e-8 - app->grid.lower[0])/app->grid.dx[0]+1;
+      int idxLCFS_m = app->gk_geom->idx_LCFS_lo;
       gkyl_range_shorten_from_below(&gks->lower_skin_par_sol, &gks->lower_skin[d], 0, app->grid.cells[0]-idxLCFS_m+1);
       gkyl_range_shorten_from_below(&gks->lower_ghost_par_sol, &gks->lower_ghost[d], 0, app->grid.cells[0]-idxLCFS_m+1);
 
@@ -1092,7 +1085,7 @@ gk_species_new_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *
     else if (gks->upper_bc[d].type == GKYL_SPECIES_GK_IWL) {
       double xLCFS = app->gk_geom->x_LCFS;
       // Index of the cell that abuts the xLCFS from below.
-      int idxLCFS_m = (xLCFS-1e-8 - app->grid.lower[0])/app->grid.dx[0]+1;
+      int idxLCFS_m = app->gk_geom->idx_LCFS_lo;
       gkyl_range_shorten_from_below(&gks->upper_skin_par_sol, &gks->upper_skin[d], 0, app->grid.cells[0]-idxLCFS_m+1);
       gkyl_range_shorten_from_below(&gks->upper_ghost_par_sol, &gks->upper_ghost[d], 0, app->grid.cells[0]-idxLCFS_m+1);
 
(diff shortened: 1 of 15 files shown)


manauref:
> Can you set the has_lcfs boolean to default to false? That way that flag doesn't need to be included in every geometry struct.

I think it's set to false by default. Did you see a case in which it wasn't?

tnbernard:
I did not see or test anything related to the current PR. I previously saw a case specific to neutral boundary conditions. So just a note of caution.

JunoRavin:
So long as we keep doing things like this:
struct gkyl_gk_geometry_inp geometry_inp = {
...
}
in gyrokinetic.c

and then making the app side struct by value (i.e., being careful to not malloc the app input struct because with @manauref's PR on changing to moment enums, https://github.com/ammarhakim/gkylzero/pull/712, we can still put the app input to the struct on the stack), then @tnbernard this value will default to false safely. 

We will need to make sure when inner-wall-limited capabilities are added to the gyrokinetic Lua wrappers, this boolean is set to false, but @JonathanGorard and I have tried to be careful on the Lua side to set *every* value of the input struct so that we don't have to worry about this.

JunoRavin (approved):

tnbernard (approved):

apps/gk_species.c
```diff
@@ -1691,6 +1659,50 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
...
+    int len_core = idx_LCFS_lo;
+    int len_sol = gks->global.upper[0]-len_core;
+    gkyl_range_shorten_from_above(&gks->global_core, &gks->global, 0, len_core);
+    gkyl_range_shorten_from_below(&gks->global_sol , &gks->global, 0, len_sol);
+    // Same for local ranges.
+    gkyl_range_shorten_from_above(&gks->local_core , &gks->local , 0, len_core);
+    gkyl_range_shorten_from_below(&gks->local_sol  , &gks->local , 0, len_sol);
+
+    int len_core_ext = idx_LCFS_lo+1;
+    int len_sol_ext = gks->global_ext.upper[0]-len_core;
```
Antoinehoff, line 1674: Are you sure about that?
I would have said that it's 
```
int len_sol_ext = gks->global_ext.upper[0]-(len_core + 1);
```
because the extended ranges

apps/gk_species.c
```diff
@@ -1691,6 +1659,50 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
...
+    for (int e=0; e<2; e++) {
+      gkyl_range_shorten_from_above(e==0? &gks->lower_skin_par_core  : &gks->upper_skin_par_core,
+                                    e==0? &gks->lower_skin[par_dir]  : &gks->upper_skin[par_dir], 0, len_core);
+      gkyl_range_shorten_from_above(e==0? &gks->lower_ghost_par_core : &gks->upper_ghost_par_core,
+                                    e==0? &gks->lower_ghost[par_dir] : &gks->upper_ghost[par_dir], 0, len_core);
+      gkyl_range_shorten_from_below(e==0? &gks->lower_skin_par_sol   : &gks->upper_skin_par_sol,
+                                    e==0? &gks->lower_skin[par_dir]  : &gks->upper_skin[par_dir], 0, len_sol);
+      gkyl_range_shorten_from_below(e==0? &gks->lower_ghost_par_sol  : &gks->upper_ghost_par_sol,
+                                    e==0? &gks->lower_ghost[par_dir] : &gks->upper_ghost[par_dir], 0, len_sol);
+    }
```
Antoinehoff, line 1692: Here I am also a bit doubtful. If you do the drawing for the ghost ranges the length is increased by one I think.
0-1|2|3||4|5|6|7|8|9|10-11
above '-' delimits ghosts, '||' shows the separatrix.
shorten from above:
    ghost 0 : 11
    idxLCFS_m = 3
    ghost_core_up[0] = ghost_lo[0]+len_core-1;
                      = global_lo[0]+len_core-1;
                      = 0 + 3 - 1 = 3
    ghost_core = 0:2
which does not work right?

Antoinehoff (changes requested):
Since it is changing indices, we should be sure it is valgrind clean also.

apps/gk_species.c
manauref: yes, I'm sure

apps/gk_species.c
manauref: see the example I just posted in the PR message. I think all the numbers are correct/as expected now

Antoinehoff (approved):
great job

PR #720, opened 2025-05-17
Add missing header

manauref:


JunoRavin:
@Antoinehoff already got this fix in https://github.com/ammarhakim/gkylzero/pull/719

PR #724, opened 2025-05-20
GPU compiling fix for CUDA 12.9

Maxwell-Rosen:
Under advice from @JunoRavin, we made a few modifications to get the GPU code to compile. I'm running an Ubuntu 24.04 laptop with an external GPU RTX 3090 with CUDA 12.9. Many unit tests are passing, so the build is successful. Newer versions of CUDA moved around a few flags that we need to shift to be able to compile on newer versions.

### DIFF
--- a/zero/array_reduce_cu.cu
+++ b/zero/array_reduce_cu.cu
@@ -9,6 +9,7 @@ extern "C" {
 #include <gkyl_array_reduce.h>
 #include <gkyl_array_reduce_priv.h>
 #include <gkyl_basis.h>
+#include <float.h>
 }
 
 // CUDA does not natively support atomics for MAX and MIN on doubles (addition/sum is fine).
--- a/zero/array_reduce_cu.cu
+++ b/zero/array_reduce_cu.cu
@@ -9,7 +9,6 @@ extern "C" {
 #include <gkyl_array_reduce.h>
 #include <gkyl_array_reduce_priv.h>
 #include <gkyl_basis.h>
-#include <float.h>
 }
 
 // CUDA does not natively support atomics for MAX and MIN on doubles (addition/sum is fine).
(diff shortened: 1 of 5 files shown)


zero/array_reduce_cu.cu
```diff
@@ -9,6 +9,7 @@ extern "C" {
 #include <gkyl_array_reduce.h>
 #include <gkyl_array_reduce_priv.h>
 #include <gkyl_basis.h>
+#include <float.h>
```
JunoRavin, line 12: just put this header in the gkyl_array_reduce.h and gkyl_array_dg_reduce.h

JunoRavin (approved):

manauref:
@Maxwell-Rosen did you check that this multib_sync unit test passes on multiple CPUs or GPUs?

Maxwell-Rosen:
I did not check this, but from the changes made, I have no reason to suspect that these would have issues with multiple CPUs or GPUs.

JunoRavin:
@Maxwell-Rosen you only need 2 GPUs for 5 minutes to check the unit test. That's all we're asking to check. As you say, by eye there looks to be no reason for NCCL to have any issue with your commit here, and it's always safe to include standard libraries like float.h, but if Stellar or Perlmutter has an issue with even more changes like this we need to know it now before we break everyone's production capabilities.

manauref:
You should run the unit test, preferably on 2 and 4 GPUs. You can just ask for 2/4 GPUs on Perlmutter or Stellar interactively for 5 min to do this.

Maxwell-Rosen:
Perlmutter helped me find an error I made in the unit test. There is one test that fails on both main and this branch, so it's independent of this work, but it should be looked at to see why it's failing. Line 632 is failing.

PR #726, opened 2025-05-21
Fixes to boundary flux diag moms and enable conservation diagnostics for GK IWL sims

manauref:
We are trying to establish (i.e. assess, and fix if needed) particle and energy conservation in GK IWL clopen simulations, in 2x at first and hopefully 3x after.

Here we:
- Restrict the ranges over which z boundary fluxes are computed in IWL sims so they are only calculated in the SOL (assuming fluxes through the twist-shift boundary are conservative).
- Fix a small error that prevented (non vol integrated) boundary flux diagnostics from being written out correctly for 2x and 3x if running in parallel.
- Enable the conservation diagnostics for IWL sims.
- Change the treatment of x_LCFS in geo so that it demands that it is at a cell boundary and doesn't shift it (because it has to be consistent with the user's input file).

Via these changes we have been able to check conservation in 2x and 3x sims using rt_gk_d3d_iwl as a testbed.

Here's rt_gk_d3d_iwl_2x:
Screenshot 2025-05-21 at 9 44 17 AM
particle balance looks good. Energy balance
Screenshot 2025-05-21 at 9 47 46 AM
Looks ok. The figure on the left doesn't inspire confidence because the error is as large as the boundary flux losses, but in relative to the energy content and the time step (right figure) it seems comparable or even better than turbulent LAPD sims.

Unfortunately rt_gk_d3d_3x (with resolution increased to 24x16x12x10x6 because it's unstable at lower res) doesn't conserve particles:
Screenshot 2025-05-21 at 11 26 25 AM
Some possibilities we could explore to fix this (in a separate branch):
- The way we are enforcing twist-shift-dicity of phi or the biasing at the limiter corner could make the numerical fluxes discontinuous.
- Fluxes through the twist-shift boundary are not actually continuous (although I tried this sim with a periodic instead of a twist-shift core, and got similar results at lower res).

### DIFF
--- a/apps/gk_species_bflux.c
+++ b/apps/gk_species_bflux.c
@@ -458,7 +458,7 @@ gk_species_bflux_write_mom_dynamic(gkyl_gyrokinetic_app* app, void *spec_in,
         if (app->cdim > 1) {
           // Project the moment down to lower dimensions.
           int num_mom_comp = bflux->moms_op[mom_idx].num_mom;
-          gkyl_translate_dim_advance(bflux->transdim[b], bflux->boundaries_conf_skin[b], &bflux->surf_local[dir],
+          gkyl_translate_dim_advance(bflux->transdim[b], bflux->boundaries_conf_skin_fullx[b], &bflux->surf_local[dir],
             bflux->moms_op[mom_idx].marr, num_mom_comp, bflux->mom_surf[b*num_diag_mom+m]);
 
           if (app->use_gpu)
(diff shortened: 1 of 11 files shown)


apps/gk_species_source.c
```diff
@@ -273,12 +273,10 @@ gk_species_source_calc(gkyl_gyrokinetic_app *app, const struct gk_species *s,
   struct gk_source *src, double tm)
 {
   if (src->source_id) {
-    struct gkyl_array *source_tmp = mkarr(app->use_gpu, s->basis.num_basis, s->local_ext.volume);
     for (int k=0; k<s->info.source.num_sources; k++) {
-      gk_species_projection_calc(app, s, &src->proj_source[k], source_tmp, tm);
-      gkyl_array_accumulate(src->source, 1., source_tmp);
+      gk_species_projection_calc(app, s, &src->proj_source[k], s->fnew, tm);
+      gkyl_array_accumulate(src->source, 1., s->fnew);
```
JunoRavin, line 278: This re-use of memory seems suspicious. For example, say you had a general time-dependent source and you needed to call gk_species_source_calc before the RHS accumulation. Then this re-use of memory would overwrite the value stored in fnew, which would mess up the RK combine in the SSP RK3 method. 

I do not think we support this functionality right now (everywhere to my knowledge we utilize a Maxwellian of some form at t=0 and then we rescale it, or even in the adaptive source case we utilize the LTE infrastructure). But I worry about this re-use of memory affecting someone's ability to utilize a general time dependent source term. 

The safest phase space array to re-use is the distribution function array in the LTE object because that *always* gets cleared before you do anything with it like project a Maxwellian. So if you need a dummy phase space array, we could just utilize the distribution function array in the LTE object.

apps/gk_species_bflux.c
```diff
@@ -431,35 +431,36 @@ gk_species_bflux_write_mom_dynamic(gkyl_gyrokinetic_app* app, void *spec_in,
     int dir = bflux->boundaries_dir[b];
     int edi = bflux->boundaries_edge[b]==GKYL_LOWER_EDGE? 0 : 1;
 
-    if ((edi == 0 && rank == 0) || (edi == 1 && rank == comm_size-1)) {
+    if (dir < app->cdim-1 || ((edi == 0 && rank == 0) || (edi == 1 && rank == comm_size-1))) {
```
JunoRavin, line 434: I think we should add a comment about this if statement about the assumption we're making here just to help ourselves further down the line. We are assuming if you have a non-periodic direction in the x direction, you own all of the x direction and the only decomposition is occurring in z. 

As a follow-up, I would then ask if this conditional works the way you expect for multi-block simulations, or at you not using boundary flux diagnostics in multi-block simulations with core blocks yet?

manauref: i'll add a comment. It's only because we are only decomposing along z.

it should work for multiblock. Each block outputs its own boundary fluxes, even if those fluxes are into another block.

apps/gk_species_source.c
manauref: done. Used the LTE memory instead

JunoRavin (approved):

Antoinehoff:
I think we should run compute-sanitizer on this (:

manauref:
> I think we should run compute-sanitizer on this (:

on what, specifically?

apps/gk_species_source.c
Antoinehoff: I will have to change that again in PR #678 because LTE memory is used in the projection step...

PR #729, opened 2025-05-23
Place checks on nu and M2 in LBO primitive moment kernels to avoid creating singular matrices

manauref:
This goes with [PR 82 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/82). I copy the same message from that PR here:

Recently (https://github.com/ammarhakim/gkylzero/pull/705) we incorporated the collision frequency into the primitive moment linear problem in order to conserve momentum and energy when using a spatial and time dependent collision frequency $\nu(x,t)$. So now, the self primitive moments are computed via solving the system

$\nu ~ u_{\parallel} ~ M_0 - \nu ~ v_t^2 ~ c_M = \nu ~ M_1$

$\nu ~ u_{\parallel} ~ M_1 + \nu ~ v_t^2 ~ (3 M_0 - c_E)  = \nu ~ M_2$

where $c_M$ and $c_E$ are boundary corrections.

But note that there are protections on $\nu(x,t)$ to avoid problematic values. The collision frequency $\nu(x,t) \propto n$, so when we detect that $n\leq0$ is are a quadrature point, $\nu(x,t)$ is set to zero there. This can make $\nu(x,t)=0$ in a cell, which would make the LHS matrix of the system shown above become singular. This wasn't apparent until recently when more normNu sims were done.

To address this we change the primitive moment kernels so that so that when $\nu(x,t)^{(0)}$ and $M_2^{(0)}$ (the cell average coefficients of $\nu$ and $M_2$ are greater than 0, it continues as before. But if either of these is not positive, when we set the entries corresponding to the LHS quantities in the system above to 1. This is just so that the matrix is not singular and our linear solvers don't error out. The resulting primitive moments $u_{\parallel}$ and $v_t^2$ will be meaningless in this cell, but that's ok because LBO kernels are wrapped in an if-statement that turns them off if $M_2^{(0)}$ is not positive.

A similar conditional was placed on the cross-primitive moment linear system. Additionally, in that system we were switching some quantities to use cell-averages when the density is zero at a cell corner, but we were not switching $u_{\parallel s}$ and $u_{\parallel r}$ to use cell-averages. Change that here, so that in this case all quantities are used in a cell-averaged sense.

The signature of the PKPM lbo had to change because it uses common infrastructure, but the contents of their kernels didn't change.

---------------------------------
We have ran the cross-species thermal relaxation test with these new kernels, and reproduced the same behavior as before:

Screenshot 2025-05-23 at 4 01 50 PM

Another thing this fixes is the `gk_ar_react_nonuniformv_1x2v_p1` test, which was erroring out but we hadn't noticed.

I think this should make collisions a wee bit more robust, but only by using it more will we know for sure.


manauref:
For the record, if one were to multiply $nu_ee$ and $nu_ei$ by about $2/\pi^{3/2}$, one could match the FPO electron isotropization rate
Screenshot 2025-05-26 at 11 54 51 AM
but the electron-ion equilibration rate would be off. We DO NOT do this in this PR, we simply wanted to document it.

kernels/lbo/gyrokinetic_cross_prim_moments_1x1v_ser_p1.c
```diff
@@ -1,39 +1,45 @@
...
   double cMr[2] = {0.0}; 
   double cEr[2] = {0.0}; 
+  double u_selfr[2] = {0.0}; 
+  double u_otherr[2] = {0.0}; 
+ 
+  if (nu[0] > 0.0 && moms_self[4] > 0.0) { 
+  
+  // If a corner value is below zero, use cell average m0.
+  bool notCellAvg = true;
+  if (notCellAvg && (-(0.5*(2.4494897427831783*moms_self[1]-1.4142135623730951*moms_self[0])) < 0)) notCellAvg = false;
```
Antoinehoff, line 29: Isn't it equivalent and more efficient to write:
```
notCellAvg = notCellAvg && (-(0.5*(2.4494897427831783*moms_self[1]-1.4142135623730951*moms_self[0])) >= 0);
```
in all these lines ? This would avoid a logical.

manauref: I think you are right. It avoids a logical but incurs an assignment. Dunno which is more taxing. I can change it if you know avoiding the logical is better

JunoRavin: There won't be a difference performance-wise because the evaluation of the logical only costs a single cycle. It's just a question of what we internally view as clearer. I always prefer conditionals myself with if statements because I don't like to parse logical assignments, but that's just my preference (so I would leave it as is).

Antoinehoff: ok what convinced me is the fact that my option is always assigning, so I think in the worst case scenario, the cost of my proposition is less (maybe with a marginal benefit) but the original code avoids many assigns so yes ok it's better as it is.

Issue #733, opened 2025-05-27
[DR] Minor improvement in positivity shift


manauref:
The positivity shift (for GK and Vlasov) works as follows:
1. Loop over conf-space cells.
2. At a conf-space cell, loop over v-space.
3. In each cell check if $f=f_{old}<0$ at control nodes (currently Gauss-Legendre nodes). If it is negative, shift $f_{old}$ to 0, and effect a nodal-to-modal transformation to obtain the new $f=f_{new}$.
4. After looping over v-space, if a shift happened at this x-cell and the cell average of density of $f_{old}$ ($n_{old}$) was positive, rescal $f_{new} = (n_{old}/n_{new}) f_{new}$.

This last step can move particles around in v-space more than was necessary, altering the temperature considerably. 

We can make a small improvement that should reduce the translation of particles across v-space in step 4, by adding the following step between steps 3 and 4 above:
3.5 If the 0th moment of $f_{old}$ (proportional to $f_{old}^{(0)}$, the 0th DG coefficient) in this phase-space cell was positive, and a shift happened, rescale $f_{new}$ in this cell so that it maintains the same density: $f_{new} = (f_{old}^{(0)}/f_{new}^{(0)}) f_{new}$.
This will also add some intra-cell diffusion, but maintain the number of particles within this cell.

Of course, when n<0 in a cell to begin with, none of the rescalings above take place, and the positivity shift has to genuinely add new particles.

We will prototype this in the `gk_pos_shift_mod` branch

manauref:
addressed in PR #737

PR #737, opened 2025-05-29
Improve positivity shift operator to avoid unnecessary changes to f 

manauref:
This makes a small improvement in the positivity shift operator which should reduce the possibility of cells changing that don't need to be changed in cases where some node had negative f, but cell average f was positivity.

Current procedure (in main)
----------------------------
The current updater performs two steps. Consider a scenario with just 1 cell in x, and 2 cells in v. We'll assume f is independent of v, and that in the lower v cell f is negative at the lower Gauss-Legendre (GL) node.
The first step goes through all the v-space cells at this x and shifts f to 0 if at GL nodes if f<0 there:
Positivity_p1

In the second step, after having looped over all v-space, we check if the old M0 was >0. If yes, we scale f by M0_old/M0_new at all v-space cells to ensure M0 remains the same at this x:
Positivity_p1

As you can see this has the side effect of lowering f in an otherwise perfectly fine cell (the upper v cell). Compare the original f in red with the new f in blue.

Addition to procedure (this PR)
-------------------------------
In this work we add a new step intermediate step: if the current cell has an M0 contribution >0 (before the shift) and a shift happened, rescale the current cell to maintain its M0 contribution. 
Positivity_p2
Compared to the previous approach at the end of step 1 (green), the result of this new intermediate step (orange) is to produce an f that is nonegative at GL nodes but also has the same cell-average value as the original f, effectively adding some subcell diffusion and preserving its contribution to M0.

After this intermediate step we proceed with step 2:
Positivity_p2
which now doesn't do anything to the orange f.

In this scenario, the new approach:
1. Added some subcell diffusion in cell (i,j) but kept its cell average the same.
2. Avoided removal of particles in other v-space cells. In this case it avoided removal of high energy particles, which would've lowered the temperature (but the opposite can also happen, i.e., avoiding removal of low energy particles which would've increased the temperature)/

This work was described at a more higher level in DR #733

### DIFF
--- a/zero/positivity_shift_gyrokinetic.c
+++ b/zero/positivity_shift_gyrokinetic.c
@@ -116,8 +116,6 @@ gkyl_positivity_shift_gyrokinetic_advance(gkyl_positivity_shift_gyrokinetic* up,
       // Shift f if needed.
       bool shifted_node = up->kernels->shift(up->ffloor[0], distf_c);
 
-      // If m0phase_in_c was positive but one of the nodes of f was
-      // shifted, rescale f in this cell so it keeps the same density.
       if (shifted_node) {
         // Compute the new number density in this phase-space cell.
         double m0phase_out_c[num_cbasis];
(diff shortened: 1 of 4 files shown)


JunoRavin (approved):
I like this a lot and it seems strictly better to me to avoid global manipulations of f. Very, very cool.

Issue #715, opened 2025-05-15
[DR] Refactoring of Geometry Representation & Geometry Generation for Use in DG, including Gyrokinetics


ammarhakim:
Motivation
--------------

The geometry in the DG side of the code needs to refactored to allow a consistent discretization of surface fluxes. At present there is significant confusion on the various quantities used in the updates, specially in the gyrokinetic code, where there is confusion between surface and volume quantities, leading to potential inconsistencies. Further, we need the DG geometry in other parts of the code, for example, the dynamic (and in the future fluids) neutrals, fluid solvers in PKPM, Maxwell solvers for GR, etc.

Hence, here I am proposing a refactored design that would closely follow the successful design used in representing geometry in the finite-volume code.  Further, the second proposal is a refactor of the mesh generation code for GK, both for mirrors as well as for tokamaks. Both are described below.

Proposed Geometry Structure
----------------------------------------

Following the design of the FV code one of the outputs of the mesh generation will be a struct that stores the normals and two tangents at each surface quadrature node. Further, at that quadrature node we would also have the ratios of surface-area in physical space to computation space as this is needed in computing the surface flux over the complete cell. The struct would be something like this:

```C
struct gkyl_dg_surf_geom {
  double Jac; // Volume Jacobian 
  double lenr // ratio of face-area in phys to comp space
  double norm[3]; // Normal to face
  // tau1 X tau2 = norm are tangents to face
  double tau1[3];
  double tau2[3];
};
```
(In GK we do not actually need the tangents but we need it in other parts of the code, including fluid neutrals).

There would be one such struct for each surf quadrature node. The surface flux kernels would take the norm vector (and other geometric quantities) and compute the needed flux at that surface quadrature point. This would then be multiplied by `lenr` and convolved with surface basis to give the surface-flux contribution to the update.

Surface expansions of these quantities would also be needed for the volume term.

This needs (probably extensive) refactoring of the GK kernels to uses the quantities in this struct to compute the flux at a quadrature node.

Proposed Refactoring of Geometry Generation
--------------------------------------------------------------

Along with the above I am proposing to split the geometry generation into two phases (Phase 1a and 1b may be separable for tokamaks, but they are for mirrors).

Phase 1 is generic to all parts of the code. Phase 2 would be specific to GK, PKPM etc.

- Phase 1a: Generate the nodes in physical space that correspond to each cell. 
[...]

PR #739, opened 2025-05-30
Mirror geometry redone with new infrastructure for nodal geometry

Maxwell-Rosen:
# Feature

## Summary

Purpose: We aim to restructure the geometry in a more readable way that generalizes to taking the mirror to $R=0$. Based on the work performed in #735, these objects are fed through to the app and the geometry quantities are all calculated. There are ZERO explicit kernels used in generating the geometry now.

Issue link and design review: #715 

## Implementation Details

Key changes: gk_geometry_mirror.c is refactored. There are significant changes to that file. Geometry is generated in 3 steps. 
1.  The nodes are generated with only essential information. This gives the duals, tangents, Jacobian, magnetic field, and node locations.
2. The derived quantities are calculated and populated into 3x. 3x is necessary here because some vectors are converted to Cartesian coordinates, and that depends on the angle.
3. Nodal arrays are projected onto DG arrays

In parallel, each process generates a full geometry. At the end of geometry generation, it is split up into each sub-range for the local process.

Input files no longer take input of the efit file. Now, the gkyl efit reader must be run to generate the `_psi.gkyl` file and that is input into the grid generator.

Automated testing: Unit tests ctest_gk_geometry_mirror, ctest_mirror_geo_gen, and ctest_mirror_geo_dg test the appropriate functionality in each of the modified files. All mirror regression tests have been modified. The nozzle regression tests also use this geometry, so they are modified.

## Example Use
```
  struct gkyl_mirror_geo_grid_inp grid_inp = {
    .filename_psi = "data/unit/wham_hires.geqdsk_psi.gkyl", // psi file to use
    .rclose = 0.2, // closest R to region of interest
    .zmin = -2.0,  // Z of lower boundary
    .zmax =  2.0,  // Z of upper boundary 
    .include_axis = false, // Include R=0 axis in grid
    .fl_coord = GKYL_MIRROR_GRID_GEN_SQRT_PSI_CART_Z, // coordinate system for psi grid
  };
```

## Community Standards

## Testing:

## Additional Notes

Here is an example grid from the WHAM 2x2v regression test

Keep in mind that cmag is not 1 because of using the sqrt(psi) grid.
image

Here is the grid. No axis labels because `pgkyl showgrid` does not have the ability to label it's axes yet.
image

Future work will include the integration of the position map along the z direction, including the magnetic field-dependent geometry. This isn't difficult to implement, as the position map is designed to handle this case


manauref:
Also, it'd be good to show that particle and energy conservation is as good as before, or better.

manauref (changes requested):
@Maxwell-Rosen aside from checking conservation, can you also make sure that a couple of the regression tests and unit tests are `valgrind` (CPU) and `compute-sanitizer` clean? Antoine and I have found these to be very useful for unearthing hidden errors.

Maxwell-Rosen:
I ran the regression tests through `valgrind` and they were clean. I have not used `compute-sanitizer` with this branch

Maxwell-Rosen:
Here are plots in the energy conservation of rt_gk_wham_1x2v_p1. It seems like the new geometry improved the energy conservation, but ever so slightly.

image

image

ammarhakim:
This looks good. Thanks for running valgrind. Once @akashukla approves, @manauref can remove the block and I will merge. Please let us do this tomorrow COB by the latest. Compute sanitize is nice to but I do not believe it is needed to complete the merge.

manauref:
> I ran the regression tests through `valgrind` and they were clean. I have not used `compute-sanitizer` with this branch

you can use compute-sanitizer as follows:

```compute-sanitizer --tool memcheck --leak-check full <regression test executable> -g -s3```

manauref:
> This looks good. Thanks for running valgrind. Once @akashukla approves, @manauref can remove the block and I will merge. Please let us do this tomorrow COB by the latest. Compute sanitize is nice to but I do not believe it is needed to complete the merge.

Just as with valgrind on CPUs, compute-sanitizer helps us insure proper use of memory and to avoid spending days debugging later.

manauref:
> Here are plots in the energy conservation of rt_gk_wham_1x2v_p1. It seems like the new geometry improved the energy conservation, but ever so slightly.
> 
> image
> 
> image

This energy conservation check seems acceptable for this PR. I think the net error is too large, but in this branch it is comparable to main's, so it should be looked into independently of this PR (in main, or after this PR is merged).

Can you post particle conservation too please?

btw, you might want to check 2x, since that is the more important use case of these improvements.

akashukla (approved):

Maxwell-Rosen:
Here are particle conservation plots from this test. It's not machine precision, which suggests something is amiss with these regression tests, but it's comparable to main.

image

image

Maxwell-Rosen:
There were some errors in the input files and diagnostics for the earlier plots. The below plots show exact particle conservation for both ions and electrons. Energy is comparable. The regression test is updated to reflect these conservation plots.

image
image


Electron particle balance
image
image

Ion particle balance
image
image

Maxwell-Rosen:
I ran compute sanitizer on rt_gk_wham_1x2v_p1 and rt_gk_wham_2x2v_p1 and they are clean

manauref (approved):
thanks for performing checks, good work Max

PR #753, opened 2025-06-09
Optional additional arguments for C input files

manauref:
This adds an `opt_args` option to `regression/rt_arg_parse.h` similar to that in `gkyl.c`, so that anyone can pass additional command line arguments to C input files (e.g. to do launch batches of simulations).

For example, one could launch a simulation that externally sets the input power, which is proportional to the particle source rate `ndot_src` like 
```
./sim_executable -o ndot_src=1.2e22
```
requesting the source rate $\dot{n}_{src}=1.2\times10^{22} ~ \mathrm{m}^{-3}/\mathrm{s}$. Inside the input file this could be read and stored in the context as follows:
```
  struct my_sim_ctx ctx = create_ctx(); // context for init functions

  // Extract variables from command line arguments.
  sscanf(app_args.opt_args, "ndot_src=%lf", &ctx.ndot_src);
```

### DIFF
--- a/regression/rt_arg_parse.h
+++ b/regression/rt_arg_parse.h
@@ -31,6 +31,7 @@ struct gkyl_app_args {
   bool skip_limiters; // should we skip limiters?
   bool is_restart; // is this a restarted sim?
   int restart_frame; // frame to restart from
+  char opt_args[1024]; // optional arguments
 };
 
 static int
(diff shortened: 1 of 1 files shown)


manauref:
I thought about that @ammarhakim , but freeing at the end would require adding a free-ing statement to every regression test/input file. So I'd prefer not to, unless you had something more practical in mind.

ammarhakim:
One issue is what we ran into before: the stack space is limited and so 1024 bytes may make make us go over the limit in complex C drivers. But I guess it’s ok for now.As a middle ground check that the opt arg input is not more that 1024 so at least we don’t run risk of someone passing a huge path.Sent from my iPhoneOn Jun 9, 2025, at 12:39 AM, Manaure Francisquez ***@***.***> wrote:﻿manauref left a comment (ammarhakim/gkylzero#753)
I thought about that @ammarhakim , but freeing at the end would require adding a free-ing statement to every regression test/input file. So I'd prefer not to, unless you had something more practical in mind.

manauref:
ok, true, that's a good check.

Antoinehoff:
Btw this max run time limit is very useful for long production simulations on HPC clusters where one wants to use the total job allocation time but also would like to terminate the run instead of just being killed by the slurm scheduler. I still think that it could be worth considering including this feature directly in Gkeyll as it is for the `-n` option that allows us to set a number of steps before exiting.

JunoRavin (approved):

Issue #730, opened 2025-05-24
[DR] Add support for NN training and acceleration for moment computations in PKPM


JonathanGorard:
This work has been prototyped within the `kann_integration_prototyping` branch: https://github.com/ammarhakim/gkylzero/tree/kann_integration_prototyping

The intention here is to add support within our input file structure for the continuous training of neural networks for the acceleration of moment computations within the PKPM app (with the hopes that an equivalent API can also be extended to Vlasov and Gyrokinetics). We use the highly minimal `kann` (https://github.com/attractivechaos/kann) C framework for performing autodiff, and for representing and training our neural networks. One near-term objective is to construct fully magnetized closures for use in future 10-moment simulations by training a neural network on PKPM simulation data to predict the heat fluxes `q_par` and `q_perp` (and their derivatives) based on the densities and pressures `rho`, `p_par` and `p_perp` (and their derivatives). Since we intend to train directly on the DG expansions themselves, the configuration space derivatives of all of the necessary quantities will already exist within the neural network, without needing to be computed within the 10-moment simulation itself. We also want to be able to test a neural network that has already been trained, to determine whether it is able to predict the PKPM moments given the existing simulation data.

The proposed additions to the PKPM input file structure are:

- `train_nn` / `trainNN`: should we be training a neural network on this simulation data? This will be `false` by default, to prevent the slowing down of simulations, but users should be encouraged to set it to `true` whenever resources permit. When set to `true`, a separate neural network will be trained on each particle species with a given frequency.
- `train_ab_initio` / `trainAbInitio`: should we construct and train a new neural network from scratch (ab initio), or should we load in an existing neural network and continue training on that? This will be `true` by default, since we cannot assume that a user has any saved neural networks on their machine, but in practice anyone using this framework will likely want to set it to `false` to ensure that their training efforts compound cumulatively with each simulation that they run.
- `nn_width` / `NNWidth`: for an ab initio training run, how wide should the neural network be (i.e. how many neurons should be contained in each layer)? This will be 256 by default, to provide enough degrees of freedom to learn all of the necessary DG coefficients.
- `nn_depth` / `NNDepth`: for an ab initio training run, how deep should the neural network be (i.e. how many hidden layers should there be)? This will be 5 by default, to provide sufficient depth to learn moderately complex relationships between the input and output data, without also being prohibitively expensive.
[...]

PR #760, opened 2025-06-14
First pass at neural network (KANN) integration

JonathanGorard:
This PR complements Design Review #730. Further API design details can be found there.

In this branch, we integrate the lightweight `kann` neural network framework (https://github.com/attractivechaos/kann) into Gkeyll, as part of `minus`. We add the ability to perform continuous online learning of diagnostic moments within PKPM simulations, via the following simple API, included as part of the input file:

```
-- Training parameters.
train_nn = false -- Train neural network on simulation data?
train_ab_initio = true -- Train neural network ab initio?
nn_width = 256 -- Number of neurons to use per layer.
nn_depth = 5 -- Number of layers to use.
train_nn_file = "pkpm_ot_p1_moms_nn_1" -- File path of neural network to train.
num_trains = GKYL_MAX_INT -- Number of times to train neural network.
num_nn_writes = 1 -- Number of times to write out neural network.
input_moms = { 1, 3, 4 } -- Array of "input" moments to train on.
output_moms = { 5, 6 } -- Array of "output" moments to train on.
test_nn = false -- Test neural network on simulation data?
test_nn_file = "pkpm_ot_p1_moms_nn_1" -- File path of neural network to test.
num_tests = 1 -- Number of times to test neural network.
```

The user is able to specify any collection of "input" and "output" moments to train on; can opt to train a new neural network of any width and depth ("ab initio"), or to load in and update an existing one; can specify any frequency with which to train the network (every time step, by default) and write the updated network to disk; and can opt to test any pre-trained neural network, by outputting the "true" diagnostic moments and the "predicted" diagnostic moments side-by-side.

Once an appropriate neural network has been trained on PKPM diagnostic moments, it can then be deployed to provide heat flux closure relations as part of 10-moment simulations via the following API:

```
-- Neural network parameters.
use_nn_closure = true -- Use neural network-based closure?
poly_order = 1 -- Polynomial order of learned DG coefficients.
nn_closure_file = "data/neural_nets/pkpm_ot_p1_moms_nn_1" -- File path of neural network to use.
```

The design of the neural network closures in the 10-moment solver directly mirrors that of the existing gradient-based closure. To compensate for the fact that this makes the input for the 10-moment equation object somewhat more unwieldy (since there is a new boolean for `has_nn_closure`, a new integer for `poly_order`, and a new pointer to a `kann_t` struct for the neural network itself), we have added a new input struct for 10-moments, just as we have in most other equation objects. The neural network closure works in both 1D (with both p1 and p2 PKPM simulation data) and 2D (with p1 simulation data only) for now. In the new neural network closure tests that have been added, namely rt_10m_riem_nn_closure_p1 and rt_10m_riem_nn_closure_p2 (1D Riemann problems, p1 and p2) and rt_10m_ot_nn_closure (2D Orszag-Tang vortex, p1), the results using pretrained neural networks match the PKPM simulation results match more closely than with either local or gradient-based closure, at the expense of some additional oscillation. We have added some examples of pretrained neural networks to the `data/neural_nets` directory. Right now there is some odd discrepancy, at least in 2D, between the Lua and C versions of the neural network-based closure tests. I have tried to track this down for the last couple of days, but can't seem to find the issue (even the input structs for all of the app fields look identical). I will look into this in more detail, but for now it is not a showstopper.

Eventually the hope is to extend the same design to perform continuous online learning of diagnostic moments in gyrokinetic and Vlasov simulations, too (e.g. to learn better equation of state relations for general relativistic hydrodynamics using general relativistic kinetic simulation data).

All regression tests build and run on both Mac OS, Linux CPU (stellar-intel) and Linux GPU (stellar-amd). No new unit tests have been added, and the existing 10-moment unit test has been updated to reflect the new API.


JonathanGorard:
> This is fantastic @JonathanGorard. Quick big picture question on the API side: it looks like the network training inputs are defined at the app level, but then we loop over species and train a network per-species?
> 
> It makes sense to me that these should be app-level inputs partially because whether or not you're going to train/test is in the public API (so C input files directly call these functions, and the Lua wrappers wrap these calls).
> 
> I would anticipate that such structure of a per-species network is inevitable right now when we don't know exactly how much we can encode in the network, but what does a simulation do at the moment if a user generates an electron closure, but then runs an electron-ion simulation? Do we have guards in place here? Or am I missing something about how the species name selects what closure network to use?

This is a good point that you raise - right now, if the user runs a simulation where the set of input species don't match the set of species for which NNs were trained, the simulation will segfault during initialization, which arguably isn't the kind of user-friendly behavior that we want. I'd be happy to add some guardrails to check whether such a species/naming mismatch has occurred, so that we can switch off the NN-based closure (and print a warning message) when it does. Unless you think we should be doing something more sophisticated?

JunoRavin (approved):

PR #763, opened 2025-06-18
Restart with a different geometry

Antoinehoff:
This PR wants to relax the assumptions when performing a restart from file using the structure
```
.init_from_file = {
  .type = GKYL_IC_IMPORT_F,
  .file_name = "restart-ion.gkyl",
  .jacobgeo_inv_file_name = "restart-jacobgeo_inv.gkyl",
},
```
The current state of g0 prevents any restart if the bounds of the grid are changed. It could however be useful to start a simulation in a new geometry with a turbulent state that was previously obtained. E.g. when changing from negative triangularity (NT) to positive triangularity (PT), the simulation domain length in the binormal coordinate is changed because it is normalized to the safety factor which changes between both configurations.

This PR simply remove the assert that impeach the restart with different grid bounds. A restart can now be done with any kind of file that has the same resolution. 
We also use this occasion to replace the `jacobgeo` scaling by a scaling by `jacobtot`. This scaling may be important if the metric changes and one should provide a `jacobgeo_inv` file within the `init_from_file` structure.

One has to pay attention to one major caveat:
The integrated moments may not be conserved as they will change with the change of the grid volume. One could implement a scaling of the distribution function to maintain the number of particle constant.

Tell me in the comments below ⬇️  if you'd like to see me implementing this and don't forget to subscribe and hit the bell 🤑

### DIFF
--- a/apps/gk_neut_species.c
+++ b/apps/gk_neut_species.c
@@ -750,10 +750,6 @@ gk_neut_species_file_import_init(struct gkyl_gyrokinetic_app *app, struct gk_neu
 
     // Perform some basic checks.
     if (pdim_do == pdim) {
-      for (int d=0; d<pdim; d++) {
-        assert(grid_do.lower[d] == grid.lower[d]);
-        assert(grid_do.upper[d] == grid.upper[d]);
-      }
       // Check if the grid resolution is the same.
       for (int d=0; d<pdim; d++)
         same_res = same_res && (grid_do.cells[d] == grid.cells[d]);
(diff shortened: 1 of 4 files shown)


manauref:
This seems good to me, but
- @Antoinehoff could you post some plots (maybe radial profiles?) of what the plasma looks like at the end of the first simulation with geometry A vs. what it looks like at the beginning of the simulation with geometry B?
- @Maxwell-Rosen could you double check the 1x -> 2x mirror import to make sure it is well behaved?
please?

JunoRavin (approved):

manauref:
Some 1D profiles of the above data:
Screenshot 2025-06-19 at 12 14 43 PM

Screenshot 2025-06-19 at 12 17 56 PM

Screenshot 2025-06-19 at 12 19 27 PM

Antoinehoff:
I think the plots above are mislabeled, the PT is the new geometry.

PR #764, opened 2025-06-18
Tensor Prototyping

johnson452:
Grouped commits together from what they were on the original gminus branch, but overall clarifies the history and really just eliminates a bunch of commits which are just for GPU commit fixes.

---- PR ----
### Summary
This PR includes the basics of the new tensor object, including unit tests, and is working on GPUs. The tensor fields are wrapped over the array to inherit many of the routines for copying, accessing, etc. We can construct tensor fields of ranks (limited to 7 currently), with size (number of points that the tensor exists at), and ndim number of dimensions in each direction.

### Creating a Tensor Field
Creating a tensor object requires a rank, dim, size, and an object of enum which can be up to the maximum rank (GKYL_MAX_DIM)"
```
/**
 * Create new tensor field. Delete using gkyl_tensor_field_release method.
 * 
 * @param rank Rank of the tensor field
 * @param ndim Number of dimensions
 * @param size Number of indices 
 * @param iloc Enum array of size GKYL_MAX_DIM which for lower or upper indices
 * @return Pointer to newly allocated tensor field.
 */
struct gkyl_tensor_field *gkyl_tensor_field_new(size_t rank, size_t ndim, size_t size, const enum gkyl_tensor_index_loc *iloc);

```

Where the enum corresponds to either raised or lowered indices.
```
enum gkyl_tensor_index_loc {
  GKYL_TENSOR_INDEX_LOWER,
  GKYL_TENSOR_INDEX_UPPER, 
};
```
These get modified during raising and lowering operations to keep track of the upstair and downstairs indices. 

### In place and set operations
Functions for modifying the tensors have raising or set methods for doing the operation in place or setting the value to another tensor field.

### Other
CPU Tests Pass
GPU Tests Pass
Valgrind clean (For CPU tests)


JonathanGorard (approved):
Nicely done!

PR #765, opened 2025-06-20
Relativistic multi-fluids and first-order Einstein prototyping work

JonathanGorard:
This branch consists largely of prototyping work towards a production-grade relativistic multi-fluid solver, as well as a production-grade solver for Einstein's equations in first-order, flux-conservative form. Neither of these solvers are ready for prime time yet, but since this work does not touch any other aspects of the Gkeyll infrastructure (except some very small additions to `wave_prop`), I think it is safer to merge this work now in its nascent state than to wait until after the directory reorg. I've added some new equation objects that make it easier to experiment with fully explicit and IMEX integrators for the moment-EM coupling source terms in curved spacetimes; these objects may eventually be deleted, once we have completed a production implementation. I have been validating these implementations against the special relativistic magnetohydrodynamic shock tube test of Balsara (https://iopscience.iop.org/article/10.1086/318941/fulltext/37994.text.html), in which it performs well even with fully explicit source terms, as well as against an electron-proton two-fluid version of the BHL accretion problem onto static and spinning black holes, for which the simulation destabilizes spontaneously at around t = 4.0 due to the accumulation of divergence errors in the EM field. Further prototyping work is required to experiment with hyperbolic divergence cleaning methods for the general relativistic Maxwell equations, which remains an open research problem.

From a software infrastructure perspective, we have added a prototype level set algorithm (to be used for enforcing dynamic excision boundaries for the first-order Einstein solver, amongst other things), whose reinitialization can be performed at user-specified intervals and is handled via the `wave_prop_advance` routine. This can be tested through the `euler_rgfm` struct, which uses the level set to implement Riemann ghost fluid method for the Euler equations, which we have validated against both 2D bubble collapse problems and the 1D Riemann problems of Fedkiw (https://physbam.stanford.edu/~fedkiw/papers/cam1999-31.pdf). We anticipate that this will need to be largely rewritten to accommodate @kltrbrds's much improved infrastructure for handling (potentially dynamic) embedded boundaries, but for now this serves as a proof-of-concept for the algorithm itself.

We also make several bug fixes to the neutron star code, significantly refactor the way that the curved spacetime source terms are handled (so that they can be folded in together with the moment-EM coupling terms), improve the stability of all of the relativistic solvers through a more principled handling of gauge constraints, and add a new `gr_spacetime` object corresponding to the initial data for a binary black hole system (i.e. the Brill-Lindquist metric). Some initial prototyping work exists for the handling of gravitational collapse with the ultra-relativistic Euler equations, essentially as a stress-test of the dynamical excision boundaries, but this has not been wrapped in Lua since we do not expect this to be a production-level test.

Unit tests all pass, no regression test issues have been noted, and it has been confirmed that the code builds correctly on both CPUs (MacOS and stellar-intel) and GPUs (stellar-amd).


ammarhakim (approved):
Ready to merge.

PR #769, opened 2025-06-24
Fix floating point comparisons in GK file import

manauref:
The bit of code that imports a distribution function from file checked whether the grids are the same with comparisons like
```
a == b
```
where `a` and `b` are doubles. This is of course not the proper way to compare floats/doubles. Change this to use `gkyl_compare_double` with a 1e-10 tolerance.


Antoinehoff:
I think it would be better to have a relative tolerance (we never know, maybe we will want to have a domain of a few nanometer one day...)
So I would propose to write instead
```C
gkyl_compare_double(grid_do.lower[d], grid.lower[d], tol_eq*grid_do.lower[d])
```

Antoinehoff (approved):
My bad 😃

PR #777, opened 2025-07-08
BUG FIX: /data files were not corrected after directories were reorganized

Maxwell-Rosen:
I found some bugs and I squashed them! 

Bug 1: There was an issue with the mirror geometry generator that I did $\epsilon = J g^{33} - J B / g_{33}$ when there should be no $B$. This term is not used anywhere, so we wouldn't have noticed. It would be used in a full 2x Poisson solve.

Bug 2: All the unit and regression tests were not pointing to the right /data/ files, so I fixed all that. I ran the unit tests, and they all work appropriately. I ran a few regression tests to make sure they were fine.

Bug 3: I had to move a data/gyrokinetic/wham_psi.gkyl file to core since the mirror geometry generators are in core for the mirror modules


manauref (changes requested):
fem_poisson_full is an empty file. Please remove

Maxwell-Rosen:
Removed

JonathanGorard (approved):

PR #791, opened 2025-07-23
Merge main into the Hamiltonian scaling branch

Maxwell-Rosen:
This PR merges main into some work we were doing before the dir-reorg merge. This merge was pretty easy since all the files were git mv'd into their new spots. I moved some of the new files into their respective places. The unit and regression tests pass and they all look normal.

More work that needs to be done before this branch is merged into main:

1) The location of the mirror throat is hardcoded right here https://github.com/ammarhakim/gkeyll/blob/gk_damp_scale_hamil-merge/gyrokinetic/apps/gk_species_fdot_multiplier.c#L165. To fix this, we can utilize code from the position map, perhaps moving it into a general method inside calc_bmag, to calculate the mirror throat automatically. This would be a straightforward task; however, we want this to generalize to 2x so that each field line has its own independent Bmax and location. I'm not sure what this structure should look like. Perhaps we find the extrema of the magnetic field at the nodes, then interpolate between.

2) Calculate the loss boundary for electrons in the expander region. There is a confined region of electrons here that traces a circle in phase space. Following the loss criterion arguments, you arrive at an equation

$$ \mu_{max} = \frac{m_s}{2 B(x)} \left(  \frac{2 e \phi} {m_s} - v_{||}^2 \right) $$

This equation is arrived at when you think about 

$$ m v^2 /2 < e \phi, $$

 then separate the velocity into parallel and perpendicular, then solve for $\mu$.

3) Put all this work into a complete time-stepper app to complement SSP-RK3. We need a DR for this. At the input file level, I think a user should just specify $\alpha$ and identify the time stepper ID. This new time stepper will modulate between the orbit-averaged phase and the full dynamics phase. I would expect an input like "frames_per_phase = f" where it prints f frames per OAP or FDP phase. This contradicts the current loop structure in the input files, so I'm not sure how to implement this. 
4) For beams and localized sources, we must compute the orbit-averaged source for the OAP. If we do not do this, the results will be terrible. We must average the source by integrating along orbit contours and taking the average flux of the source. We must compute

 $$\mathring{S}(\vec x,\vec v) = \frac{\oint S(\vec\ell_x,\vec\ell_v)d\vec{\ell}}{\oint d\vec{\ell}}. $$

The orbit contours are 

$$    d\ell_z = \frac{\partial \mathcal{H}(z,v_\parallel)}{\partial v_\parallel},\\
    d\ell_{v_\parallel} = - \frac{\partial \mathcal{H}(z,v_\parallel)}{\partial z}.
$$

Greg Hammett suggested an alternative. We could transform the source into $(E, \mu)$ coordinates and the orbit average will be much easier to perform.

5) We must identify the loss boundary with more accuracy. Scaling with a factor of 2 is not good enough. We can identify the "loss" cells through the corners of the cells, i.e., the nodal values, rather than the cell-centered coordinate. This may improve accuracy.


manauref:
I did this in creating the gk_oap_multirate-production branch. No need for this PR anymore.

Comment about point 3 above: I don't actually think a new time stepper is needed, we just need to make the loss cone mask available in the input file (maybe write it out and read it in), and have an input file that cycles between the FDP and the OAP, without having to destroy/create the app or re-project the loss cone mask every time.

Maxwell-Rosen:
Checking out the gk_oap_multirate-production branch, it uses the old directory structure. This is a new branch to me. It is hard to believe that main has been merged into that branch recently.

manauref:
uh, weird, it has the new directory structure locally, but I see that online it doesn't.

manauref:
fixed, I hadn't pushed properly

Issue #806, opened 2025-08-28
[DR] Gyrokinetic heating source


manauref:
At the moment the gyrokinetic solver allows for a plasma source term on the RHS of the equation $\mathcal{S}$
```math
\partial_t f + \dots = \mathcal{S}
```
which injects particles, momentum and energy. In some cases it may be desirable to inject energy alone without injecting particles (e.g. ohmic or RF heating). For this purpose we wish to introduce a heating source $\mathcal{S}_Q$
```math
\partial_t f + \dots = \mathcal{S} + \mathcal{S}_Q
```
of the BGK form
```math
\mathcal{S}_{Q} = \nu_{Q}  \left(f_{M}(n, u_{\parallel}, v_{tQ}^2) - f\right)
```

The heating rate $\nu_Q$ is likely problem dependent, and different simulations will require different values. The Maxwellian $f_M$ is constructed with the same particle number density $n$ and parallel drift $u_\parallel$ as the distribution $f$ so that no particles or momentum is injected. The thermal speed of this Maxwellian, $v_{tQ}$, will be adapted in time to ensure a desired amount of heating power $P_{in}$ is injected according to
```math
\int \mathrm{d}\mathbf{x}\int \mathrm{d}\mathbf{v} ~ \frac{1}{2}mv^2 ~ \mathcal{S}_{Q} = P_{in}
```
which, if we assume a constant unit shape $s_Q(\mathbf{x})$ to the heating thermal speed $v_{tQ}=(T_Q(t)/m) ~ s_Q(x)$, allows us to compute the amplitude as
```math
\frac{T_Q(t)}{m} = \left(2P_{in} / (d_v m) + \int\mathrm{d}\mathbf{x} ~ \nu_Q n v_t^2\right) \left( \int\mathrm{d}\mathbf{x} ~ \nu_Q n s_Q(x)\right)^{-1}
```
Here $d_v$ is 1 for 1v and 3 for 2v.

Input file interface
------------------

In a GK input file we will request this type of heating source with a `heating` table in the species table, for example:
```
  .elc = {
    ...
    .heating = {
      .heating_id = GKYL_HEATING_DEFAULT, // Type of heating model.
      .rate_profile = heat_rate_func_elc, // Heating rate nu_Q(x).
      .rate_profile_ctx = &ctx,
      .temp_shape = heat_temp_shape_func_elc, // Maxwellian temperature unit shape s_Q(x)
      .temp_shape_ctx = &ctx,
[...]

PR #808, opened 2025-09-02
GK heat source 

manauref:
This PR adds a heat-only source term as described in DR #806 , allowing us to heat a species (i.e. increase its thermal energy) without adding additional particles or momentum.

The code changes pretty much follow the description in the DR. Only two changes/comments not addressed there come to mind:

1. We added a `write_diagnostics` optional input, which writes out the initial heating rate $\nu_Q(x)$ and squared thermal speed $s_Q(x)$ profiles, as well as the squared thermal speed amplitude $T_{Q}(t)/m$.
2. At the moment this term is explicit only. We have not implemented an implicit version because its current use does not limit $\Delta t$, usually the LBO does. But with implicit BGK collisions, if the heating rate is set to $\nu_{ee}$, and if the plasma is collisional enough, then one may actually want this term to be implicit. Such modification can easily be made in the future.

Example use
-------------

In many previous simulations we sourced and heated with the same source $\mathcal{S}$. But because parallel electron heat  transport can be so fast, we found ourselves having to use a relatively high (e.g. 1.5X, 2X or higher) source temperature. This was the case in LAPD simulations where, for example, for a target $T_e$ of 6 eV a source temperature $T_{src,e}=13.8$ eV was needed. In this LAPD case, one could instead use $\mathcal{S}$ with a source temperature of $T_{src,e}=6$ eV, and introduce an additional heating source $S_Q$ for the electrons, injecting 18 kW of power in the central region, which raises the electron temperature a little above 6 eV, without needing to use a very hot $\mathcal{S}$:

Screenshot 2025-09-01 at 10 39 03 PM

Note that for the sim with additional heating, we used a homogeneous shape $s_Q=1$ and a heating rate that peaks at $nu_{ee}$ near the center and drops outside the cathode region. Also, this sim used no extra heating for the ions and we lowered $T_{src,i}$ to 3 eV from 5 eV (the latter for the original sim w/o $\mathcal{S}_Q$.

Here are a couple of snapshots at $t=6$ ms:

Screenshot 2025-09-01 at 10 49 40 PM

Interestingly, notice that a rise in the temperature near the outer boundary that sometimes occurs in simulations no longer appears when using $\mathcal{S}_Q$.

Performance
-------------

In the above 3x2v LAPD simulation the heating source took 9% of the Forward Euler time on 4 A100 GPUs. Both simulations (w/ and w/o extra heating) took virtually the same amount of time $\pm$ 3% of runtime.

Notes for use elsewhere
-------------------------

Keep in mind that the choices of $\nu_Q(x)$, $s_Q(x)$ and heating power vary from one simulated experiment to another, and that these free parameters/profiles need to be adjusted to resemble the experimental heating mechanism one intends to mimic numerically.


Antoinehoff (approved):

Antoinehoff:
I approved it but maybe it should be valgrind and compute sanitizer tested before merging if it hasn't been done yet.

manauref:
> I approved it but maybe it should be valgrind and compute sanitizer tested before merging if it hasn't been done yet.

I checked, it is compute sanitizer and valgrind clean.

PR #809, opened 2025-09-02
Pitagora config files

jaimix4:
Add mkdeps and configure files for successfully installing Gkeyll in Pitagora, the Eurofusion supercomputer.

### DIFF
--- a/machines/configure.pitagora-gpu.sh
+++ b/machines/configure.pitagora-gpu.sh
@@ -1,4 +1,4 @@
-# Trying the only available version of cuda, openmpi, nccl in pitagora
+# Loading modules, pitagora currently only supports cuda 12.6
 
 module load cuda/12.6
 module load openmpi/4.1.6--gcc--12.3.0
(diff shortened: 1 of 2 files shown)


manauref (approved):

JunoRavin (approved):
This looks good to me. By the way, I noticed the version of NCCL you're using is 2.22. What simulations have you tried running on this cluster? I recently ran into a strange issue that seemed to be NCCL version dependent (didn't show up in 2.18, did show up in 2.19) where periodic boundary conditions would fail to be applied and I'm curious if you have run any simulations yet with periodic boundary conditions on multiple GPUs.

JunoRavin:
@jaimix4 yea can you just try running vlasov/luareg/rt_vlasov_twostream_p2.lua 

on multiple GPUs? 

Also, do you have any sense of what changed when you did the fresh installation with Gkeyll that is now working with NCCL 2.22?

jaimix4:
@JunoRavin I ran rt_vlasov_twostream_p2 (C input file version) on multiple GPUs and it worked with NCCL 2.22. 
Not really, since I was always trying to install solely with NCCL 2.22.

PR #810, opened 2025-09-09
Switch skin_ghost_ranges from global to local

jtenbarg:
This PR switches skin_ghost_ranges from global to local, which fixes a segfault issue for running wedge BCs in parallel


JunoRavin:
This PR also brings in a hot fix from the grad-closure-prototype work by @kltrbrds https://github.com/ammarhakim/gkeyll/blob/grad-closure-prototype/apps/moment.c#L127C3-L127C69

that this change was needed for because otherwise we were allocating large amounts of memory in large parallel runs (since every MPI process was creating bc_buffers based on the *global* extended range).

JunoRavin (approved):

JonathanGorard (approved):

PR #811, opened 2025-09-10
Miscellaneous fixes and updates (PKPM NN, Vlasov Lua wrappers, GK output, and regression tests)

JonathanGorard:
This branch is a bit of a ragbag of minor updates and fixes:

1. Fixing the `gyrokinetic_app_cout` function so that we correctly flush the buffer on rank 0 (was previously screwing up the step messages when running GK simulations via the Lua wrappers).
2. Various updates and fixes to the Vlasov Lua wrappers (e.g. adding new wrappers for diffusion tensors and arbitrary order diffusion operators, and standardizing some function names and equation object names to make them more consistent between apps).
3. Various optimizations to the NN training implementation within the PKPM app (mostly just replacing heap allocations with stack allocations, preallocating heap memory outside of training loops where possible, cleaning up some function signatures, etc.). Also fixed a typo in the PKPM output that had been bugging me for a while.
4. Modifying the Moments and Lua regression tests (GEM and two-stream instability) to match what we have written in the documentation (i.e. making the input file modifications suggested by @JunoRavin).


JonathanGorard:
> This all looks good to me, though I would love to think about the interface for the neural network training at the Lua and C level where we have some sort of general helper functions for the allocations and initial setup, because as of right now every C file and the Lua wrappers has a pretty dense block of code for the neutral network setup, and I think this is going to confuse folks that might need to make a PKPM C input file for whatever reason (same reason we try to make the setup of the communicators a little more streamlined for quickly testing things through the C input file and improving maintainability of the Lua wrapper).

Yes, I agree - if we feel that this API design on the input file side is now more-or-less stable (which I think arguably it is), then we should eventually bake in the NN loading and initialization logic into a set of helper functions that can be called directly within the C input files.

ammarhakim (approved):
Ready to merge so we can bring this into production and test.

PR #814, opened 2025-09-11
Implementation of general relativistic hydrodynamics (GRMHD) equations

JonathanGorard:
This PR adds new solvers for the equations of general relativistic magnetohydrodynamics (GRMHD) to the Gkeyll Moments app, using both the conservation law form of [Antón et al. 2005](https://arxiv.org/abs/astro-ph/0506063), as well as the SRMHD form in our own locally-flat tetrad basis, allowing for direct comparison between the two (with unit tests confirming that the results are equivalent). These are represented by the equation objects `wv_gr_mhd` and `wv_gr_mhd_tetrad`, respectively. We add both Lax-Friedrichs and HLLE-type solvers in each case, with HLLE being the default, and we use the same modification to the geometric source terms to eliminate spacetime metric terms using the constraint equations (allowing support for cases where the spacetime metric is not known a priori, i.e. when solving Einstein's equations numerically) that we have already employed and validated in the relativistic hydrodynamics case. These solvers therefore work equally well in both stationary and dynamic spacetimes.

For the reconstruction of the primitive variables, we employ the effective 2D method of [Noble et al. 2006](https://iopscience.iop.org/article/10.1086/500349), with a fallback to the effective 1D method of [Newman and Hamlin 2014](https://epubs.siam.org/doi/10.1137/140956749) (in whose paper we discovered a crucial typo) in the event that the former does not converge. As demonstrated via our unit tests, this allows for highly robust reconstruction within arbitrary spacetimes, even in regions of high gamma, high magnetization, and low pressure/density: the only floors/ceilings that we need to impose are the existing ceilings on the gamma factor and density/pressure that we must impose anyway as part of the hydrodynamic part of the solver.

For dealing with divergence errors, we initially implemented the hyperbolic divergence cleaning scheme of [Dedner et al. 2002](https://www.sciencedirect.com/science/article/abs/pii/S002199910196961X), although we discovered that these results were unsatisfactory, and so modified the scheme to be more similar to a variant of the hyperbolic divergence cleaning approach used by GIZMO, as outlined by [Fedrigo and Lupi 2025](https://arxiv.org/abs/2506.15775v1). In the locally-flat tetrad basis, this divergence cleaning scheme looks very similar to what we already implement for Maxwell's equations. We are now able to run highly magnetized black hole accretion problems (for high black hole spin) for more than 50M, with the divergence errors maintained below 10^-12.

The scheme has been validated against the SRMHD Riemann problems of [Balsara 2001](https://iopscience.iop.org/article/10.1086/318941) (including, in particular, the relativistic generalization of the Brio-Wu shock tube, which we include as a regression test), as well as against magnetized versions of the BHL accretion problems around Schwarzschild and Kerr black holes that were already used as validation test cases for general relativistic hydrodynamics. We also include a verification test to confirm that the schemes work in the dynamic spacetime case (at least in spherical and axial symmetry), using a magnetized version of the gravitational collapse problem to a Kerr black hole that was added for relativistic hydrodynamics.

Our aim is to use these solvers primarily as a baseline for validating and extending our relativistic multi-fluid work, as well as for the collapsar studies being led by @nikolabukowiecka.


JunoRavin:
> Just noting here something which I mentioned to @ammarhakim but didn't record publicly: we're switching to using a consistent representation of momentum as a _covector_ (with 3-velocity and 4-velocity still being vectors) across all general relativistic equation objects. This is something that was previously inconsistent, and which I would like to standardize in light of the addition of GRMHD equations (whose formulation is much cleaner with a covector representation of momentum). This necessitates minor modifications to e.g. the transformation rules used within the GR Euler equations.

Is this not yet standardized? It seems like you have rearranged the component manipulations of (all of?) the general relativistic equation objects. Or is there more work that needs to be done on this front?

JunoRavin (approved):

nikolabukowiecka (approved):
Looks good. 
One comment: Something that I discovered while using the new solvers for GRMHD equations is that as with other GRMHD solving codes, if one wants to obtain magnetic filed strength in Gauss, one should include the sqrt(4*pi) factor, which is not obvious considering the choice of units in Gkeyll.

PR #818, opened 2025-09-12
Fix a compatibility issue between the adaptive source and LTE moment correction

Antoinehoff:
This pull request makes a targeted change to the way density is initialized in the `init_maxwellian_gaussian` function. The update ensures that the density calculation now includes the `f_floor` value, which helps prevent issues when considering an adaptive source that aims to only compensate the loss of particle (in this case the `inp.total_num_particles` is 0 at `t=0`).
This resolves an incompatibility issue when using BGK + adapt source, because the `.correct.correct_all_moms = true` flag was introducing a division by 0 in the adaptive source initialization.

We also implement a fix to the nan observed in the rt_gk_tcv_iwl_adapt_src_2x2v. The definition of the Gaussian shape used for configuration space dependency in the MAXWELLIAN_GAUSSIAN projection was not appropriate for 2x2v as there was a loop over GKYL_MAX_CDIM which is always 3, even in 2x2v. We now pass app->cdim to this function to ensure that this loop is well defined.


JunoRavin:
@Antoinehoff do you want to keep the same changes to the regression test in this PR? Otherwise I think this is good to go.

Antoinehoff:
@JunoRavin @manauref this is ready to be merged I think

manauref (approved):
i love bug fixes

Antoinehoff:
> i love bug fixes

we can always introduce more to resolve them later 😌

PR #819, opened 2025-09-16
Add helper functions for the input file to use for the POA scheme

Maxwell-Rosen:
I tried editing the app in the regression test directly, but it needs gyrokinetic_priv.h to be able to edit the app. This is not good programming, so a few functions are added to gyrokinetic.c to toggle different features.

gkyl_gyrokinetic_app_reset_fdot_mult: Toggles the fdot_mult term. It must be initilized to the type during the OAP phase. In the mirror tests, this is type 2.
gkyl_gyrokinetic_app_reset_enforce_positivity: Toggles the positivity hack. Must be initilized on so that the appropriate arrays are created.
gkyl_gyrokinetic_app_reset_update_field: Toggles between a dynamic and static field. Must be initilized dynamic.
gkyl_gyrokinetic_app_reset_update_species: Toggles between static/dynamic species. Must be initilized dynamic.

A function gk_species_reset_static is added to gk_species because the function pointers need to be changed for this app, depending on whether we are static or dynamic.

The regression test gyrokinetic/creg/rt_gk_wham_1x2v_p1_POA.c is added to demonstrate how the POA scheme can be performed in an input file


Maxwell-Rosen:
I have not tested or thought about restarts. I need to fix that too

ammarhakim:
@Maxwell-Rosen  Please change the title to tell us what "POA" means. Also, the PR description seems very obscure. Please do elaborate a little so I know what is happening here.

PR #821, opened 2025-09-17
Correct the fact that some nodal arrays were not copied over to the G…

manauref:
…PU version of gk_geom; this was throwing a seg fault at the end of a simulation (e.g. rt_gk_lapd_cart_3x2v_p1), as discovered by @antoinehoff. Puzzlingly, compute-sanitizer didn't find an issue with it, but cuda-gdb did. Also correct the fact that gk_species_projection was calling the maxwellian correction routines for the PROJ_FUNC option, which doesn't make sense and was also throwing a seg fault.


Antoinehoff:
Thank you @manauref. I will double check this branch on perlmutter right now. The changes are kind of hard to read because of all these coding style changes, is this necessary? It does not feel to comply with Gkeyll's standards...

manauref:
> see message above

JunoRavin (approved):

PR #822, opened 2025-09-17
Added warning for headerless eqdsk files

SjoerdPR:
In https://github.com/ammarhakim/gkeyll/commit/c2d8c56907afeb7c6922aff7790577d771cef348 a change was made to the eqdsk parser so that it now reads the header, removing the need to manually remove the header from the eqdsk. However, if an eqdsk file used in previous gkeyll simulations (with the header removed) is used, this causes errors: The parser assumes the first 48 characters of the file are the header, and then attempts to read the `nr` and `nz` values from the `rdim`, `zdim` etc. block. This will generally cause cryptic error messages that can be difficult to debug for the end user unaware of these changes to the parser.

This change checks if the first 48 characters contain a newline character, which is a sign the header has been removed, and gives the user a warning if this is the case.


manauref:
approving after a long time of this sitting around. I just spent an hour tracking down a seg fault, only to find out it was because of a missing header in the eqdsk file.

manauref (approved):

PR #833, opened 2025-09-19
GPU bug fixes for newer versions of NCCL (>2.18) and fixes to Makefile 

JunoRavin:
There were two residual issues when running GPU and multi-GPU simulations on my workstation which I believe we have now fixed, but because these are low level changes to core, I am tagging a bunch of folks in to make sure this branch does not affect any production runs. 

The first issue is that on newer versions of NCCL (>2.18), the fact that we set config.blocking=0 (so by default non-blocking communication) now *also* sets collective calls to be non-blocking. Meaning the status returned by NCCL's allreduce, broadcast, etc. is not automatically ncclSuccess, it is ncclInProgress. I have confirmed that on nccl 2.26 on my workstation (despite Nvidia's documentation suggesting otherwise) that the ncclStatus is ncclInProgress, so I have added a new wait_for_collective call which is modeled after the same checkNCCL(ncclCommGetAsyncError(nccl->ncomm, &async_stat)); we have to do when we initialize the communicators, barrier, wait, and such, though at Codex's recommendation this is slightly more aggressive than our other barriers and waits. We don't have any other instances where we surface any asynchronous errors that occurred during execution. I am not sure if this is something that *can* happen, but I do know that this wait call fixes the issue on my workstation of an invalid communicator (because the unit tests for example try to do a NCCL call when the status is ncclInProgress) and given the fact that I have seen NCCL instability on other clusters with NCCL versions > 2.18, I think we need to be watchful for these types of default behavior changes (I will admit **I would not** have figured this error out without OpenAI's Codex, because Nvidia's documentation **explicitly** states NCCL's allreduce returns ncclSuccess, but I checked this running our unit tests.. it's not true, it returns ncclInProgress on a NCCL 2.26 installation). 

The second issue is that with the new hierarchal app executable, we were accumulating the CFLAGS to determine which app is enabled (since if you're building the whole system you tell Gkeyll it has the Moment app, Vlasov app, Gyrokinetic app, and PKPM app), but this accumulation was happening **before** the nvcc check and CFLAGS designation (since if nvcc is enabled, we assign CFLAGS again to give us some flexibility to assign different flags on GPUs instead of CPUs instead of just trying to accumulate the GPU specific flags). So I moved the CFLAGS accumulation until after the nvcc check and violà apps are enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,8 +369,16 @@ on CPUs with e.g.
 valgrind --leak-check=full <executable> <command_line_arguments>
 ```
 
-and compute-sanitizer clean on GPUs with e.g.
+and compute-sanitizer clean on one GPU with e.g.
 
 ```
-compute-sanitizer --tool memcheck <executable> <command_line_arguments>
+compute-sanitizer --tool memcheck --leak-check full <executable> <command_line_arguments>
 ```
+
+To check using multiple GPUs on a Perlmuttter node with SLURM e.g.
+
+```
+srun -N 1 --ntasks=2 --gpus-per-task=1 --gpu-bind=closest compute-sanitizer --tool memcheck --leak-check full <executable> <command_line_arguments>
+```
+
+where the command line arguments must at least contain `-g -M -direction 2` with `direction` being the direction along which the domain is decomposed, e.g. `-e 2` for 3x2v gk simulations.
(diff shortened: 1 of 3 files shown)


manauref (approved):

manauref:
very nice investigative work @JunoRavin

PR #835, opened 2025-09-19
Software improvements to TS BC

manauref:
doesn't change the core algorithm, just software changes to minimize code duplication and improve design.

Unit test continues to pass


Antoinehoff (approved):
thanks for cleaning this

PR #838, opened 2025-09-24
The q-profile cubic fit coefficients got reversed in the TCV regression tests.

Antoinehoff:
This happened when we put back the cubic fit instead of the piecewise linear ones during the tsbc_fix branch merge.


Antoinehoff:
> what about the d3d test? is it consistent?

no it has been adapted. I am almost sure that @tnbernard is using `qprofile(R)` as it was under her advice that I used it for TCV simulations.

Issue #839, opened 2025-09-26
GK GEOMETRY_FROMFILE option not working for 1D


manauref:
A 1D tokamak simulation runs fine if I generate the geometry. But if I immediately run it again with GEOMETRY_FROMFILE, the time step blows up after one time step.

Probably an easy fix. We should double check 2D and 3D FROMFILE to be safe.

manauref:
Another thing:
when FROMFILE is used incorrectly the simulation seg faults without explanation or crashes. We should add a print statement and do some basic checks (e.g. that the data read has the same DOF as the simulation expects)

PR #840, opened 2025-09-29
1x Fromfile Fix

akashukla:
Addresses the bug in issue #839 . Thanks to @manauref for catching this.

modal to nodal and nodal to modal operations on surfaces are different for 1x simulations since there is no 0d basis. This commit addresses the issue and makes geometry fromfile work correctly for 1x. Below is the final potential from the gk_sheath_1x2v regression test run normally with mapc2p geometry and then with geometry fromfile. 
image

### DIFF
--- a/core/zero/nodal_ops.c
+++ b/core/zero/nodal_ops.c
@@ -208,11 +208,61 @@ gkyl_nodal_ops_n2m_surface(const struct gkyl_nodal_ops *nodal_ops,
 }
 
 void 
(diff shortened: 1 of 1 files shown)


manauref:
> > thanks @akashukla is a corresponding change in n2m needed?
> 
> yes. Did this now. The n2m isn't actually used anywhere, but it's good to have the inverse function.

indeed, thank you

JunoRavin (approved):

PR #841, opened 2025-09-29
Remove the lower x bc enforcement for the field 

Antoinehoff:
We discovered that the current state of the main is not able to run a typical coarse production run of tcv turbulence simulation. The latest stable commit was identified (https://github.com/ammarhakim/gkeyll/commit/7e18b246f38f25bc7641bbbb10e21c8079e54409) which is just before putting the x_ssfg_lo updater and enforcing the BC of phi after applying TSBC. We indeed observed that the TSBC was breaking the 0 Dirichlet BC of phi (see DR https://github.com/ammarhakim/gkeyll/issues/698) but many production runs were done since then without suffering from any unphysical nor numerical issuses.
This is why we propose here to remove the lower x ssfg application and let the lower radial skin values of phi to be altered by the TSBC application.

#### More details:
The following files show how the same input file yield different dynamics. The current version of the main is showing a drop of time step size at ~16mus whereas an older commit shows a constant dt up to t>300mus.

[tcv_pt_coarse_main.log](https://github.com/user-attachments/files/22606962/tcv_pt_coarse_main.log)
[tcv_pt_coarse_commit_7e18b2.log](https://github.com/user-attachments/files/22606963/tcv_pt_coarse_commit_7e18b2.log)

The two C input files are the same except the change of BC table.
[gkeyll_main.c](https://github.com/user-attachments/files/22606971/gkeyll_main.c)
[gkeyll_commit_7e18b2.c](https://github.com/user-attachments/files/22606970/gkeyll_commit_7e18b2.c)

Removing the lower x ssfg updater solves this issue.

### DIFF
--- a/gyrokinetic/apps/gk_field.c
+++ b/gyrokinetic/apps/gk_field.c
@@ -127,23 +127,6 @@ gk_field_enforce_zbc_none(const gkyl_gyrokinetic_app *app, struct gk_field *fiel
 {
 }
 
-// Ensure that the value at the lower x boundary matches the Dirichlet boundary condition value.
-static void 
-gk_field_enforce_xbc(const gkyl_gyrokinetic_app *app, struct gk_field *field, struct gkyl_array *finout)
-{
-  gkyl_array_clear_range(finout, 0.0, &app->lower_skin[0]);
-  double dg_norm = pow(sqrt(2.0),app->basis.ndim);
-  struct gkyl_gyrokinetic_bc *bc_lo_x = gk_fetch_bc_with_dir_edge(field->info.poisson_bcs, 2*(app->cdim-1), 0, GKYL_LOWER_EDGE);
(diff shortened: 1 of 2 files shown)


manauref:
> such a sad PR.... Screenshot 2025-09-29 at 6 08 46 PM

removing lines is better than adding them IMO

JunoRavin (approved):
Is this something we want to optionally have that does work at high resolution? Or is your conclusion just that fixing the boundary condition in x is more trouble than it is worth?

Antoinehoff:
> Is this something we want to optionally have that does work at high resolution? Or is your conclusion just that fixing the boundary condition in x is more trouble than it is worth?

yes 🙃

Antoinehoff:
joke aside, let me merge it once I'm sure that this is the only issue when running this coarse production run

PR #843, opened 2025-09-30
Refactor gk_field.c

Maxwell-Rosen:
This PR is highly painful to merge other changes into, so I'd like this to be a priority.

gk_field.c is quite messy and has needed refactoring for a while. This refactor makes individual functions for the init and solves initialization protocols. It essentially just moves some things around.

----
Regression tests, I checked that are looking normal
rt_gk_d3d_iwl_2x2v_p1
rt_gk_wham_1x2v_p1
rt_gk_wham_2x2v_p1
rt_gk_wham_3x2v_p1
rt_gk_bgk_asdex_2x2v_p1
rt_gk_lapd_cyl_3x2v_p1 
rt_gk_sheath_1x2v_p1
rt_gk_sheath_2x2v_p1
rt_gk_sheath_3x2v_p1

### DIFF
--- a/gyrokinetic/apps/gk_field.c
+++ b/gyrokinetic/apps/gk_field.c
@@ -11,6 +11,133 @@
 #include <float.h>
 #include <time.h>
 
+// Function pointer types for FEM object initialization
+typedef void (*gk_field_fem_init_func_t)(struct gkyl_gyrokinetic_app *app, struct gk_field *f, 
+  double polarization_weight);
+
+// FEM initialization functions for different field types and dimensions
+static void
+gk_field_fem_init_boltzmann(struct gkyl_gyrokinetic_app *app, struct gk_field *f, 
+  double polarization_weight)
+{
+  f->ambi_pot = gkyl_ambi_bolt_potential_new(&app->grid, &app->basis, 
+    f->info.electron_mass, f->info.electron_charge, f->info.electron_temp, app->use_gpu);
+  
+  // Sheath_vals contains both the density and potential sheath values.
+  for (int j=0; j<app->cdim; ++j) {
+    f->sheath_vals[2*j]   = mkarr(app->use_gpu, 2*app->basis.num_basis, app->local_ext.volume);
+    f->sheath_vals[2*j+1] = mkarr(app->use_gpu, 2*app->basis.num_basis, app->local_ext.volume);
+  }
+}
+
+static void
+gk_field_fem_init_1d(struct gkyl_gyrokinetic_app *app, struct gk_field *f,
+  double polarization_weight)
+{
+  // Allocate array for the polarization weight times geometric coefficients.
+  f->epsilon = mkarr(app->use_gpu, (2*(app->cdim/3)+1)*app->basis.num_basis, app->local_ext.volume);
+
+  // Need to set weight to kperpsq*polarizationWeight for use in potential smoothing.
+  gkyl_array_copy(f->epsilon, app->gk_geom->geo_int.jacobgeo);
+  gkyl_array_scale(f->epsilon, polarization_weight);
+  gkyl_array_scale(f->epsilon, f->info.kperpSq);
+
+  if (f->gkfield_id == GKYL_GK_FIELD_ADIABATIC) {
+    // Add the contribution from adiabatic electrons
+    double n_s0 = f->info.electron_density;
+    double q_s = f->info.electron_charge;
+    double T_s = f->info.electron_temp;
+    double quasineut_contr = q_s*n_s0*q_s/T_s;
+    
+    struct gkyl_array *epsilon_adiab = mkarr(app->use_gpu, f->epsilon->ncomp, f->epsilon->size);
+    gkyl_array_copy(epsilon_adiab, app->gk_geom->geo_int.jacobgeo);
+    gkyl_array_scale(epsilon_adiab, quasineut_contr);
+    gkyl_array_accumulate(f->epsilon, 1., epsilon_adiab);
+    gkyl_array_release(epsilon_adiab);
+  }
+  // Gather epsilon for (global) smoothing in z.
+  f->epsilon_global = mkarr(app->use_gpu, f->epsilon->ncomp, app->global_ext.volume);
+  gkyl_comm_array_allgather(app->comm, &app->local, &app->global, f->epsilon, f->epsilon_global);
+}
+
+static void
+gk_field_fem_init_2d3d(struct gkyl_gyrokinetic_app *app, struct gk_field *f, 
+  double polarization_weight)
+{
+  // Initialize the polarization weight.
+  struct gkyl_array *Jgij[3] = {app->gk_geom->geo_int.gxxj, app->gk_geom->geo_int.gxyj, app->gk_geom->geo_int.gyyj};
+  for (int i=0; i<app->cdim-2/app->cdim; i++) {
+    gkyl_array_set_offset(f->epsilon, polarization_weight, Jgij[i], i*app->basis.num_basis);
+  }
+
+  // Translate input file BCs into Poisson BCs.
+  struct gkyl_poisson_bc poisson_bcs = { };
+  for (int d=0; d<app->cdim-1; d++) {
+    struct gkyl_gyrokinetic_bc *bc_lo = gk_fetch_bc_with_dir_edge(f->info.poisson_bcs, 2*app->cdim, d, GKYL_LOWER_EDGE);
+    if (bc_lo != 0) {
+      poisson_bcs.lo_type[d] = gkyl_gyrokinetic_translate_poisson_bc_type(bc_lo->type);
+      for (int i=0; i<3; i++)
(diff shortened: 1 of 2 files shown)


gyrokinetic/apps/gk_field.c
```diff
@@ -191,128 +495,48 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
...
+    es_energy_fac_1d_adiabatic = 0.5*quasineut_contr;
+  }
+
+  if (f->gkfield_id == GKYL_GK_FIELD_BOLTZMANN) { 
+    gk_field_fem_init_boltzmann(app, f, polarization_weight);
+    f->field_solve = gk_field_boltzmann_solve;
+  }
+  else {
+    if (app->cdim == 1) {
+      gk_field_fem_init_1d(app, f, polarization_weight);
```
Antoinehoff, line 531: I think it would be better to call this function `gk_field_fem_init_1x` to be consistent with `gk_field_poisson_solve_1x`. Same for the following `2d3d` that should be `2x3x`

Antoinehoff:
Can you use this PR to remove this line please? I think I forgot to remove it when removing the lower x ssfg in PR #841 
https://github.com/ammarhakim/gkeyll/blob/f776521ff91c9800e30fc4d601f1ab1034d958b5/gyrokinetic/apps/gk_field.c#L722

Maxwell-Rosen:
Thank you for the suggestions.
rt_gk_d3d_iwl_3x2v_p1 works and is valgrind clean. Doing a production check would be incredibly helpful.
Avoiding the if statement at 418 would be possible with a pointer to a null function or a working function. There are so many if statements here that the one at 418 is small in comparison.

Antoinehoff:
I think that we should extend this PR to the part of `gyrokinetic.c` that are dealing with the field. For example this routine:
https://github.com/ammarhakim/gkeyll/blob/41ddbaaf1ff25f762519f956a80731d2d45711f1/gyrokinetic/apps/gyrokinetic.c#L503
I don't understand why this should not replaced be a call like :
```C
app->field->update(app, tcurr, fin);
```
because all the functions that are called in `gyrokinetic_calc_field_update` are defined in `gk_field.c`. I don't see the point to have `gk_field_accumulate_rho_c` callable outside of `gk_field.c` since we always call it right before solving the field.
Also this will give a good structure to add EM solvers because the structure will be very similar.

gyrokinetic/apps/gk_field.c
```diff
@@ -49,6 +50,52 @@ gk_field_calc_energy_dt(gkyl_gyrokinetic_app *app, const struct gk_field *field,
   field->calc_energy_dt_func(app, field, dt, energy_reduced);
 }
 
+void
+gk_field_calc_phi_wall(gkyl_gyrokinetic_app *app, struct gk_field *field, double tm)
```
manauref, line 54: shouldn't this be in one of the sub gk_field files? as it only applies to some cases

manauref (changes requested):
Nice split Max. My initial hunch is that there's likely a better way to split the GK field, but I can't think of it now, so we can go with this.

Since this is a very dangerous change (affecting all sims) can you make sure the following tests are valgrind and compute-sanitizer clean:
```
rt_gk_bgk_im_asdex_2x2v_p1
rt_gk_d3d_iwl_2x2v_p1
rt_gk_d3d_iwl_3x2v_p1
rt_gk_lapd_cyl_3x2v_p1
rt_gk_lbo_relax_nonuniformv_1x2v_p1
rt_gk_ltx_1x2v_p1
rt_gk_ltx_boltz_elc_1x2v_p1
rt_gk_mirror_boltz_elc_1x2v_p1
rt_gk_mirror_kinetic_elc_1x2v_p1
rt_gk_neut_recycle_1x3v_p1
rt_gk_rad_nonuniformv_1x2v_p1
rt_gk_sheath_flr_2x2v_p1
rt_gk_step_out_2x2v_p1
rt_gk_tcv_iwl_adapt_source_3x2v_p1
```
please?

Moreover, given the delicate rearrangement, I think you should check for correctness too. So please post plots comparing the final `-field` and the `-field_energy` time trace to those obtained in main (side by side for 2D plots and on top of each other for 1D plots)

manauref:
> I have checked, and these tests are all valgrind clean. Compute sanitizer and verification plots are a bit trickier locally. Either I have to deconstruct my PC and labotomize the GPU temporarily, or spend an hour on plotting scripts.
> 
> ```
> rt_gk_bgk_im_asdex_2x2v_p1
> rt_gk_d3d_iwl_2x2v_p1
> rt_gk_d3d_iwl_3x2v_p1
> rt_gk_lapd_cyl_3x2v_p1
> rt_gk_lbo_relax_nonuniformv_1x2v_p1
> rt_gk_ltx_1x2v_p1
> rt_gk_ltx_boltz_elc_1x2v_p1
> rt_gk_mirror_boltz_elc_1x2v_p1
> rt_gk_mirror_kinetic_elc_1x2v_p1
> rt_gk_neut_recycle_1x3v_p1
> rt_gk_rad_nonuniformv_1x2v_p1
> rt_gk_sheath_flr_2x2v_p1
> rt_gk_step_out_2x2v_p1
> rt_gk_tcv_iwl_adapt_source_3x2v_p1
> ```
> 
> The only valgrind error I get is this, which isn't really a valgrind error, but a warning flag. It has nothing to do with this refactor.
> 
> ```
> ==496706== Memcheck, a memory error detector
> ==496706== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
> ==496706== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
> ==496706== Command: ./build/gyrokinetic/creg/rt_gk_sheath_flr_2x2v_p1 -s1
> ==496706== 
> ==496706== Conditional jump or move depends on uninitialised value(s)
> ==496706==    at 0x8E8FC82: gkyl_deflated_fem_poisson_new (zero/deflated_fem_poisson.c:142)
> ==496706==    by 0x8F0FD44: gk_species_init (apps/gk_species.c:1842)
> ==496706==    by 0x8F26CC4: gkyl_gyrokinetic_app_new_solver (apps/gyrokinetic.c:662)
> ==496706==    by 0x8F27789: gkyl_gyrokinetic_app_new (apps/gyrokinetic.c:759)
> ==496706==    by 0x10BCFA: main (creg/rt_gk_sheath_flr_2x2v_p1.c:645)
> ==496706== 
> ==496706== Conditional jump or move depends on uninitialised value(s)
> ==496706==    at 0x8E90048: gkyl_deflated_fem_poisson_new (zero/deflated_fem_poisson.c:142)
> ==496706==    by 0x8F0FD44: gk_species_init (apps/gk_species.c:1842)
> ==496706==    by 0x8F26CC4: gkyl_gyrokinetic_app_new_solver (apps/gyrokinetic.c:662)
> ==496706==    by 0x8F27789: gkyl_gyrokinetic_app_new (apps/gyrokinetic.c:759)
> ==496706==    by 0x10BCFA: main (creg/rt_gk_sheath_flr_2x2v_p1.c:645)
> ==496706== 
> ```

Yes, plotting the comparison takes time, but it's necessary.

Regarding the valgrind error: can you look into it? because I've run all the above-mentioned cases in main and didn't see this error/warning, so it could be an indication that something is not initialized after this reorg.

Maxwell-Rosen:
Compute sanitizer reveals that this branch is just as clean as main.

Here are the comparison plots with main

gk_tcv_iwl_adapt_source_3x2v_p1-field_1
gk_step_out_2x2v_p1-field_1
gk_sheath_flr_2x2v_p1-field_1
gk_rad_nonuniformv_1x2v_p1-field_0
gk_mirror_kinetic_elc_1x2v_p1-field_1
gk_mirror_boltz_elc_1x2v_p1-field_1
gk_ltx_boltz_elc_1x2v_p1-field_1
gk_d3d_iwl_3x2v_p1-field_1
gk_ltx_1x2v_p1-field_1
gk_lbo_relax_nonuniformv_1x2v_p1-field_0
gk_lapd_cyl_3x2v_p1-field_1
gk_d3d_iwl_2x2v_p1-field_1
gk_bgk_im_asdex_2x2v_p1-field_1

Maxwell-Rosen:
Here is the ion sound test comparison. Compute sanitizer doesn't give any warnings about the field either.
gk_ion_sound_adiabatic_elc_1x2v_p1-field_1

gyrokinetic/apps/gk_field.c
```diff
@@ -457,7 +336,7 @@ gk_field_new(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app)
     f->invert_flr = gk_field_invert_flr;
 
     double flr_weight = 0.0; 
-    for (int i=0; i<app->num_species; ++i) {
+    for (int i = 0; i < app->num_species; ++i) {
```
manauref: FLR only makes sense in 2x/3x, so it should move to that file

manauref (approved):

PR #844, opened 2025-10-01
Micro fix of a CPU warning: hardcoded variable for an array size

Antoinehoff:
This is a micro fix of a warning occuring because the CPU compiler does not like to use a variable (even if the value is hardcoded) to initialize the size of an array.

the warnings are:
```bash
zero/rescale_ghost_jacf.c:73:32: warning: variable length array folded to constant array as an extension [-Wgnu-folding-constant]
   73 |     double jacghost_surf_inv_c[num_basis_conf_surf];
      |                                ^~~~~~~~~~~~~~~~~~~
zero/rescale_ghost_jacf.c:83:24: warning: variable length array folded to constant array as an extension [-Wgnu-folding-constant]
   83 |       double jf_surf_c[num_basis_phase_surf];
      |                        ^~~~~~~~~~~~~~~~~~~~
```

This PR simply put the hardcoded values directly in the array definitions instead of declaring it above through a variable. Also I think that it is better since we see now directly at the line that the array have hardcoded size.

BTW there is still this other warning:
```bash
apps/mom_species.c:460:23: warning: comparison of different enumeration types ('enum gkyl_eqn_type' and 'enum gkyl_wv_mhd_div_constraint') [-Wenum-compare]
  460 |       if (sp->eqn_type==GKYL_MHD_DIVB_GLM) {
      |           ~~~~~~~~~~~~^ ~~~~~~~~~~~~~~~~~
``` 
which I think requires a deeper fix.

### DIFF
--- a/gyrokinetic/zero/rescale_ghost_jacf.c
+++ b/gyrokinetic/zero/rescale_ghost_jacf.c
@@ -53,8 +53,6 @@ gkyl_rescale_ghost_jacf_advance(const struct gkyl_rescale_ghost_jacf *up,
   int rem_dir[GKYL_MAX_DIM] = {0};
   for (int d=0; d<cdim; ++d) rem_dir[d] = 1;
 
-  const int num_basis_conf_surf = 8; // MF 2024/11/12: Hardcoded to 2x p2 for now.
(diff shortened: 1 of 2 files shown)


manauref:
this is weird, I thought declaring it as a const int would avoid this warning

PR #848, opened 2025-10-03
Attempting to clean up the final Makefile in the hierarchal build

JunoRavin:
We need to not duplicate symbols on GPUs by only linking against the final app requested by the user. I still think there are potential issues with the way LDFLAGS are specified though I'd like to get @ammarhakim's thoughts.


JonathanGorard (approved):

ammarhakim (approved):
Ready to merge

Issue #849, opened 2025-10-03
The adaptive sources are introducing an error in particle conservation


Antoinehoff:
This issue is raised because of some DIII-D simulations with adaptive sources showing a significant variation of the number of particles even if the adaptive sources are meant to maintain it constant. This problem was judged benign in TCV simulation but is of much higher amplitude in the DIII-D simulations.

### Verification of conservation of particle in the DIII-D 2x2v regression test
Currently, the regression test `rt_gk_d3d_iwl_2x2v_p1` is conserving the number of particle to machine precision (see Fig.1 and 2).

<img width="478" height="328" alt="Image" src="https://github.com/user-attachments/assets/1c2d8efc-b88c-4b63-9c7e-808db0eb09a9" />

_Fig.1: Particle balance in `rt_gk_d3d_iwl_2x2v_p1` _

<img width="478" height="328" alt="Image" src="https://github.com/user-attachments/assets/dd7c7778-8d68-45a5-adc0-f4104bad7de4" />

_Fig.2: Particle balance relative error in `rt_gk_d3d_iwl_2x2v_p1`_

Stemming from this test, we add adaptive sources and study how the balance error is affected. We set up one source per species that needs to compensate the losses of particle through each domain boundary:
```C
  // Source adaptation parameters.
  struct gkyl_gyrokinetic_adapt_source adapt_srcCORE_e ={
    .adapt_to_species = "elc",
    .adapt_particle = true,
    .adapt_energy = false,
    .num_boundaries = 4, // Only the inner radial boundary.
    .dir = {0, 0, 1, 1},
    .edge = {GKYL_LOWER_EDGE, GKYL_UPPER_EDGE, GKYL_LOWER_EDGE, GKYL_UPPER_EDGE},
  };
  struct gkyl_gyrokinetic_adapt_source adapt_srcCORE_i ={
    .adapt_to_species = "ion",
    .adapt_particle = true,
    .adapt_energy = false,
    .num_boundaries = 4, // Only the inner radial boundary.
    .dir = {0, 0, 1, 1},
    .edge = {GKYL_LOWER_EDGE, GKYL_UPPER_EDGE, GKYL_LOWER_EDGE, GKYL_UPPER_EDGE},
  };
```
Figure 3 shows that this set up breaks particle conservation. Note that this conservation error is not spectacular but may result in a non-negligible particle number deviation in production runs where the error accumulates over >1M time steps. 

[...]

PR #856, opened 2025-10-08
Conservative method to adapt the source

Antoinehoff:
As observed in DR #849, the current state of the adaptive source algorithm is introducing error in the particle balance. This is related to the importance of placing the source adaptation at the right place in the time stepping algorithm, so that the source is adapted to losses at the same numerical time.

## Experimental evidences

Here we take as basis the [TCV regression test](https://github.com/ammarhakim/gkeyll/blob/main/gyrokinetic/creg/rt_gk_tcv_iwl_adapt_source_2x2v_p1.c) as study case. We increase the resolution to 24x12x12x8 to be closer to a production scale run and we set the electron recycling source to compensate electron loss (instead of ions) to illustrate the compensation of fast parallel electron loss at the initialization phase.

#### Conservation of particles in `main`

Figure 1 shows that the particle conservation is currently not holding in the `main`. The error is decreasing and it is actually related to the time variation of the source (see Fig. 2), which gives a hint that the issue resides in the time integrator.

image

_Fig1: Error on the particle balance with the `main` branch._

image

_Fig2: Evolution of the source particle injection rate and the loss of particle through every boundaries for electrons (left) and ions (right) with the `main` branch._

#### Conservation of particles in `gk_conservative_adapt_src`

Figure 3 shows that the particle conservation holds in the PR branch. The flux and source amplitude presented in Fig. 4 are extremely similar to Fig. 2 but now the particle balance is conserved up to machine precision.

image

_Fig3: Error on the particle balance with the `gk_conservative_adapt_src` branch._

image

_Fig4: Evolution of the source particle injection rate and the loss of particle through every boundaries for electrons (left) and ions (right) with the `gk_conservative_adapt_src` branch._

## Theoretical explanation of the problem
Let's consider the SSP RK2 algorithm of Eq. 8 in https://arxiv.org/pdf/1708.02595 (the following derivation will hold true for the SSP RK3) and let's simplify the RHS of our equation to be only the source $S(f)$, where $f$ is the state of the system. The scheme writes

$$
f^{(1)} = f^n + \Delta t S(f^n),
$$
$$
f^{n+1} = \frac{1}{2}f^n + \frac{1}{2}\left[f^{(1)} + \Delta t S(f^{(1)})\right].
$$

which yields the simple update rule

$$
f^{n+1} = f^{n} + \frac{1}{2} \Delta t \left[S^{n} + S^{(1)}\right]
$$

If the source is constant, we have $S^{n} = S^{(1)}=S$ so the scheme is correct regardless how we treat the source in the RHS. However, if we have a time varying source, one needs to be careful. In `main`, the source is adapted after the last RK stage if the step is successful based on the `gk_species_bflux_get_flux` routine which provides the boundary fluxes in phase space `struct gkyl_array **gk_boundary_fluxes::flux`. The issue is that this variable `bflux->flux` is updated when `gk_species_bflux_rhs_calc` is called. Consequently, the source is adapted to the values of the boundary fluxes at the last RK stage, i.e. the. flux depending on $f^{(1)}$ for our RK2 example above.

### Solution
This PR proposes to solve this issue by adapting the source only [in the first call](https://github.com/ammarhakim/gkeyll/blob/52274682f296cf2ab772b0a75f126aa5d98c2924/gyrokinetic/apps/gyrokinetic_update_ssp_rk3.c#L87) of `gyrokinetic_forward_euler` in the RK3 SSP updater. It means that we have $S^{n}$ in all calls of the RHS function. This conserves particle but introduces a slight error in loss compensation as the source is evolved as in an explicit Euler scheme. One can think of more advanced method to adapt the source later.

### Implementaton
For now this is, I think, badly implemente. I introduced a new boolean in the arguments of the `gyrokinetic_forward_euler` routine to indicate if we adapt the source or not. I think a better way would be to adjust the update rule of the `bflux->flux` array but I do not know how this will interact with other places of the code yet.


JunoRavin (commented):
One thing I'm wondering about at this point: 
The adaptive source is a relatively complex program flow, made slightly more complex at the moment by the fact that it is trying to use a bunch of existing infrastructure for boundary fluxes and regular sources. 

What if we split gk_species_source.c into gk_species_const_source.c and gk_species_adapt_source.c, similar to how we've broken apart other objects (like the gk_field PR under review and the gk_collisionless PR under review) and then we even added to the separate gk_species_adapt_source updater creation that is specific to adapt source (like why not have the adapt_source create its *own* boundary fluxes based on the gk_species_bflux in the same way that many parts of the app create their own gk_species_moment). 

I think there is probably a design here where there is some small amount of code duplication, but a clean separation between these two different source types that then streamlines what and how the adapt source computes what it needs.

gyrokinetic/apps/gk_species_bflux.c
```diff
@@ -681,19 +696,19 @@ gk_species_bflux_init(struct gkyl_gyrokinetic_app *app, void *species,
...
+        || (bflux->calc_mom_names[m] == GKYL_F_MOMENT_HAMILTONIAN));
       bflux->is_hamiltonian_mom[m] = bflux->calc_mom_names[m] == GKYL_F_MOMENT_HAMILTONIAN;
       bflux->a_hamiltonian_mom = bflux->a_hamiltonian_mom || bflux->is_hamiltonian_mom[m];
     }
 
     if (need_m2perp) {
       // For moments that contain M2perp=2*mu*B/m we must fill the ghost cell of B. Use the option
       // in bc_basic that fills the ghost cell by evaluating the skin cell at the boundary.
-      long buff_sz = 1;
+      long buff_sz = 0;
```
manauref: this can cause a problem. Some machines don't like it when you allocate size 0 memory. It used to be 0 (either here or in gk_species), and I changed it to 1 after seeing some valgrind errors

manauref:
> One thing I'm wondering about at this point: The adaptive source is a relatively complex program flow, made slightly more complex at the moment by the fact that it is trying to use a bunch of existing infrastructure for boundary fluxes and regular sources.
> 
> What if we split gk_species_source.c into gk_species_const_source.c and gk_species_adapt_source.c, similar to how we've broken apart other objects (like the gk_field PR under review and the gk_collisionless PR under review) and then we even added to the separate gk_species_adapt_source updater creation that is specific to adapt source (like why not have the adapt_source create its _own_ boundary fluxes based on the gk_species_bflux in the same way that many parts of the app create their own gk_species_moment).
> 
> I think there is probably a design here where there is some small amount of code duplication, but a clean separation between these two different source types that then streamlines what and how the adapt source computes what it needs.

PR #864, opened 2025-10-16
Add skip_cell_threshold parameter to collisionless flux functions

Maxwell-Rosen:
- Introduced skip_cell_threshold to gkyl_gk_collisionless_flux_new and related functions to improve handling of small values in flux calculations.
- Updated relevant structures and CUDA implementations to accommodate the new parameter.

I noticed that the skip_cell_threshold feature was not working in some production runs and it turns out that it wasn't working at all! There was a new function introduced in the geometry reorganization that neglected this feature in the gk_collisionless_flux_surf. The functionality is added in and things look normal. The regression test has this feature as well.

Compiled and tested locally (CPU) and on Stellar-amd (GPU)

The regression test is rt_gk_wham_1x2v and I'm using a threshold value of 1e-16 for both species.
Selecting --z2 1.0 selects the maximum mu, where the particles experience the most $\mu \nabla B$ force, so very little distribution function makes it to the mirror throat.

Signs that the feature works are that the CFL rate goes to zero where the distribution function is less than the threshold, 1e-16.

pgkyl gk_wham_1x2v_p1-elc_cflrate_10.gkyl sel --z2 1.0 pl
image
pgkyl gk_wham_1x2v_p1-elc_10.gkyl interp sel --z2 1.0 pl --logz --zmin 1e-18
image

 pgkyl gk_wham_1x2v_p1-ion_cflrate_10.gkyl sel --z2 1.0 pl   
image
 pgkyl gk_wham_1x2v_p1-ion_10.gkyl interp sel --z2 1.0 pl --logz --zmin 1e-18
image


Maxwell-Rosen:
Here is an example simulation. I'm running here a Gkeyll Mirror simulation with the lorentzian geometry, comparing a simulation with skip_cell_threshold=1e-16 to one without the feature. They are using the pseudo orbit averaging scheme and have been run for 3s in the OAP and 100μs in FDP. Both simulations agree very well and the skip_cell feature reduces heating in the expander.

image

Issue #869, opened 2025-10-21
Neutral rescaling fails with implicit BGK collisions in gk_fluid_neut-prototyping_scaling branch.


SjoerdPR:
In the `gk_fluid_neut-prototyping_scaling` branch, if BGK collisions are used with the implicit collision scheme, the neutral profile rescaling does not occur. Instead, the neutral profile is kept completely static. The user is given no warning or error to notify them the rescaling is not working.

Minimal reproduction: in `rt_gk_sheath_fluid_neut_1x2v_p1.c` replace `.collision_id = GKYL_LBO_COLLISIONS,` by `.collision_id = GKYL_BGK_COLLISIONS,` and add `.has_implicit_coll_scheme` in the `.collisions` block of the `elc` and `ion` species. The rescaling of the `D0` profile now does not occur. 

EDIT: interestingly, the `gk_neut_species_rrs_apply_enabled` appears to still be called, and calculates its scaling factor properly.

manauref:
Thanks @SjoerdPR 
There was a flag missing in the regression test. I'll double check to make sure that it works.

manauref:
@SjoerdPR please switch to the `gk_fluid_neut-production` branch and try it there please. I just made some changes that should enable using the rescaled neutrals with implicit BGK collisions for charged species.

FYI @tnbernard

Issue #790, opened 2025-07-18
[DR] Unify species types in GK app, use it for fluid neutrals (a shell to do static or EIRENE/DEGAS neutrals)


manauref:
Current status of neutrals in GK app
===========================

The implementation of an early version of Canonical PB neutrals in the GK app was needed since our first version of dynamic neutrals (T. N. Bernard, et al. PoP 29, 052501 (2022)) could not work in 2D. However, in doing so we lost the ability to do static neutrals in geometries other than Cartesian. This will be fixed eventually, with a more mature version of the Canonical PB solver. But even then, it'll be a difficult to simulate neutrals including detailed wall geometry (from CAD), which some users expect. The latter reason is strong motivation to couple to other codes. The coupling does not deprecate Gkeyll neutrals in the least: Gkeyll dynamic neutrals can be used up to the Gkeyll domain boundary coupled to a neutral code evolving neutrals between the Gkeyll boundary and the CAD wall. Gkeyll neutrals may also be necessary to step neutrals internally for some period of time if Gkeyll is coupled to a neutral code loosely.

Charged and neutral species in the GK app are defined through structs like
```
struct gk_species {
  ...
  struct gkyl_array *f;  \\ Distribution function.
  struct gk_species_moment *moms; // Compute diagnostic moments.
  ...
};

struct gk_neut_species {
  ...
  struct gkyl_array *f;  \\ Distribution function.
  struct gk_species_moment *moms; // Compute diagnostic moments.
  ...
};
```
and throughout the code we have pointers and function signatures that differentiate between `gk_species` and `gk_neut_species`. One downside of having two separate types is that it makes species parallelization a little harder. Also it sets the precedent to introduce fluid neutrals through yet a third species, the way a fluid species exists in the Vlasov app, e.g. something like
```
struct gk_fluid_neut_species {
  ...
  struct gkyl_array *f;  \\ Moments
  struct gk_species_moment *moms; // Compute diagnostic moments.
  ...
};
```
This approach however:
1) Will make species parallelization even harder
2) Will require changes to some modules, like `gk_species_react`, so that they work with both `gk_neut_species` and `gk_fluid_neut_species`, or require us to write two versions of `gk_species_react`, one for each type of neutral. Either way is not ideal because it introduces more mods or more code.

Proposed alternative
================

We propose to unify species types into `gk_species` using anonymous unions such as:
```
struct gk_species {
  ...
  struct gkyl_array *f;  \\ Distribution function.
[...]

Issue #775, opened 2025-07-04
[DR] Set boolean to create diagonal metrics for static gk_neut_species


tnbernard:
This DR proposes a flag `.is_cartesian` to set the metric coefficients of a static gk neutral species to be diagonal. A high-phase-space resolution is necessary to project the desired moments for a gk_neut_species in general geometry. However, a memory limit is reached on GPUs when coupling to highly resolved GK species. A velocity resolution of 8^3 produces the following density and temperature profiles with the general geometry metrics:

![Image](https://github.com/user-attachments/assets/3abd28e2-4de0-4d1e-ab73-a1bae752c0c3)
![Image](https://github.com/user-attachments/assets/c2b83c89-731e-4cc7-b96d-5447f1e6626f)

The density has anomalies in xmax at the z boundaries and a dependency in z, where it should be constant. Any higher resolution produces a CUDA Malloc error on GPUs. On the other hand, a Cartesian metric produces the expected moments at the same: 

![Image](https://github.com/user-attachments/assets/73392363-09dc-4b20-a151-51eb484151d2)
![Image](https://github.com/user-attachments/assets/3a1196b8-f387-439a-a462-15f9c05d9e70)

Furthermore, the resolution can be reduced to 6^3:

![Image](https://github.com/user-attachments/assets/1bbc8ff2-afa7-468d-8015-9899663a8cde)

This DR has been prototyped on branch: https://github.com/ammarhakim/gkylzero/tree/neut-cart-metric

Issue #613, opened 2025-02-19
Make vdim a species attribute instead of an app atribute


manauref:
Currently `vdim` belongs to the app, but neutrals have a different `vdim`, and they end up having to do things like `app->vdim+1`. It'd be better to just move vdim to species.

JunoRavin:
Alternatively we can just set vdim to be 3 and throw an assertion if someone doesn't specify a three dimensional velocity grid for neutrals, since in the GK system we will never run with vdim < 3 for neutrals.

manauref:
It'd be better to just have vdim live in species. It is a species attribute after all.

PR #873, opened 2025-10-23
Fluid/kinetic neutrals infrastructure, scaled neutrals option, and improvements to boundary fluxes to use get_flux_mom

manauref:
Following the design in PR #790 , we have extended the neutrals solver in the gyrokinetic app so that it can support fluid neutrals. This serves two immediate purposes, and one long term purpose:
Immediate benefits:
- Run with static neutrals, without needing to worry about whether the distribution function is resolved well, or without even needing to allocate distributions (saves memory).
- Enable an option where the neutral density is simply scaled every time step according to a balance of ionization and recycling (more below).
Long term benefits:
- Use (cheaper) fluid instead of kinetic neutrals when needed on when the parameter space allows it.

On the development side, there are two other advancements here:
1. Following lessons by @Antoinehoff and the adaptive sources, we modify the boundary_flux app and the stepping/rhs functions, so that we pass the current RK-stage's moments of the boundary fluxes to more places. This is because previously the `_boundary_flux_get_mom` method was only returning the flux moments from the end of the step (i.e. `bflux->f`), but some capabilities (e.g. adaptive sources, scaled neutrals, Boltzmann electrons), care about having the instantaneous flux moment (i.e. `bflux->f`, `bflux->f1` or `bflux->fnew` depending on the RK stage).
2. We show that we can incorporate two different species, in this case kinetic and fluid neutrals, within the same struct (i.e. gk_neut_species) by using function pointers and anonymous struct unions. I believe that, in future work, we can actually remove gk_neut_species and place both charged and neutral species within `gk_species`. This will simplify many function signatures, remove some function pointer duplication, and more importantly, it'll make species parallelization easier.

Also, keep in mind that fluid neutrals are implemented with a similar formalism to that used by the DG fluid solver in the Vlasov app. This means that the state vector (or rather, `gkyl_array`) contains the fluid moments $rho$ (mass density), momentum density $\rho\mathbf{u}$ and total energy $\mathcal{E}$, allowing us to use some of the updaters already implemented for DG fluids. We additionally implemented methods to obtain Maxwellian moments ($n$, $\mathbf{u}$, $T/m$) from the state vector, so that the fluid neutrals can interact with charged species using the existing infrastructure, as do kinetic neutrals.

This PR also addresses issues #869, #775, #613 and issue #876 . It also incorporated PR #856 which improves the conservative accuracy of adaptive sources. This PR goes with [PR 91 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/91).

Fluid neutral interface
----------------------

Requesting fluid neutrals in gyrokinetic input files is done in a manner identical to requesting dynamic neutrals, all one is to do is NOT provide the `lower`, `upper` and `cells` entries defining the velocity space grid, e.g.
```
  struct gkyl_gyrokinetic_neut_species D0 = {
    .name = "D0",
    .mass = ctx.mass_ion,
    .vdim = 0,
    .gas_gamma = ctx.gas_gamma,
    ...
  }
```
Also, there are two new entries here:
- The `vdim` entry, which was previously given to the (conf-space) app, is not given to each species (see issue #613).
- Fluid neutrals require specification of the adiabatic index (`gas_gamma`, e.g. 5/3 for monoatomic hydrogen).

Semi-dynamic/scaled neutrals
------------------------------

This semi-dynamic, reduced neutral model assumes that you have an initial profile for the neutral moments, $n_n$, $\mathbf{u}_n$, $T_n$, and every time step the initial neutral density $n_{n0}(x)=n_n(t=0)$ is rescaled by the ratio of the total ion fluxes to the wall to the volume integrated ionization source:
Screenshot 2025-10-23 at 10 21 55 AM

This option is requested in the input file through a table similar to that used for adaptive sources, specifying which is the ion hitting the wall, and which walls we want to integrate over:
```
    .recycling_reaction_scaling = {
      .type = GKYL_GK_SPECIES_SCALING_RECYCLING_IZ_BALANCE,
      .impacting_ion_name =  "ion" ,
      .impacting_ion_id = GKYL_ION_H,
      .electron_name = "elc",
      .recycling_coeff = ctx.rec_frac,
      .num_boundaries = 2,
      .boundaries_dir = {0, 0,},
      .boundaries_edge = {GKYL_LOWER_EDGE, GKYL_UPPER_EDGE,},
    },
```
An example is provided in `rt_gk_sheath_fluid_neut_1x2v_p1.c`.

This can be used as we've used it so far, or in the future, we could evolve the neutrals self-consistently (through Gkeyll or an external PIC/MC code) for a few time steps, and in-between self-consistent updates we could scale the neutrals this way.

Static neutral demonstration
-----------------------------

Fluid neutrals have been made to work with the existing infrastructure to interact with charged species (e.g. through charge exchange, ionization, recombination). That, together with the fact that they don't take up much memory and don't have to worry about velocity-space representation, makes it easy to use them as a static yet reactive species.

We tested this by running an LAPD simulation with static, reactive hydrogen:
Screenshot 2025-10-23 at 10 34 43 AM
Notice the density is much larger when neutral fueling is added and, concurrently, the temperature is lower
Screenshot 2025-10-23 at 10 39 37 AM
I show only $T_{\parallel e}$ because this sim was done with hydrogen (unlike Eric Shi's He sims), with realistic collision frequency, so the collisionality is very high and the electrons are very Maxwellian.

Semi-dynamic neutral demonstration
-------------------------------------

@tnbernard has used scaled hydrogen neutrals in a 3x turbulent simulation of DIII-D. Here are some profiles:

image (13)
image (14)

### DIFF
--- a/core/zero/gkyl_eqn_type.h
+++ b/core/zero/gkyl_eqn_type.h
@@ -20,6 +20,7 @@ enum gkyl_eqn_type {
   GKYL_EQN_GR_MAXWELL_TETRAD, // General relativistic Maxwell equations in the tetrad basis.
   GKYL_EQN_GR_MEDIUM, // Coupled fluid-Einstein equations in plane-symmetric spacetimes.
   GKYL_EQN_GR_TWOFLUID, // General relativistic two-fluid equations.
+  GKYL_EQN_GR_TWOFLUID_TETRAD, // General relativistic two-fluid equations in the tetrad basis.
   GKYL_EQN_GR_MHD, // General relativistic magnetohydrodynamics equations.
   GKYL_EQN_GR_MHD_TETRAD, // General relativistic magnetohydrodynamics equations in the tetrad basis.
   GKYL_EQN_REACTIVE_EULER, // Reactive Euler equations.
--- a/gyrokinetic/apps/gkyl_gyrokinetic_run.h
+++ b/gyrokinetic/apps/gkyl_gyrokinetic_run.h
@@ -1,7 +1,6 @@
 #include <gkyl_gyrokinetic.h>
 #include <gkyl_gyrokinetic_multib.h>
-
-struct gkyl_gyrokinetic_timings {
+struct gkyl_gyrokinetic_time_stepping_inp {
   double t_end; // End time for the simulation
   int num_frames; // Number of output frames. Output every (t_end/num_frames) time units.
   double write_phase_freq; // Frequency (in multiples of num_frames) of writing phase-space data. e.g. 0.2 means write every 5 frames.
@@ -12,6 +11,12 @@ struct gkyl_gyrokinetic_timings {
   bool is_restart; // Is this a restarted simulation?
   int restart_frame; // Frame to restart from.
   int num_steps; // Maximum number of time-steps to take.
+};
+
+struct gkyl_gyrokinetic_run_verbosity_inp {
+  bool enabled; // Is verbosity enabled? Prints information every time step. Defaults false
+  double frequency; // Print information with given frequency. Defaults to 1.0
+  bool estimate_completion_time; // Estimate completion time based on current progress. Defaults false
 };
 
 enum gkyl_gyrokinetic_run_app_type {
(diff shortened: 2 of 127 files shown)


Antoinehoff:
By restarting production-like IWL TCV and DIII-D simulations, we observed a significant change of dynamics when using this branch. We found out that this issue is present in `main` and it is not related to this branch. This will be investigate in another DR/PR.

PR #879, opened 2025-10-28
BUGFIX: Position map constant B mapping

Maxwell-Rosen:
There was a bug in the region detection where it would get stuck if one side is too close to the z which the function is trying to evaluate. I added an else{} statement to the if statement, as well as some elseif statements for finding if we are too close to the boundary.

Here is an example of a simulation which was hanging before that is now producing a good grid. This is mc2nu_deflated.gkyl
image

### DIFF
--- a/gyrokinetic/zero/gkyl_position_map_priv.h
+++ b/gyrokinetic/zero/gkyl_position_map_priv.h
@@ -655,7 +655,7 @@ position_map_constB_z_numeric(double t, const double *xn, double *fout, void *ct
         return;
       }
       else {
-        printf("Warning: Unexpected interval evaluation state in position_map_constB_z_numeric. Using theta directly.\n");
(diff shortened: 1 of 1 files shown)


gyrokinetic/zero/gkyl_position_map_priv.h
```diff
@@ -634,6 +644,21 @@ position_map_constB_z_numeric(double t, const double *xn, double *fout, void *ct
...
+        fout[0] = interval_lower;
+        return;
+      }
+      else if (fabs(interval_upper_eval) < 1e-14) {
+        // Upper evaluation is very close to zero
+        fout[0] = interval_upper;
+        return;
+      }
+      else {
+        printf("Warning: Unexpected interval evaluation state in position_map_constB_z_numeric. Using theta directly.\n");
```
manauref: use fprintf(stderr, ...

Issue #761, opened 2025-06-16
[DR] Features for GK multirate time stepping in systems with trapped-lost / banana-passing orbits


manauref:
In both tokamak and mirror problems there can exist a multiscale time stepping challenge in which two disparate time scales must be resolved:
a) A fast time scale set by parallel dynamics, e.g. parallel transit time, bounce time, and omega_H.
b) A slow time scale set by collisional thermalization, pitch angle scattering, or diffusive transport.
Unfortunately one has to step for long (slow) times, but with small time steps to resolve the short (fast) times scales, before reaching steady states. This causes certain simulations to have many many small time steps, and thus increased cost.

In this work we experiment with a form of multi-rate time-stepping proposed by @Maxwell-Rosen and @gwhammett. In order to so we add 3 capabilites, and experiment with 2 forms of time-stepping, testing them on the HTS mirror problem.

This work is prototyped in the `gk_damp_scale_hamil-prototyping` branch.

New capabilities added
==================
1. Multiply the collisionless terms by a scalar number $\alpha$, so that the equation evolved is like $\partial_t f = \alpha \left\lbrace H, ~ f\right\rbrace + C + S + D$.
2. Multiply the whole GK equation by a multiplicative function $\mathcal{M}(x,v)$.
3. Add a damping term to the RHS of the GK equation: $\partial_t f = \dots - \nu(x,v) f$.

Multi-rate stepping for trapped-passing orbits (Option 1)
===========================================

This option alternated between a trapped-equilibrium phase (TEP) and an exact dynamics phase (EDP).

TEP:
-----

We solve

$$\partial_t f = \mathcal{M}(x,v) \left[ \alpha \left\lbrace H, ~ f\right\rbrace + C + S + D \right]$$

with $\alpha\ll 1$ and $\mathcal{M}(x,v) = 1$ in trapped region or 0 in loss region. The time step here will be set by collisions and diffusion, and in principle we can take very large time steps compared to that required by the $\left\lbrace H, ~ f\right\rbrace$ term, by setting $\alpha\propto \Delta t_{collisions} / \Delta t_{Hamiltonian}$ but for accuracy and possibly stability it may be better to use a larger $\alpha$.

This set up can be accomplished with the following input file interface:
```
  struct gkyl_gyrokinetic_species ion = {
    ...
    .collisionless_scale_factor = ctx.alpha,

    .time_rate_multiplier = {
      .type = GKYL_GK_SCALING_USER_INPUT,
[...]

PR #884, opened 2025-11-03
Multirate time integration and pseudo orbit averaging for mirrors

manauref:
This PR follows an updated scheme of what's described in Issue #761 . It is primarily geared towards providing a pseudo orbit averaging (POA) algorithm for mirror simulations, but this approach can also be used to build multirate integrators for tokamak simulations too.

Model
=====
Consider the gyrokinetic equation with collisionless ($\{ H, ~ f \}$), collisional ($C$) and source ($S$) terms:

$$\partial_t f = \left\lbrace H, ~ f\right\rbrace + C + S$$

Under some conditions, this system may have two very disparate time scales:
a) Fast transit time scales, given by the ratio of the length of the system along the magnetic field and the thermal speed: $\tau_{trans} = L / v_t$.
b) Slow collisional time scales, given by the inverse of the collision frequency: $\tau_{coll} = \nu^{-1}$.

**Problem:** We wish to step the collisional time scales over large periods of time, with big time steps. If we were to step the whole equation above as we normally do, our time step would be limited by $\tau_{trans}$, forbidding us from taking big time steps of order $\tau_{coll}$.

**Solution**

The approach here is to perform time integration by alternating two types of phases, a full dynamics phase (FDP) and an orbit-averaged phase (OAP). In the FDP we evolve the whole kinetic equation as given above. In the OAP we evolve the equation

$$\partial_t f = M_{loss}\left(x,v\right)\left[\alpha\left\lbrace H, ~ f\right\rbrace + C + S\right]$$

where $\alpha \ll 1$ is a small parameter which slows down the parallel dynamics (i.e. decreases $\tau_{trans}$), and $M_{loss}(x,v)$ is the $\dot{f}$ multiplier function
```math
M_{loss}(x,v) = \begin{cases}
0 \qquad (x,v) \text{ is in loss cone/passing region}, \\
1 \qquad (x,v) \text{ is in confined/trapped region}.
\end{cases}
```

Implementation
============

From a high-level point of view, this type of time integration is performed by creating an input file that loops over `num_phases`, alternating between an FDP and an OAP, i.e. something like
```
for (int pit=0; pit<num_phases; pit++) {
  - Reset operators
  - Run phase
}
```
see for example `gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1.c`. The operators that need to be reset in this case are:
- The `fdot_multiplier` $M_{loss}$.
- The collisionless scale factor $\alpha$.
- The field solver (in case the OAP is performed with a static field).
- The positivity operator (in case one of the phases, likely the FDP, needs it on).

For this purpose we have added three new capabilities

Time rate multiplier
--------------------
Input files can request that the time, $df/dt = \dot{f} = $`fdot`, be multiplied by a function $M(x,v)$ ($M_{loss}$ above), by providing a table such as
```
.time_rate_multiplier = {
//  .type = GKYL_GK_FDOT_MULTIPLIER_USER_INPUT,
//  .profile = my_multiplier,
//  .profile_ctx = &ctx,
  .type = GKYL_GK_FDOT_MULTIPLIER_LOSS_CONE,
  .cellwise_const = true,
  .write_diagnostics = true,
},
```
in the species table. At the moment the user can either pass their own function using `type = GKYL_GK_FDOT_MULTIPLIER_USER_INPUT` and defining this multiplier through the `profile` option, or request the $M_{loss}$ multiplied described above with `.type = GKYL_GK_FDOT_MULTIPLIER_LOSS_CONE`; the latter option evolves in time as the trapped-passing boundary depends on the evolving potential.

The multiplier $M(x,v)$ is projected onto the basis, and at the end of the `gyrokinetic_rhs` we multiply $df/dt$ and $\omega_{CFL}$ by it.

Collisionless scale factor
-------------------------
The user can provide a scalar parameter to multiply collisionless terms by, using the `.scale_factor` ($\alpha$ above) entry in the `collisionless` table:
```
.collisionless = {
  .scale_factor = 0.01,
},
```
this is simply used to multiply $df/dt_{collisionless}$ and $\omega_{CFL,collisionless}$ by `scale_factor` within the collisionless app.

Reset methods
---------------
We have added `_reset` methods to a few operators, specifically:
- The field solver.
- The collisionless app.
- The fdot_multiplier app.
- The positivity enforcer app.
The purpose of a reset method is, generally speaking, to re-initialize the that app. This is done by passing the same input struct that use in the app's input table, but perhaps with different parameters. For example, the main app may initialize the field with
```
struct gkyl_gyrokinetic_field field = {
  .gkfield_id = GKYL_GK_FIELD_BOLTZMANN,
  .electron_mass = ctx.me,
  .electron_charge = ctx.qe,
  .electron_temp = ctx.Te0,
  .is_static = false,
};
```
but later, the OAP may wish to use a static field, so before running an OAP one would call
```
struct gkyl_gyrokinetic_field field_inp = {
    .gkfield_id = GKYL_GK_FIELD_BOLTZMANN,
    .electron_mass = ctx->me,
    .electron_charge = ctx->qe,
    .electron_temp = ctx->Te0,
    .polarization_bmag = ctx->B_p,
    .is_static = true,
  };

  gkyl_gyrokinetic_app_reset_field(app, t_curr, field_inp);
```
The contents of the `_reset` method vary from one app to another, depending on what they need to do to reset the operator in a memory safe way. For the field we simply reset some function pointers:
```
void
gkyl_gyrokinetic_app_reset_field(gkyl_gyrokinetic_app* app, double tm,
  struct gkyl_gyrokinetic_field field_inp)
{
  app->field->info.is_static = field_inp.is_static;
  app->field->update_field = !field_inp.is_static;
  if (app->field->update_field)
    app->calc_field_func = gyrokinetic_calc_field_update;
  else
    app->calc_field_func = gyrokinetic_calc_field_none;
}
```

The other `_reset` methods implemented are
```
void gkyl_gyrokinetic_app_reset_species_fdot_multiplier(gkyl_gyrokinetic_app* app, double tm,
  const char* species_name, struct gkyl_gyrokinetic_fdot_multiplier fdot_mult_inp);

void gkyl_gyrokinetic_app_reset_species_collisionless(gkyl_gyrokinetic_app* app, double tm,
  const char *species_name, struct gkyl_gyrokinetic_collisionless gkcls_inp)`

void gkyl_gyrokinetic_app_reset_species_positivity(gkyl_gyrokinetic_app* app, double tm,
  const char *species_name, struct gkyl_gyrokinetic_positivity pos_inp);
```
We will need other reset methods in the future, e.g. to set species static or dynamic on the fly, but for now this is the bare minimum needed for this first proof of concept.

Note that to facilitate the resetting of the positivity operator, we have moved the positivity enforcing operation to its own module (e.g. gk_species_positivity), and now input files must request it with a table such as
```
.positivity = {
  .type = GKYL_GK_POSITIVITY_SHIFT,
  .write_diagnostics = true,
},
```
in the species table (the `enforce_positivity` flag has been removed). This modularity will also make easier to extend the positivity options in the future

Demonstration
===========

We used this to evolve the profiles of a high temperature superconducting (HTS) mirror in a 1D simulation with a Lorentzian magnetic field, see M. Francisquez, et al. PoP 30, 102504 (2023). The input file follows `rt_gk_mirror_boltz_elc_poa_1x2v_p1` closely, except that in the simulation below we did not use the positivity operator at in either the FDP or the OAP.

Note that this simulation, as described in M. Francisquez, et al. "Conservative velocity mappings for discontinuous Galerkin kinetics" (https://arxiv.org/abs/2505.10754), if done without the multirate stepping (i.e. using FDP only), marches on 2 A100 GPUs at a rate of 18.125 hours per ms, meaning it would've taken 652 hours to reach an ion-ion collision time ($\tau_{ii}=\nu_{ii}^{-1}\approx 36$ ms). Instead, with the multirate integrator we can reach 36 ms, on 4 A100 GPUs, in 9.9 hours. This means
- A 33X reduction in the cost of the simulation.
- A 66X reduction in the wallclock duration of the simulation.
Note however that @Maxwell-Rosen has done a lot more simulations and found more optimal ways to run this algorithm (e.g. set $\alpha$, and the duration of each phase) and has in some cases achieved a speed up of 20,000X-30,000X.

Here's a plot of the ion density at the mirror center ($z=0$), mirror throat ($z_m$), and expander wall $z_{max}$:
Screenshot 2025-11-02 at 9 14 56 PM

An important figure of merit is what the electrostatic potential does, as it is responsible for confining the electrons and can be compared to semi-analytic estimates (e.g. by Pastukhov). Here's the potential at these three locations:

Screenshot 2025-11-02 at 9 25 57 PM

Alternative algorithm with a damping term
------------------------------------------

As mentioned in issue #761 , we experimented with an OAP that evolves

$$\partial_t f = \alpha \left\lbrace H, ~ f\right\rbrace + C + S - (1-\alpha)\nu_{\parallel}\mathcal{M}(x,v) f$$

but this time $\mathcal{M}(x,v)=1$ for passing particles and 0 for trapped ones, with $\nu_{\parallel}\approx v_t/L_{\parallel}$. For this purpose we added a damping term,

$$\partial_t f = -\nu(x,v) f$$

which the user can specify through the `.damping` entry in the species table:
```
.damping = {
//      .type = GKYL_GK_DAMPING_USER_INPUT,
//      .rate_profile = loss_cone_damping_rate_profile,
//      .rate_profile_ctx = &ctx,
      .type = GKYL_GK_DAMPING_LOSS_CONE,
      .rate_profile = loss_cone_damping_rate_scaling,
      .rate_profile_ctx = &ctx,
    },
```
and just like with the `time_rate_multiplier`, the user can:
- Provide their own function for the damping rate $\nu(x,v)$, setting `.type = GKYL_GK_DAMPING_USER_INPUT,` and setting `.rate_profile` to the function that defines $\nu(x,v)$, or
- Set `.type = GKYL_GK_DAMPING_LOSS_CONE,` to use a damping function that is 1 in the loss/passing region and 0 in the confined/trapped region, multiplied by the function pointer the user provides in `rate_profile`.

Although our final algorithm doesn't use this term, we have kept it because in the future we wish to use this damping term to remove energy added by the positivity operator like Eric Shi did in the first version of Gkeyll.

### DIFF
--- a/gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
+++ b/gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
@@ -222,7 +222,6 @@ gkyl_loss_cone_mask_gyrokinetic_inew(const struct gkyl_loss_cone_mask_gyrokineti
     up->c2p_pos_ctx = up;
   }
   else {
-    assert(!up->use_gpu); // Need to set the c2p_pos function pointer in GPU.
     up->c2p_pos = inp->c2p_pos_func;
     up->c2p_pos_ctx = inp->c2p_pos_func_ctx;
   }


gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1.c
```diff
@@ -0,0 +1,1087 @@
...
+  // Calculate phase structure
+  double t_end = (tau_oap + tau_fdp)*num_cycles + tau_fdp_extra;
+  double tau_pair = tau_oap+tau_fdp; // Duration of an OAP+FDP pair.
+  int num_phases = 2*num_cycles + 1;
+  int num_frames = num_cycles * (num_frames_oap + num_frames_fdp) + num_frames_fdp_extra;
+
+  struct gk_poa_phase_params *poa_phases = gkyl_malloc(num_phases * sizeof(struct gk_poa_phase_params));
+  for (int i=0; i<(num_phases-1)/2; i++) {
+    // OAPs.
+    poa_phases[2*i].phase = GK_POA_OAP;
```
Maxwell-Rosen: I noticed this earlier, but I think this parameter can be removed. It's not used anywhere.

Or instead, we could use this enum to specify some default types of phases with specified properties so that we don't have to specify it's properties in this table and we just have an array of enum values. I don't really like that idea, but that's another way of doing the loop. i.e., instead of specifying all these things for a big array, we specify two types of phases, then the array populates with enums to specify which phase we are in. Inside the loop, it asks which enum we have, then recalls the properties of that type of phase. This might actually be cleaner than what we have because it simplifies the `for` loop here

gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1.c
```diff
@@ -0,0 +1,1087 @@
...
+    .int_diag_calc_freq   = int_diag_calc_freq  , 
+    .dt_failure_tol       = dt_failure_tol      , 
+    .num_failures_max     = num_failures_max    , 
+  };
+
+  // Populate a couple more values in the context.
+  ctx.psi_eval = psi_RZ(ctx.RatZeq0, 0., &ctx);
+  ctx.z_min    = z_psiZ(ctx.psi_eval, ctx.Z_min, &ctx);
+  ctx.z_max    = z_psiZ(ctx.psi_eval, ctx.Z_max, &ctx);
+  printf("\n");
```
Maxwell-Rosen: We should remove all print statements from the regression tests

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -0,0 +1,292 @@
...
+      else {
+        for (int d=0; d<app->cdim; d++)
+          bmag_max_coord_local[d] = -DBL_MAX;
+      }
+      gkyl_comm_allreduce_host(app->comm, GKYL_DOUBLE, GKYL_MAX, app->cdim, bmag_max_coord_local, bmag_max_coord_global);
+
+      if (app->use_gpu) {
+        fdmul->bmag_max = gkyl_cu_malloc(sizeof(double));
+        fdmul->bmag_max_coord = gkyl_cu_malloc(app->cdim*sizeof(double));
+	gkyl_cu_memcpy(fdmul->bmag_max, &bmag_max_global, sizeof(double), GKYL_CU_MEMCPY_H2D);
```
Maxwell-Rosen: This line is just one example, but I remember there being a whole bunch of funky indenting issues in these modules. Likewise with the loss_cone_mask zero updater. All the spacing should be consistent

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -0,0 +1,292 @@
...
+        .vel_map = gks->vel_map,
+        .bmag = app->gk_geom->geo_int.bmag,
+        .bmag_max = fdmul->bmag_max,
+        .bmag_max_loc = fdmul->bmag_max_coord,
+        .mass = gks->info.mass,
+        .charge = gks->info.charge,
+        .qtype = qtype,
+        .num_quad = num_quad,
+        .cellwise_trap_loss = cellwise_const,
+        .c2p_pos_func = 0, // Change for nonuniform position grid.
```
Maxwell-Rosen: Are there issues with this updater and non-uniform grids? I don't think there are because of how all this logic is in computational space. This comment may be misleading

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        // KEparDbmag = 0.5*mass*pow(vpar,2)/(bmag_max-bmag[0]).
+        double KEparDbmag = 0.0;
+        if (Dbmag_quad[cqidx] > 0.0)
+          KEparDbmag = 0.5*up->mass*pow(xmu[cdim], 2.0)/Dbmag_quad[cqidx];
+        else
+          KEparDbmag = 0.0;
+
+	double mu_bound = GKYL_MAX2(0.0, KEparDbmag+qDphiDbmag_quad[cqidx]);
```
Maxwell-Rosen: Here is another example of funky indenting

Maxwell-Rosen:
As of right now, I have verified that ./build/gyrokinetic/creg/rt_gk_wham_boltz_elc_poa_1x2v_p1 and /build/gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1 are valgrind and compute-sanitizer clean

Issue #867, opened 2025-10-17
[DR] "Get EM back!": an EMbitious but EMpowering project to EMjoy electromagnetic effects again, ahEM...


Antoinehoff:
_...I hope the PR will not create mayhEM..._

In this DR, we revive the GK electromagnetic (EM) capabilities, in particular the computation of the perpendicular magnetic fluctuations by evolving the parallel component of the magnetic potential vector $A_\parallel$. The following is mainly based on Sec. 2.3 of Noah's [thesis](https://arxiv.org/abs/2103.16062), and the equation numbering is kept.

The parallel magnetic vector potential enters in the GK equation as

<img width="782" height="107" alt="Image" src="https://github.com/user-attachments/assets/da5f2e23-5db4-42db-bff7-65eeaf71fc75" />

where the * term is the electrostatic part of the RHS (the RHS currently implemented in Gkeyll).
One can obtain $A_\parallel$ through the Ampere's law

<img width="716" height="83" alt="Image" src="https://github.com/user-attachments/assets/b106a4cc-9705-4897-83cb-24fef50e1d92" />

However, cancellation issues will rise from this equation (see thesis). It is hence more tractable to solve the Ohm's law for the parallel component of the inducted electric field $E_\parallel$,

<img width="775" height="104" alt="Image" src="https://github.com/user-attachments/assets/0d1912b9-84a2-4313-a679-4ccbb3311b66" />

Following the GK equation Eq. (2.124), the Ohm's law can be written as an Helmholtz equation,

<img width="949" height="108" alt="Image" src="https://github.com/user-attachments/assets/69e5f2fa-3d22-4140-85ff-175ba50a11ad" />

Once the Ohm's law is solved, we can simply integrate $\dot A_\parallel$ in time using our RK scheme, to obtain $A_\parallel$. 
Once Eq. (2.125) solved and $A_\parallel$ stepped, we need to include $A_\parallel$ contribution in the Poisson bracket

<img width="851" height="146" alt="Image" src="https://github.com/user-attachments/assets/f99e2065-7863-441e-9175-dddde69b89f9" />

We can extract the electromagnetic contributions using the definition of the perturbed magnetic field,
```math
\boldsymbol B^*_h = \boldsymbol B^{*ES}_{h} + \delta \boldsymbol B_{\perp h},
```
with 
```math
[...]

PR #888, opened 2025-11-06
Get EM back

Antoinehoff:
This PR is related to the DR #867 and [gkylcas#89](https://github.com/ammarhakim/gkylcas/pull/89)

# Test cases
We report on a few test cases and benchmark, most of them taken from the work of @nmandell (see, e.g., [Mandell et al. 2020](https://w3.pppl.gov/~hammett/gyrofluid/papers/2020/Mandell%202020%20JPP%20EM%20Gkeyll%20NSTX.pdf) and [Mandell thesis](https://arxiv.org/abs/2103.16062))

## Alfven waves
We create the new associated regression tests.
```sh
gyrokinetic/creg/rt_gk_em_alfven_1x2v_p1.c
gyrokinetic/creg/rt_gk_em_alfven_2x2v_p1.c
gyrokinetic/creg/rt_gk_em_alfven_3x2v_p1.c
```
and verify that the results in [Mandell thesis](https://arxiv.org/abs/2103.16062) are retrieved.
kaw_scan_results

The scan was performed on Perlmutter using one CPU node and running many Gkeyll session in parallel using the following C input file and python scripting:
- [kaw_1x2v.c](https://github.com/user-attachments/files/27130290/kaw_1x2v.c) 
- [kaw_scan.py](https://github.com/user-attachments/files/27130307/kaw_scan.py) 
- [kaw_analysis.py](https://github.com/user-attachments/files/27130322/kaw_analysis.py)
- [gkeyll_interface.py](https://github.com/user-attachments/files/27130325/gkeyll_interface.py).

## 3x2v helical scrape-off layer
We create a new associated regression test
```sh
gyrokinetic/creg/rt_gk_em_helical_sol_3x2v_p1.c
```
and we verify that the results are similar to [Mandell et al. 2020](https://w3.pppl.gov/~hammett/gyrofluid/papers/2020/Mandell%202020%20JPP%20EM%20Gkeyll%20NSTX.pdf). 
*Note*: the geometry is adjusted to pass the right hand check (see PR #933).

_Integrated moments_
image

_Electron density and perp. magnetic fluctuations_

#### Helical SOL performance benchmark
We benchmark our different electromagnetic implementations with 3x2v shearless helical SOL turbulence, which uses a 3x2v phase-space grid with resolution $N_x=16$, $N_y=32$, $N_z=12$, $N_{v_\parallel}=10$, $N_\mu=5$. The table below summarizes the performance benchmarks for different versions of the solver, using 3 A100 GPUs on Perlmutter.

| Version | Throughput |
|---|---:|
| Full upwind, non-optimized | ~2.2 steps/s |
| w. Kernel vpar flux optim. | ~2.8 steps/s |
| w. Ohm's LHS updater (this branch) | ~7.7 steps/s |
| Electrostatic | ~9.1 steps/s |

image

_this branch_
image
_Mandel et al. 2020_
Screenshot 2026-04-30 at 3 08 04 PM

## Need to smooth $A_\parallel$ along $z$ in tokamak geometry
We now test EM effects on known simulation like TCV simulations. Before smoothing $A_\parallel$, it was not possible to run any TCV simulation for more than O(10) microseconds before observing a violent event breaking the simulation. This phenomenon was observed present in the following cases.
- TCV 2x2v SOL only (24x16x16x8)
- TCV 3x2v SOL only (24x16x12x12x8)
- TCV 2x2v clopen (24x16x16x8)
- TCV 3x2v clopen (24x16x12x12x8)

We tested a version where we apply $\dot A_\parallel \leftarrow \dot A_\parallel \langle \dot A_\parallel \rangle_{fs}$ at each Ohm solve to remove any zonal component of EM fluctuations. Suprisingly it did not even affect the time where the catastrophic event occur.

~The simulation crashes approx. 450 mus after the restart. It does not seem that it reached steady state as shown with the evolution of $A_\parallel$ which continuously drops.~ This is no longer valid and was solved by restoring the bias plane in the potential (lost during some input file editing...).

When $A_\parallel$ is smoothed just after the stepping using $\dot{A}_\parallel$, the TCV PT coarse simulation runs smoothly:
image

For comparison purposes, this is the same simulation but electrostatic:
image

and time averaged profile comparisons averaged between 0.5ms and 1ms:
Screenshot 2026-09-08 at 11 57 08 AM


Antoinehoff:
@JunoRavin, @manauref, I think this is ready for review. I don't think the design will change, the previous results were reproduced and the new tokamak geometry is not stable yet but it is capable. I would like to merge this so that I can stop resolving conflicts every week :)

Antoinehoff:
> Collaborator

@JunoRavin @manauref , a kind reminder now that the simulations are stable.

JunoRavin:
And to clarify on the smoothing $A_\parallel$ piece, the intuition for the improvement is what we talked about that this now makes the modification of the characteristics due to magnetic flutter continuous, without affecting the reason we leave $\partial A_\parallel/\partial t$ discontinuous in $z$ (so that $E_\parallel \sim 0$ as we expect for the Alfvén wave without any spurious issues from $\nabla_\parallel \phi$ being discontinuous).

PR #889, opened 2025-11-07
Refactor of skipping phase space cell updates

### LINKED ISSUE #878: Change the Gk skip_cell option to use a generic mask instead
At the moment the skip_cell method in the GK app introduces if-statements like
```
  if (fabs(qInEdge[0]) < gyrokinetic->skip_cell_thresh && fabs(qInSkin[0]) < gyrokinetic->skip_cell_thresh) {
     return 0;
  }
```
in nearly every single solver called in the time loop.

There's a better way to do this that is more extensible and easier to maintain:
1. Create a mask boolean gkyl_array, call it skip_cell.
2. At the beginning of every time step, or RK stage, check f in every cell. If it's below the threshold then cert skip_cell to true in that cell, else it remains false.
3. In every solver, instead of the if-statement above, have an if-statement like
```
  if (skip_cell) {
     return 0;
  }
```

This is more extensible because `skip_cell` may be set differently by other modules (e.g. perhaps we want to use it to handle non-conforming boundaries the way planets are handled in magnetospheric fluid simulations), but adding a different criteria for setting `skip_cell` will only require modifying a few lines in one place, while right now adding a new criteria for skipping cells requires modifying dozens of lines and maybe making over 100 edits. It's easier to maintain for the same reason.

Skip cell is a boolean array, so it'll occupy ncomp*8 times less memory than a distribution function. For 3x2v this means `skip_cell` will occupy 384 times fewer bytes than the distribution, so it shouldn't be a problem to add this to the app.

Maxwell-Rosen:
Update by @manauref 2026-02-12: ultimately we decided that the better way to do this is for the skip_cell object to have a vol, surf and boundary surf mask. Then the if-statements in each object just check a single condition. However we don't have time to do this right now, and we actually don't need the skip_cell in GK at the moment. So we deleted skip_cell from GK but kept the improved skip_cell in core, so it can be improved and implemented in GK later. 

------

This is a continuation of #864, where we added skip_cell to the new collisionless updaters. This PR aims to fix #878 most of all, where Mana commented that it would be better designed if we determined the mask in one place, then applied it to all the updaters. 

This means creating a new updater, called zero/skip_cell.c, which determines the mask. 

Array.c is modified to support boolean arrays, reducing memory usage for this project.

Modifications are made to pass this new skip_cell struct to each relevant place.

The input file is modified, so now we specify inside each species.
```
.skip_cell = {
  .threshold = 1e-16
}
```
Each species has its own skip-cell updater, which is updated at every RK stage during the Euler step.

Future work may be interested in extending this mask to the moment calculators; however, I'm not sure how to do so within the current infrastructure. For instance, there are many calls to the moment updaters, and I don't know how each one should be changed.

Regression test results showing that the skip_cell works fine.

pgkyl gk_wham_1x2v_p1-elc_cflrate_10.gkyl sel --z2 1.0 pl
image

pgkyl gk_wham_1x2v_p1-elc_10.gkyl interp sel --z2 1.0 pl --logz --zmin 1e-18
image

pgkyl gk_wham_1x2v_p1-ion_cflrate_10.gkyl sel --z2 1.0 pl
image

pgkyl gk_wham_1x2v_p1-ion_10.gkyl interp sel --z2 1.0 pl --logz --zmin 1e-18
image

### DIFF
--- a/gyrokinetic/zero/gk_collisionless_flux.c
+++ b/gyrokinetic/zero/gk_collisionless_flux.c
@@ -117,7 +117,6 @@ void gkyl_gk_collisionless_flux_surf(struct gkyl_gk_collisionless_flux *up,
 
     double *flux_surf_d = gkyl_array_fetch(flux_surf, loc_phase);
     double *cflrate_d = gkyl_array_fetch(cflrate, loc_phase);
-    double cfl_temp = 0.0;
 
     for (int dir = 0; dir<cdim; ++dir) {
       gkyl_copy_int_arr(pdim, idx, idxL);
@@ -135,14 +134,13 @@ void gkyl_gk_collisionless_flux_surf(struct gkyl_gk_collisionless_flux *up,
 
       if (idx[dir] == phase_range->lower[dir]) {
         // Lower domain/block boundary.
-        cfl_temp = up->flux_surf_edge_lo[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
+        cflrate_d[0] += up->flux_surf_edge_lo[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
           dgs, gkdgs, bmag_d, jacgeo_rat_surfL_d, jacgeo_rat_surfR_d, phi_d, fL, fR, flux_surf_d);
       } else {
         // Interior, lower cell surface.
-        cfl_temp = up->flux_surf[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
+        cflrate_d[0] += up->flux_surf[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
           dgs, gkdgs, bmag_d, jacgeo_rat_surfL_d, jacgeo_rat_surfR_d, phi_d, fL, fR, flux_surf_d);
       }
-      cflrate_d[0] += cfl_temp;
 
       // Upper domain/block boundary.
       // If the phase space index is at the local configuration space upper value, we
@@ -167,10 +165,8 @@ void gkyl_gk_collisionless_flux_surf(struct gkyl_gk_collisionless_flux *up,
 
         double* flux_surf_ext_d = gkyl_array_fetch(flux_surf, loc_phase_ext);
 
-        cfl_temp = up->flux_surf_edge_up[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
+        cflrate_ext_d[0] += up->flux_surf_edge_up[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
           dgs, gkdgs, bmag_d, jacgeo_rat_surfL_d, jacgeo_rat_surfR_d, phi_d, fL, fR, flux_surf_ext_d);
-
-        cflrate_ext_d[0] += cfl_temp;
       }  
     }
   }
@@ -194,7 +190,6 @@ void gkyl_gk_collisionless_flux_surf(struct gkyl_gk_collisionless_flux *up,
 
     double *flux_surf_d = gkyl_array_fetch(flux_surf, loc_phase);
     double *cflrate_d = gkyl_array_fetch(cflrate, loc_phase);
-    double cfl_temp = 0.0;
 
     int dir=cdim;
     gkyl_copy_int_arr(pdim, idx, idxL);
@@ -211,11 +206,8 @@ void gkyl_gk_collisionless_flux_surf(struct gkyl_gk_collisionless_flux *up,
 
     const struct gkyl_dg_vol_geom *dgv = gkyl_dg_geom_get_vol(up->dg_geom, idx);
     const struct gkyl_gk_dg_vol_geom *gkdgv = gkyl_gk_dg_geom_get_vol(up->gk_dg_geom, idx);
-
-    cfl_temp += up->flux_surfvpar[0](xc, up->phase_grid.dx, vpL, vpR, vmap_d, vmapSq_d, up->charge, up->mass,
+    cflrate_d[0] += up->flux_surfvpar[0](xc, up->phase_grid.dx, vpL, vpR, vmap_d, vmapSq_d, up->charge, up->mass,
       dgv, gkdgv, bmag_d, phi_d,  fL, fR, flux_surf_d);
-
-    cflrate_d[0] += cfl_temp;
   }
 }
 
--- a/gyrokinetic/zero/gk_collisionless_flux_cu.cu
+++ b/gyrokinetic/zero/gk_collisionless_flux_cu.cu
@@ -53,7 +53,6 @@ gkyl_gk_collisionless_flux_surf_conf_cu_kernel(struct gkyl_gk_collisionless_flux
 
     double* flux_surf_d = (double*) gkyl_array_fetch(flux_surf, loc_phase);
     double *cflrate_d = (double*) gkyl_array_fetch(cflrate, loc_phase);
-    double cfl_temp = 0.0;
 
     for (int dir = 0; dir<cdim; ++dir) {
       gkyl_copy_int_arr(pdim, idx, idxL);
@@ -71,15 +70,13 @@ gkyl_gk_collisionless_flux_surf_conf_cu_kernel(struct gkyl_gk_collisionless_flux
 
       if (idx[dir] == phase_range.lower[dir]) {
         // Lower domain/block boundary.
-        cfl_temp = up->flux_surf_edge_lo[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
+        cflrate_d[0] += up->flux_surf_edge_lo[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
           dgs, gkdgs, bmag_d, jacgeo_rat_surfL_d, jacgeo_rat_surfR_d, phi_d, fL, fR, flux_surf_d);
       } else {
         // Interior, lower cell surface.
-        cfl_temp = up->flux_surf[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
+        cflrate_d[0] += up->flux_surf[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
           dgs, gkdgs, bmag_d, jacgeo_rat_surfL_d, jacgeo_rat_surfR_d, phi_d, fL, fR, flux_surf_d);
       }
-
-      cflrate_d[0] += cfl_temp;
 
       // If the phase space index is at the local configuration space upper value, we
       // we are at the configuration space upper edge and we also need to evaluate 
@@ -103,10 +100,8 @@ gkyl_gk_collisionless_flux_surf_conf_cu_kernel(struct gkyl_gk_collisionless_flux
 
         double* flux_surf_ext_d = (double*) gkyl_array_fetch(flux_surf, loc_phase_ext);
 
-        cfl_temp = up->flux_surf_edge_up[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
+        cflrate_ext_d[0] = up->flux_surf_edge_up[dir](xc, up->phase_grid.dx, vmap_d, vmapSq_d, up->charge, up->mass,
           dgs, gkdgs, bmag_d, jacgeo_rat_surfL_d, jacgeo_rat_surfR_d, phi_d, fL, fR, flux_surf_ext_d);
-
-        cflrate_ext_d[0] = cfl_temp;
       }  
     }
   }
@@ -152,7 +147,6 @@ gkyl_gk_collisionless_flux_surf_surfvpar_cu_kernel(struct gkyl_gk_collisionless_
 
     double* flux_surf_d = (double*) gkyl_array_fetch(flux_surf, loc_phase);
     double *cflrate_d = (double*) gkyl_array_fetch(cflrate, loc_phase);
-    double cfl_temp = 0.0;
 
     int dir = cdim;
     gkyl_copy_int_arr(pdim, idx, idxL);
@@ -170,12 +164,10 @@ gkyl_gk_collisionless_flux_surf_surfvpar_cu_kernel(struct gkyl_gk_collisionless_
     const struct gkyl_dg_vol_geom *dgv = gkyl_dg_geom_get_vol(up->dg_geom, idx);
     const struct gkyl_gk_dg_vol_geom *gkdgv = gkyl_gk_dg_geom_get_vol(up->gk_dg_geom, idx);
 
-    cfl_temp += up->flux_surfvpar[0](xc, up->phase_grid.dx, 
+    cflrate_d[0] += up->flux_surfvpar[0](xc, up->phase_grid.dx, 
       vpL, vpR,
       vmap_d, vmapSq_d, up->charge, up->mass,
       dgv, gkdgv, bmag_d, phi_d,  fL, fR, flux_surf_d);
-
-    cflrate_d[0] += cfl_temp;
   }
 }
 


Maxwell-Rosen:
I've tested this on GPUs and there were no apparent errors. Valgrind tests are fine too

gyrokinetic/zero/skip_cell.c
```diff
@@ -0,0 +1,116 @@
...
+    gkyl_array_release(skip_cell->mask);
+  
+  if (GKYL_IS_CU_ALLOC(skip_cell->flags))
+    gkyl_cu_free(skip_cell->on_dev);
+  
+  gkyl_free(skip_cell);
+}
+
+struct gkyl_skip_cell*
+gkyl_skip_cell_new(struct gkyl_skip_cell_inp skip_cell_inp, struct gkyl_range phase_rng, bool use_gpu)
```
JunoRavin: Is there a reason phase_rng and use_gpu are not part of the skip_cell_inp?

Maxwell-Rosen: The reason is because the skip_cell_inp is the input struct from the input file, while the others are properties of the app

gyrokinetic/zero/skip_cell.c
```diff
@@ -0,0 +1,116 @@
...
+  // Initialize the mask array on host.
+  skip_cell->mask = gkyl_array_new(GKYL_DOUBLE, 1, phase_rng.volume);
+  gkyl_array_clear(skip_cell->mask, false);
+
+  skip_cell->flags = 0;
+  GKYL_CLEAR_CU_ALLOC(skip_cell->flags);
+  skip_cell->ref_count = gkyl_ref_count_init(gkyl_skip_cell_free);
+  skip_cell->on_dev = skip_cell; // CPU skip_cell points to itself
+
+  struct gkyl_skip_cell *skip_cell_out = skip_cell;
```
JunoRavin: I would not do the assignment here. Just do the declaration. 

Though my personal preference would be that this ifdef just be moved to the top of the function and just pass the input struct to the cu_dev_new method and immediately return the output of the cu_dev_new method like we do with other struct initialization methods.

gyrokinetic/zero/skip_cell.c
```diff
@@ -45,14 +45,7 @@ gkyl_skip_cell_new(struct gkyl_skip_cell_inp skip_cell_inp, struct gkyl_range ph
...
   
-  // Initialize all cells to false (not skipped).
-  struct gkyl_range_iter iter;
-  gkyl_range_iter_init(&iter, &skip_cell->phase_rng);
-  while (gkyl_range_iter_next(&iter)) {
-    long linidx = gkyl_range_idx(&skip_cell->phase_rng, iter.idx);
-    bool *skip_c = gkyl_array_fetch(skip_cell->booleans, linidx);
-    *skip_c = false;
-  }
+  gkyl_array_clear(skip_cell->booleans, false);
```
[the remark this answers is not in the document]
Maxwell-Rosen: This is important for the updaters which do not advance this mask. We want it initialized to be passive and do nothing

Maxwell-Rosen:
If we are discussing whether the mask should be (-1,1) or (0,1), I think (0,1) is clearer whether we update the cell or not. This makes the array boolean-like, where 0 (false) means we do not skip the cell update and 1 (true) means we do skip the cell update.

Maxwell-Rosen:
Yes. This is just a general mask for identifying whether `f[0]` is less than some threshold, creating a binary mask. I'm using this in the time-dilation scheme to create a mask as well. Perhaps we should re-think the `skip_cell` name.

ammarhakim:
@Maxwell-Rosen Can you make your mask field consistent with what we do in fluids? Very soon Grant will need to also do skip-cell-ing in the same way as Jonathan is for black-holes.

One way to make this transparent for you, if to define a function like:

`static inline bool update_cell_p(double val) { return val > 0; }`

or something like that. Then, you can use this function in your loops. If you decide to change the sign or value you can simply change the function and not your loops.

JunoRavin:
This is not a boolean mask. It is a mask for determining whether you are inside or outside a region, with the ability to define regions for updating cells or other conditions. At minimum, I need you to switch the convention to match moments. -1 means you are inside the region. 1 means you are outside the region. In this case, 1 being outside the region means we are updating cells and not skipping the updates.

core/zero/gkyl_dg_array_mask.h
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @param mask Mask object.
+ * @param idx Multi-dimensional index array.
+ * @return True if the mask is true at the index, false otherwise.
+ */
+bool gkyl_dg_array_mask_eval_idx(struct gkyl_dg_array_mask *mask, const int* idx);
+
+/**
+ * mask = mask * arr_to_multiply
+ * Scale the dg_array_mask by an array, modifying the mask.
```
manauref: swap lines

core/zero/gkyl_dg_array_mask.h
```diff
@@ -0,0 +1,116 @@
...
+void
+gkyl_dg_array_mask_scale_by_cell(struct gkyl_dg_array_mask *mask, const struct gkyl_array *arr_to_multiply);
+
+
+/**
+ * Acquire a reference to the mask object.
+ * 
+ * @param mask Mask object to acquire.
+ */
+
```
manauref: delete

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,317 @@
...
+      mask->threshold = mask_inp.threshold;
+    }
+    
+    // Initialize the mask array on host.
+    mask->phase_rng = mask_inp.phase_rng;
+    mask->phase_rng_ext = mask_inp.phase_rng_ext;
+    mask->conf_rng = mask_inp.config_rng;
+    mask->conf_rng_ext = mask_inp.config_rng_ext;
+    mask->vel_rng = mask_inp.vel_rng;
+    mask->mask = gkyl_array_new(GKYL_DOUBLE, 1, mask_inp.phase_rng_ext.volume);
```
manauref: additionally, you assume the mask always exists in phase_rng. But what if we want to use this with a fluid species? you should change the logic so that there's a private mask_rng, and the mask_rng gets assigned to conf_rng if phase_rng  is 0.

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,317 @@
...
+    mask_out = gkyl_dg_array_mask_cu_dev_new(mask);
+    gkyl_dg_array_mask_release(mask);
+  }
+#endif
+
+  return mask_out;
+}
+
+void
+gkyl_dg_array_mask_advance(struct gkyl_dg_array_mask *mask, const struct gkyl_array *arr_to_mask)
```
manauref: this method has a lot of if-statements for a variable that doesn't change in time. Have this method call a function pointer set in the _new method instead.

core/zero/gkyl_dg_array_mask_priv.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include <gkyl_dg_array_mask.h>
+#include <gkyl_ref_count.h>
+
+/**
+ * Skip cell object definition.
+ */
+struct gkyl_dg_array_mask {
+  enum gkyl_dg_array_mask_types type; // Type of mask operation.
+  bool default_value; // Default value for mask (true/false) if no masking is applied. Defaults to false.
```
manauref: see my suggestion to change the mask array to an INT type. Furthermore, `bool default_value` is confusing because as it is now the mask array is a double, and we'll change it to an int. But in any case, we could have the default value be anything double (for double array) or any int (for int array). So why make `default_value` a bool? seems like we are missing an opportunity here to be clearer AND allow the mask to take any value for other purposes.

core/unit/ctest_dg_array_mask.c
```diff
@@ -0,0 +1,1579 @@
...
+  TEST_CHECK( mask != NULL );
+  TEST_CHECK( mask->type == GKYL_DG_ARRAY_MASK_C0_LESS_THAN_THRESHOLD );
+  TEST_CHECK( mask->use_gpu == use_gpu );
+  TEST_CHECK( mask->mask != NULL );
+  TEST_CHECK( mask->mask->size == range.volume );
+  TEST_CHECK( mask->mask->ncomp == 1 );
+  
+  // Check that mask is initialized to -1.0 (false)
+  const struct gkyl_array *mask_arr = gkyl_dg_array_mask_get_mask(mask);
+  struct gkyl_array *mask_ho = mkarr(false, 1, range.volume);
```
manauref: please change all _ho array allocations to be like
a_ho = mkarr(false, a->ncomp, a->size)

core/zero/gkyl_dg_array_mask.h
```diff
@@ -0,0 +1,116 @@
...
+const struct gkyl_array*
+gkyl_dg_array_mask_get_mask(const struct gkyl_dg_array_mask *mask);
+
+/**
+ * Get device pointer to mask object.
+ *
+ * @param mask Mask object.
+ */
+struct gkyl_dg_array_mask*
+gkyl_dg_array_mask_get_dev_ptr(struct gkyl_dg_array_mask *mask);
```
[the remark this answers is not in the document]
Maxwell-Rosen: These getter functions are necessary because the type definition is inside the private header. I did this intentionally to ensure that no one misuses the memory inside the mask without authorized functions. For instance, the "get_mask" function returns a const array to prevent modification.
The device pointer is needed here because the CUDA files need it to copy to the device. It's along the lines of it being safer to use getter functions than to access memory directly, because we can control access.

core/zero/dg_array_mask.c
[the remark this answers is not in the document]
Maxwell-Rosen: I am using the scale_by_cell function for the time dilation stuff. It avoids allocating an extra array in that instance. It needs to be of type double so that it can multiply the CFL rate

gyrokinetic/apps/gk_neut_species_kinetic.c
```diff
@@ -676,6 +678,35 @@ gk_neut_species_kinetic_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *ap
...
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_FRAC_THRESHOLD_SPATIAL;
+  } else if (s->info.skip_cell.type == GKYL_GK_SKIP_CELL_ABOVE_FRAC_SPATIAL) {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_LESS_THAN_FRAC_THRESHOLD_SPATIAL;
+  } else if (s->info.skip_cell.type == GKYL_GK_SKIP_CELL_BELOW_FRAC) {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_FRAC_THRESHOLD;
+  } else if (s->info.skip_cell.type == GKYL_GK_SKIP_CELL_ABOVE_FRAC) {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_LESS_THAN_FRAC_THRESHOLD;
+  } else {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_NONE;
+  }
```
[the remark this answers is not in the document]
Maxwell-Rosen: Fixed. New header file created `gyrokinetic/zero/gkyl_dg_array_mask_gyrokinetic.h`

gyrokinetic/apps/gk_species.c
```diff
@@ -1382,6 +1382,35 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
...
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_FRAC_THRESHOLD_SPATIAL;
+  } else if (gks->info.skip_cell.type == GKYL_GK_SKIP_CELL_ABOVE_FRAC_SPATIAL) {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_LESS_THAN_FRAC_THRESHOLD_SPATIAL;
+  } else if (gks->info.skip_cell.type == GKYL_GK_SKIP_CELL_BELOW_FRAC) {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_FRAC_THRESHOLD;
+  } else if (gks->info.skip_cell.type == GKYL_GK_SKIP_CELL_ABOVE_FRAC) {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_C0_LESS_THAN_FRAC_THRESHOLD;
+  } else {
+    skip_cell_mask_type = GKYL_DG_ARRAY_MASK_NONE;
+  }
```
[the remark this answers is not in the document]
Maxwell-Rosen: Fixed. New header file created `gyrokinetic/zero/gkyl_dg_array_mask_gyrokinetic.h`

core/zero/gkyl_dg_array_mask.h
```diff
@@ -0,0 +1,116 @@
...
+typedef struct gkyl_dg_array_mask gkyl_dg_array_mask;
+
+enum gkyl_dg_array_mask_types {
+  GKYL_DG_ARRAY_MASK_NONE = 0, // No mask applied.
+  GKYL_DG_ARRAY_MASK_C0_LESS_THAN_THRESHOLD, // Mask applied based on the 0th component of the array.
+  GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_THRESHOLD, // Mask applied based on the 0th component of the array.
+  GKYL_DG_ARRAY_MASK_C0_LESS_THAN_FRAC_THRESHOLD, // Applies the mask based on the 0th component of the input array, which is spatially dependent.
+  GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_FRAC_THRESHOLD, // Applies the mask based on the 0th component of the input array, which is spatially dependent.
+  GKYL_DG_ARRAY_MASK_C0_LESS_THAN_FRAC_THRESHOLD_SPATIAL, // Applies the mask based on the 0th component of the input array, which is spatially dependent.
+  GKYL_DG_ARRAY_MASK_C0_GREATER_THAN_FRAC_THRESHOLD_SPATIAL, // Applies the mask based on the 0th component of the input array, which is spatially dependent.
```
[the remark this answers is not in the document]
Maxwell-Rosen: Renamed to 
```
enum gkyl_dg_array_mask_types {
  GKYL_DG_ARRAY_MASK_NONE = 0,             // No mask applied.
  GKYL_DG_ARRAY_MASK_C0_LESS,              // Mask applied based on the 0th component of the array.
  GKYL_DG_ARRAY_MASK_C0_LESS_FRAC,         // Applies the mask based on the global maximum 0th component of the input array.
  GKYL_DG_ARRAY_MASK_C0_LESS_FRAC_CONF,    // Applies the mask based on the local maximum 0th component of the input array, which is spatially dependent.
  GKYL_DG_ARRAY_MASK_C0_GREATER,           // Mask applied based on the 0th component of the array.
  GKYL_DG_ARRAY_MASK_C0_GREATER_FRAC,      // Applies the mask based on the global maximum 0th component of the input array.
  GKYL_DG_ARRAY_MASK_C0_GREATER_FRAC_CONF, // Applies the mask based on the local maximum 0th component of the input array, which is spatially dependent.
};
```

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,380 @@
...
+    // Store all ranges from input as pointers.
+    mask->phase_rng = mask_inp.phase_rng;
+    mask->phase_rng_ext = mask_inp.phase_rng_ext;
+    mask->conf_rng = mask_inp.conf_rng;
+    mask->conf_rng_ext = mask_inp.conf_rng_ext;
+    mask->vel_rng = mask_inp.vel_rng;
+    
+    // Determine mask range: use phase_rng for kinetic species, conf_rng for fluid species.
+    // A NULL or zero-volume phase_rng indicates a fluid species.
+    bool is_kinetic = (mask_inp.phase_rng && mask_inp.phase_rng->volume > 0);
```
manauref: this isn't the correct/best way to check if kinetic. Best to compute vdim from phase_rng.ndim-conf_rng.ndim, and if >0 then kinetic. Because an (fluid) app may pass a "phase range" to this that is actually a conf_range.

core/zero/gkyl_dg_array_mask.h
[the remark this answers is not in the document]
Maxwell-Rosen: It's partly because the on_dev pointer is inside the private header for the mask, so modules should not access the pointer directly but should use the getter function. I actually think other parts of the code should utilize this structure because we can apply the const modifier to the pointer to make sure that nothing is moving the pointers around. Albeit, this code does not put the const modifier there and the CUDA files already include the private header. We can't really make this a const in the CUDA code either because we assign the device pointer, then rewrite it onto the host pointer, which isn't exactly memory safe. This is, however, a higher standard of memory safety than the other modules, so I don't think this should be a discarded change because it's not the same.

core/zero/dg_array_mask.c
Maxwell-Rosen: Added support for only specifying a configuration space range.
```
    // Determine mask range: use phase_rng for kinetic species, conf_rng for fluid species.
    // A NULL or zero-volume phase_rng indicates a fluid species.
    bool is_kinetic = (mask_inp.phase_rng && mask_inp.phase_rng->volume > 0);
    if (is_kinetic) {
      mask->mask_rng = mask_inp.phase_rng;
      mask->mask_rng_ext = mask_inp.phase_rng_ext;
    } else {
      mask->mask_rng = mask_inp.conf_rng;
      mask->mask_rng_ext = mask_inp.conf_rng_ext;
    }
```

core/zero/gkyl_dg_array_mask_priv.h
```diff
@@ -0,0 +1,84 @@
...
+  struct gkyl_array *mask; // Mask array (1.0 is true, -1.0 is false).
+  double threshold; // Threshold for marking cells as masked. Scaled absolute value for *_THRESHOLD types, fraction for *_FRAC_THRESHOLD types.
+  struct gkyl_range phase_rng; // Phase-space range.
+  struct gkyl_range phase_rng_ext; // Extended phase-space range.
+  struct gkyl_range conf_rng; // Configuration-space range.
+  struct gkyl_range conf_rng_ext; // Extended configuration-space range.
+  struct gkyl_range vel_rng; // Velocity-space range.
```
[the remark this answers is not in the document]
Maxwell-Rosen: Done

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,317 @@
...
+          &mask->conf_rng, &mask->vel_rng, &mask->phase_rng, iter_conf.idx, frac_threshold);
+      }
+    }
+  }
+}
+
+void
+gkyl_dg_array_mask_scale_by_cell(struct gkyl_dg_array_mask *mask, const struct gkyl_array *arr_to_multiply)
+{
+  if (mask->type == GKYL_DG_ARRAY_MASK_NONE) {
```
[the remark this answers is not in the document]
Maxwell-Rosen: Done.. See earlier comment

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,317 @@
...
+      local_max = abs_val;
+    }
+  }
+  
+  return local_max;
+}
+
+// Static helper: Apply spatial fractional mask (less-than) for a given config cell
+static void
+apply_spatial_mask_less_than(struct gkyl_array *mask_arr, const struct gkyl_array *arr_to_mask,
```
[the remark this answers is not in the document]
Maxwell-Rosen: Changed

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,317 @@
...
+#include <gkyl_alloc.h>
+#include <gkyl_alloc_flags_priv.h>
+#include <gkyl_array_ops.h>
+#include <gkyl_array_reduce.h>
+#include <gkyl_dg_array_mask.h>
+#include <gkyl_dg_array_mask_priv.h>
+
+// Static helper: Apply less-than threshold mask over phase range
+static void
+apply_mask_less_than(struct gkyl_array *mask_arr, const struct gkyl_array *arr_to_mask,
```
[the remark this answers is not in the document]
Maxwell-Rosen: Changed to `arr_in`

core/zero/dg_array_mask.c
```diff
@@ -0,0 +1,398 @@
...
+const struct gkyl_array *
+gkyl_dg_array_mask_get_mask(const struct gkyl_dg_array_mask *mask)
+{
+  return mask->mask_arr;
+}
+
+void
+gkyl_dg_array_mask_release(struct gkyl_dg_array_mask *mask)
+{
+  if (mask) {
```
manauref: why this if-statement? i don't think other objects have it

Maxwell-Rosen: The mask array is not always allocated. There are so many enums that seperate free functions for each one would be cumbersome. This if statement ensures that the array is only freed when it is not allocated, as the pointer defaults to zero.
```
  if (mask->mask_arr) {
    gkyl_array_release(mask->mask_arr);
  }
```

Maxwell-Rosen:
> 1. It seems to me like there are lots of apps you haven't added and need to add the skip cell to. For example
> 
> ```
> gk_neut_species_bgk.c
> gk_neut_species_positivity.c
> gk_neut_species_react.c
> gk_neut_species_recycle_react_scale.c
> gk_neut_species_source.c
> gk_species_bgk.c
> gk_species_damping.c
> gk_species_fdot_multiplier.c
> gk_species_heating.c
> gk_species_radiation.c
> gk_species_react.c
> gk_species_source.c
> ```
> 
> 2. When I look at the deepest if-statement that actually does the masking, in various updaters, you always negate the mask. Why not change the meaning/default value of the mask then, so that instead of negating everywhere you just evaluate it?
> 3. I think you should adapt the mask in core/zero so it works with conf-space quantities, i.e. to evolve fluid moments and/or to mask moments.
> 4. Have you run all the unit tests? seems to me like you changed lots of updaters but only changed 1 unit test.
> 5. You changed many, arguably most of the core updaters. I think when all changes are made you need to do some serious regression test running and comparison with main. I suggest running and comparing:
> 
> ```
> rt_gk_ar_react_nonuniformv_1x2v_p1.c
> rt_gk_bgk_asdex_2x2v_p1.c
> rt_gk_bgk_im_asdex_2x2v_p1.c
> rt_gk_ion_sound_adiabatic_elc_1x2v_p1.c
> rt_gk_lapd_cyl_3x2v_p1_nonuniformr.c
> rt_gk_lbo_cross_relax_1x2v_p1.c
> rt_gk_lbo_relax_1x2v_p1.c
> rt_gk_lbo_relax_varnu_1x2v_p1.c
> rt_gk_li_react_nonuniformv_3x2v_p1.c
> rt_gk_ltx_boltz_elc_1x2v_p1.c
> rt_gk_mirror_boltz_elc_damped_1x2v_p1.c
> rt_gk_mirror_boltz_elc_poa_1x2v_p1.c
> rt_gk_multib_asdex_2x2v_p1.c
> rt_gk_multib_step_2x2v_p1.c
> rt_gk_neut_recycle_1x3v_p1.c
> rt_gk_neut_sheath_3x2v_p1.c
> rt_gk_rad_low_Te_1x2v_p1.c
> rt_gk_sheath_1x2v_p1.c
> rt_gk_sheath_2x2v_p1.c
> rt_gk_sheath_3x2v_p1.c
> rt_gk_sheath_cx_2x2v_p1.c
> rt_gk_sheath_flr_2x2v_p1.c
> rt_gk_sheath_fluid_neut_1x2v_p1.c
> rt_gk_sheath_heat_source_2x2v_p1.c
> rt_gk_static_3x2v_p1.c
> rt_gk_step_out_2x2v_p1.c
> rt_gk_tcv_iwl_adapt_source_2x2v_p1.c
> rt_gk_tcv_iwl_adapt_source_3x2v_p1.c
> rt_gk_wham_1x2v_p1.c
> rt_gk_wham_1xIC_2x2v_p1.c
> rt_gk_wham_2xIC_3x2v_p1.c
> ```

1. I do not intend this PR to be an expansion of where the skip_cell technique is used. In the simulations I'm working with, this is a reasonable scope of work and these apps do not need the skip_cell object. If the updaters work with the current infrastructure, then they are accepting the appropriate and correct objects, which means the array mask is passed in the right places. The intent of this PR is to generalize the skip_cell object into the dg_array_mask, which is the real contribution of this PR. I think extending the scope of the skip_cell object is outside the scope of this PR. For instance, we modified the positivity updater, but since the neutral positivity app is working, there should be no issues. Neutrals are controlled by the vlasov app, which does not have the same skip_cell infrastructure as GK, so modifications of the neutrals code is irrelevant.
2. addressed.
3. addressed.
4. yes.
5. The tests you laid out all incorporate the same null mask. Only one null mask case is necessary, which is why I have been comparing sheath regression tests and the only regression test that uses the skip_cell feature. Regression comparison with the main over 20 tests is incredibly tedious, and the testing I have done sufficiently shows no errors with the main. This is very very difficult to do every time you suggest a refactor of a function, then I have to spend hours on regression tests over again. I'm reluctant to do comprehensive testing because of the lack of automation in our code, which should be addressed and is critical to our research. Thus, comparisons of gk_sheath and gk_wham are enough to show that the refactor has not changed the output of the code in a meaningful way without spending too much valuable human time on testing.
image
image

PR #892, opened 2025-11-07
Fix boundary  flux bug in MB

akashukla:
Fix a bug that was introduced in MB along with the boundary flux changes. I think there were just some typos in the multib  rk3 update. Pattern matching with the SB rk3 update has resolved the issue and now the diagnostics reflect that MB conserves particles. To be clear, this was just an issue  with the boundary flux calls, so it didn't actually break conservation and simulation results (in the absence of adaptive sources) were unaffected by the bug. This fix is important for the correct behavior of adaptive sourcing in MB and also for being able to check conservation. 

Here is particle conservation relative error from the current main branch on 11/7 for rt_gk_multib_slab_2x2v_p1.c:
image

Here it is from this branch:
image

Here it is from before (commit f4ee715b720920fe9c94db6788c93f4eb1a5554a) the fluid neutrals PR where the bug was introduced
image


Antoinehoff:
This may help also the source to adjust well right? did it change the behavior of your run with adaptive source?

akashukla:
> This may help also the source to adjust well right? did it change the behavior of your run with adaptive source?

I have been running on a pre-fluid_neut branch, so this bug was not present in my simulations. However overall yes I would like to run on main  with this fix, because the new placement of the boundary flux calls should help my  adaptive sourcing simulations behave even better. By the way, the  adaptive sourcing is going pretty well and I got a simulation to work, I'll update you more on slack.

Antoinehoff (approved):

PR #900, opened 2025-11-12
Update version of OpenMPI built so that eqdskreader tool works

jjz73:
When running Gkeyll with the version of OpenMPI that is currently built by default (v4.1.6), Gkeyll's native eqdskreader tool throws an MPI-related error message. I have tested this when building with v5.0.7 though, and the tool cooperates and generates the .gkyl files containing $\psi$, a necessity for informing our simulation grids.

### DIFF
--- a/install-deps/build-openmpi.sh
+++ b/install-deps/build-openmpi.sh
@@ -3,7 +3,7 @@
 source ./build-opts.sh
 
 # Install prefix
-PREFIX=$GKYLSOFT/openmpi-4.1.6
+PREFIX=$GKYLSOFT/openmpi-5.0.7
 # Location where dependency sources will be downloaded
 DEP_SOURCES=$GKYLSOFT/dep_src/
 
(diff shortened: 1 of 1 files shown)


jjz73:
It says that all checks on here have passed, but I checked the log and it doesn't look like it runs the mpi_comm tests in that sequence, it just builds them. I tried running them on my local Gkeyll build and they don't seem to pass yet, but I don't know if they did even when I was running v4.1.6 of MPI, so I'm not sure if it's an issue with how I'm building the code in general.

manauref:
you have to run them yourself, try doing this in main
```
mpirun -np 1 build/core/mctest_mpi_comm -M -E
mpirun -np 2 build/core/mctest_mpi_comm -M -E
mpirun -np 4 build/core/mctest_mpi_comm -M -E
mpirun -np 8 build/core/mctest_mpi_comm -M -E
```
(with different numbers of processes because there are multiple tests, some use specific number of processes) and then do it again on this branch and make sure the behavior is the same

jjz73:
On both branches, the test mctest_mpi_comm passes with 1, 4, and 8 processes, but fails with 2 processes. The behavior is identical on both branches.

manauref:
can you post failure message? or whatever it says please?

Issue #897, opened 2025-11-11
Nonuniform grids do not work with mirror geometry


Maxwell-Rosen:
Since the geometry refactor, the non-uniform grids have been depricated inside the mirror geometry modules. This started when we re-wrote the mirror geometry to use axial Z. The mirrorgridgen functions are now in core, but they need the position map for non-uniform grids. We need to move the mirrorgridgen tools to gyrokinetic/, but that will break the lua tools. There is something funky with the layering. Perhaps we move position map to core, but I don't like that idea. It's written non-GK specific intentionally though.

PR #901, opened 2025-11-13
Fix the non-uniform grids for numeric mirror geometry

Maxwell-Rosen:
# Bug fix

## Summary

Description: 
Introduced in the re-write of mirror geometry in terms of RZ nodes and sqrt(psi), we did not fully re-implement the non-uniform grids. Here, I'm fixing that. It does work, as shown in the testing below.

Issue link: #897

## Solution

We had to move the mirror_grid_gen.c into gyrokinetic from core, which breaks the Lua tool. We need this functionality in the simulations to make real research progress. I do not want to get bogged down by one lua tool not working. The lua functionality needs to be extended for this to work. The alternative is moving position_map.c into core, which sounds worse.

Impacted files:gk_geometry_mirror.c, mirror_grid_gen.c

Automated testing: Clean up the mirror regression tests. Update the non-uniform regression tests. I'm only using the numerical map, so that's the only one I want in the regression tests. The polynomial map is not as good.

## Community Standards

- [x]  Documentation has been updated.

- [ ] I added a unit test for this feature.

## Failing unit tests
../build/moments/unit/ctest_gr_spacetime
../build/moments/unit/ctest_wv_gr_mhd
../build/moments/unit/ctest_wv_gr_mhd_tetrad
../build/moments/unit/ctest_wv_gr_ultra_rel_euler
../build/moments/unit/ctest_wv_gr_ultra_rel_euler_tetrad
ctest_rescale_ghost_jacf.c (This one crashed my computer)
Several tests segfault too

## Additional Notes

Magnetic field when s=1 is the only parameter specified. Perfectly optimal magnetic field hat function. This one is so cool!
pgkyl gk_wham_nonuniformx_1x2v_p1_numeric-bmag.gkyl interp pl
image

However when using this map, jacobgeo_inv goes crazy.
pgkyl gk_wham_nonuniformx_1x2v_p1_numeric-jacobgeo_inv.gkyl interp pl
image

Here is mc2nu_pos_deflated, which is the non-uniform map itsself. It's so extreme, which is why the geometry goes crazy.
pgkyl gk_wham_nonuniformx_1x2v_p1_numeric-mc2nu_pos_deflated.gkyl interp pl
image

However, when we use the gaussian weighted moving average filter, the coefficients become much smoother. However, we do loose some hattyness of B.
image
image
image


gkeyll/Makefile-gkeyll
```diff
@@ -18,7 +18,7 @@ DEPS := $(OBJS:.o=.d)
 EXT_INCS := -I${MPI_INC_DIR} -I${LUA_INC_DIR} -I${LAPACK_INC} -I${SUPERLU_INC}
 APP_LW_INCS := -I../core/apps -I../moments/apps -I../vlasov/apps -I../pkpm/apps -I../gyrokinetic/apps
 
-CORE_INCS := -I../core/minus -I../core/zero
+CORE_INCS := -I../core/minus -I../core/zero -I../gyrokinetic/zero
```
JunoRavin: This is not right. You should not be include the gyrokinetic zero directory here. There are the core includes. The gyrokinetic includes are handled elsewhere. If you found this is necessary, we need to take another look at the Makefile-gkeyll to correctly handle the gyrokinetic includes.

Maxwell-Rosen:
I moved the mirror geometry to gyrokinetic because it needs to depend on position_map, which is in gyrokinetic. Either we elevate position_map objects or move this file layer. I am fine either way. @ammarhakim suggested that it's okay to move the mirror grids lower to gyrokinetic for the lua tools, but that did break them.

Maxwell-Rosen:
Here are a few plots from that test. I didn't touch any parts of the toakamk code, so it's all identical. These plots are showing the absolute difference between the regression tests on main and in this branch.

image

image

image

manauref (approved):

PR #902, opened 2025-11-13
Fix rescaling in positivity_shift to conserve cell-wise M0 contribution & add positivity option to init_from_file

manauref:
In PR #737 we meant to, after shifting f to enforce positivity at nodes, rescale f in the current cell by the ratio of the old $f^{(0)}$ (the 0th DG coefficient) to the new $f^{(0)}$, see 3rd hand-drawn sketch in https://github.com/ammarhakim/gkeyll/pull/737 .

But instead we did this re-scaling using weak multiplication and division by the x-varying contribution from this cell to M0. That can cause issues because such operation is like using weak multiplication/division with Gauss-Legendre quadrature, but the old f was 0 at one or more Gauss-Legendre nodes. Indeed when @dingyunl tried to re-run the 2D ASDEX simulation with positivity in the new version of the code, the sim crashed.

Here we fix this, by rescaling by the ratio of $f^{(0)}$'s as PR #737 intended. With this change the 2D ASDEX sim proceeded without a problem.

Also, we add an `enforce_positivity` option to `init_from_file`, so that when importing ICs the user can force negative f (at nodes) to 0. And we also added the ability to pass the `jacobvel` file to `init_from_file` so that we rescale by the ratio of the velocity-space jacobians when refining velocity space (with nonuniform velocities). To facilitate this we added a `gkyl_array_divide_by_cell` array ops method, analogous to `scale_by_cell`, and its unit tests.

### DIFF
--- a/gyrokinetic/zero/positivity_shift_gyrokinetic.c
+++ b/gyrokinetic/zero/positivity_shift_gyrokinetic.c
@@ -134,17 +134,16 @@ gkyl_positivity_shift_gyrokinetic_advance(gkyl_positivity_shift_gyrokinetic* up,
           m0phase_out_c[k] = 0.0;
         up->kernels->m0(up->grid.dx, vmap_c, up->mass, bmag_c, distf_c, m0phase_out_c);
 
-        if (up->kernels->is_m0_positive(m0phase_in_c)) {
-          // Rescale f in this cell so it keeps the same density.
-          double m0ratio_c[num_cbasis];
-          up->kernels->conf_inv_op(m0phase_out_c, m0ratio_c);
-          up->kernels->conf_mul_op(m0phase_in_c, m0ratio_c, m0ratio_c);
+        if (m0phase_in_c[0] > 0.0 && m0phase_out_c[0] > 0.0) {
+          // Rescale f in this cell so it keeps the same cell-averaged density.
+          double m0ratio = m0phase_in_c[0]/m0phase_out_c[0];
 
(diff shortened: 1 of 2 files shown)


manauref:
some more detail about the crashing ASDEX sim for the record:
by evolving the solution until the step before it crashed, and restarting, we determined that the issue was precisely this rescaling in the positivity shift. Interestingly, in these restart tests it seemed that the CPU rescaling was fine, but the GPU rescaling was not. Printing the values I realized that the ratio of $n^{old}(x)/n^{new}(x)$ (where these are the density contributions from a single cell in velocity space) used polynomials where:
1. both the numerator and denominator were very close to zero, and this division behaved differently on CPUs and GPUs (in this particular occasion it resulted in a very large number on the GPU), or
2. although the cell average of $n^{old}(x)$ was positivity, and hence that's why this rescaling was taking place, $n^{old}(x)$ had a value very close to 0 or in fact negative at Gauss-Legendre nodes, so that when we did the weak rescaling $n^{old}(x)/n^{new}(x)$ then we were in fact multiplying by a (possibly large because $n^{new}(x)$ might've been small) negative number.

This is all avoided by the fix above, which was the original intent of PR #737

Maxwell-Rosen:
This seems to be a "positive" change, as it reduced heating in the mirror expander for my problems. I highly recommend a complete rebuild when testing this branch

image

JunoRavin (approved):
Very, very cool stuff. Ready to merge.

PR #911, opened 2025-11-25
Fix wall plates for IWL geometry in GK solver

manauref:
Wall plates were not working for IWL geometry. This fixes that, so we can do realistic "plates" in the SOL and produce a grid like this:
Screenshot 2025-11-25 at 12 26 08 PM

We also refactor the creation of core and SOL ranges so that it works with IWL numerical magnetic equilibrium, both for cases that psi increases towards the core or towards the outside.


Antoinehoff (approved):

manauref:
For the record, I checked that TCV and D3D regression tests are completely unchanged.

PR #912, opened 2025-11-26
Set the moment correction option when using a Maxwellian Gaussian projection

Antoinehoff:
This was causing a seg fault when using the max-gauss proj for species IC on GPU as the programm flow is slightly different from the source program flow. Basically it was failing at
```
zero/gk_maxwellian_correct.c:97
```
because the correct mom updater was not initialized. Now the max-gauss projection is setting the flag correctly and will avoid trying to correct the moment. I am not clear how this resolves also the hack done in PR #818, not sure it does.


manauref (approved):
Can you take this opportunity to change the 1/3 in L334 to 1/vdim_phys where `vdim_phys = vdim == 1? 1 : 3` please? and similarly for the other factors of 3 in the adaptive source file.

Issue #896, opened 2025-11-10
Lack of robustness in default settings architecture.


SjoerdPR:
In Gkeyll, the standard design of input settings appears to be to initialise structs to zero. Any zeros that remain after the user has overwritten some of these values in their input file are then interpreted as being unset, and then replaced under the hood by their default values.

An issue arrises with this architecture when these values are numeric, and a value of 0 would not be an unreasonable value to set. An example of this would be a user performing a collisionality scan, where the `nu_frac` value is set to various values between 0.0 and 1.0. At the moment, the end-user would have no way to know that the value they deliberately set to 0.0 would have been replaced by 1.0 under the hood.

The end-user that wants to run simulations should not be expected to be aware of this method to apply default values, so I think this lack of robustness is problematic and deserves to be addressed.

Some possible solutions to this include:

- Clear documentation indicating the variables where a numerical value of 0 would give rise to undefined behaviour.
- A warning or log that indicates whenever specified values are replaced by default values (probably undesirable due to log spam)
- Include switches that must be set to `true` to be able to manually alter values that give rise to such undefined behaviour (i.e. .scale_collisionality = true for nu_frac). If this switch is set, never replace by the default value. If the switch is unset but the value is set to nonzero anyway, throw an error.
- replacing `double` by a `double`-wrapper type or a `double` pointer for such variables, where it becomes possible to distinguish between a null pointer (unset) or a numerical value of 0.
- (the most robust but also the most effort to implement): redesign of the architecture to globally remove the equivalence between unset parameters and numerical 0.0.

manauref:
@SjoerdPR I wouldn't refer to this as an issue with robustness (i.e. the capacity if simulation proceeds to a solution), but rather a conflict in managing default options and values.

I think some specificity is needed in this issue. To this end you offer the case of scanning `nu_frac`, so let's use that as an example. How are you scanning or proposing to scan `nu_frac`? i.e. can you provide sample code, input files and/or job scripts?

SjoerdPR:
@manauref thanks for the quick response, and apologies for using incorrect terminology.

Suppose someone wanted to see how parallel heat conductivity (with the collision operator implementation in Gkeyll) depended on collisionality. So they take their favourite simulation and run it 11 times ranging from collisionless electrons to regular collisionality with the following setup.
```
.collisions =  {
      .collision_id = GKYL_LBO_COLLISIONS,
[...]

PR #919, opened 2025-12-08
Added note that scale_factor must be nonzero

SjoerdPR:
For multirate timestepping, one may generally want to effectively disable the collisionless terms during certain phases. Currently, in the regression tests (such as `rt_gk_mirror_boltz_elc_poa_1x2v_p1.c`), this is done by setting the scale_factor value to a small value (0.01).

```
struct gkyl_gyrokinetic_collisionless collisionless_inp = {
    .scale_factor = pparams->alpha,
};
```

It seems almost inevitable that someone trying to replicate this unit test is at some point going to want to disable certain collisionless terms entirely during certain phases to speed these phases up. Using the unit tests as reference, they'd set alpha to 0. At the moment, doing so would result in `scale_factor` stealthily being set to 1.0 under the hood (see #896).

As a band-aid fix, we can add comments to at least specify such behaviour as undefined behaviour, although a more structured solution might be desired in the future.

### DIFF
--- a/gyrokinetic/apps/gkyl_gyrokinetic.h
+++ b/gyrokinetic/apps/gkyl_gyrokinetic.h
@@ -77,7 +77,7 @@ struct gkyl_phase_diagnostics_inp {
 struct gkyl_gyrokinetic_collisionless {
   enum gkyl_gk_collisionless_type type; // Type of collisionless terms.
   bool write_diagnostics; // Whether to output diagnostics.
-  double scale_factor; // Factor multiplying collisionless terms. NOTE: Trying to disable collisionless terms entirely by setting to 0 will NOT work; set to negligible nonzero value or set type to GKYL_GK_COLLISIONLESS_NONE instead.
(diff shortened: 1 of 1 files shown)


Maxwell-Rosen:
Yes, I do find this confusing. Since we have the expression
$ \partial_t f \propto \alpha \{H, f\} $
when the user says alpha = 0, we might expect it to disable the collisionless updater, but it instead sets it to 1.0, since that is the unmodified equation.

SjoerdPR:
> @SjoerdPR the correct way to disable collisionless terms entirely would be to set `.type = GKYL_GK_COLLISIONLESS_NONE`

@manauref Unless I am misunderstanding the flow, I don't think it's possible at the moment to change to `GKYL_GK_COLLISIONLESS_NONE` during a phase since `gk_species_collisionless.gk_species_collisionless_reset()` currently does not touch `gkcls.collisionless_id`. I'm a bit hesitant to create a pull request that changes that function because I'm not entirely clear on what else would need to change for the reset to occur properly.

PR #927, opened 2025-12-29
Fix failing unit tests related to the position map

Maxwell-Rosen:
# Bug fix

## Summary

Description: 

Mana informed me that the twist-shift BC unit test was failing and it seemed to be caused by the position map.
image
image

I found this quite strange because when I ran these unit tests on perlmutter GPU, they were just fine on main.
image

In any case, I did see a potential issue with memory handling regarding how the gk_geometry modules create a null position map if none is provided. This memory handling could lead to different behavior on different machines (although here, both Mana and I were testing on Perlmutter GPU). Even though I can't reproduce the same failure, I cleaned up the code by intentionally allocating a null position map in the unit tests and passing that to the geometry generators.

## Solution

Create a null position map in each unit tests.
Remove code in the gk_geometry generators to detect if no position map is passed.
Removed unused "flags" variable from the position map.

Automated testing: unit tests are updated. position_map tests are updated accordingly.

## Community Standards

## Testing: 

- [ ] I added a unit test for this feature.

## Additional Notes

I would like @manauref to run the appropriate tests before this is merged. I am unable to reproduce the original error, so I cannot verify that these modifications fix the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,16 +19,17 @@ CFLAGS ?= -O3 -g -ffast-math -fPIC -MMD -MP -DGIT_COMMIT_ID=\"$(GIT_TIP)\" -DGKY
 LDFLAGS = 
 PREFIX ?= ${HOME}/gkylsoft
 
-# Default lapack include and libraries: we prefer linking to static library
-LAPACK_INC = $(PREFIX)/OpenBLAS/include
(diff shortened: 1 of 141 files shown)


Maxwell-Rosen:
Merged into https://github.com/ammarhakim/gkeyll/pull/941, which addresses CI and failing unit tests. This branch is not valgrind clean, but I valgrind-sanitized it in that PR.

Issue #898, opened 2025-11-11
[DR] Biasing in perpendicular Poisson solver for gyrokinetics


manauref:
In IWL simulations with open and closed field lines we wish to bias the limiter corner. Previously we were doing that in `fem_poisson` and `deflated_fem_poisson`, but these solvers break energy conservation inside the domain.

Here we propose to bias the same location within the `fem_poisson_perp`, which has been shown to be necessary to conserve energy in other simulations. Unlike the previous implementation which specified biased planes, we will now specify biased lines, e.g. in the input file we'll provide:
```
  struct gkyl_poisson_bias_line target_corner_bcs[] = {
    {
     .perp_dirs = {0, 1}, // Directions perpendicular to line.
     .perp_coords = {ctx.x_LCFS, ctx.z_min}, // Coordinates of the line in perpendicular directions.
     .val = 0.0, // Biasing value.
    },
    {
     .perp_dirs = {0, 1}, // Directions perpendicular to line.
     .perp_coords = {ctx.x_LCFS, ctx.z_max}, // Coordinates of the line in perpendicular directions.
     .val = 0.0, // Biasing value.
    },
  };

  struct gkyl_poisson_bias_line_list bias_line_list = {
    .num_bias_line = 2,
    .bl = target_corner_bc,
  };

  struct gkyl_gyrokinetic_field field = {
    ...
    .bias_line_list = &bias_line_list,
  };
```

And the biasing under the hood is done as it was done for fem_poisson: we use additional kernels and loops to set entries in the LHS matrix to 0 or 1, corresponding to the nodes on this line, and the corresponding entries in the RHS vector to the biasing value.

This is being prototyped in the `poisson_perp_bias` branch.

Antoinehoff:
Great !

PR #928, opened 2026-01-07
Swap plane Poisson solver with slab Poisson solver in GK IWL simulations & 2x LBO energy conservation

manauref:
This goes with [PR 93 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/93)

## Problem

Currently the IWL simulations are using the Poisson solver on planes (`fem_poisson_deflated`). This has been shown to break energy conservation. One reason we have continued to do that is that:
a) We didn't have the ability to bias the limiter corner in the slab Poisson solver (`fem_poisson_perp`).
b) We didn't have a working workflow that used the slab Poisson solver and twist-shift BCs (TS BCs).

Here we address both of these points, and deprecate the use of `fem_poisson_deflated` in favor of `fem_poisson_perp`.

### Biasing

We implemented the ability to bias the limiter corner in `fem_poisson_perp`. Some info is in the DR #898 .

Unlike biasing in `fem_poisson`, where can bias whole planes in the solver, in `fem_poisson_perp` we bias lines. That's at the moment we just need to bias the lines at (x,z)=(x_LCFS,z_min) and (x,z)=(x_LCFS,z_max).

Tests with biasing were added to the fem_poisson_perp unit test.

### Workflow with TS BCs

We implemented a new set of steps to make sure the potential is twist-shift periodic with this new slab solver. These are:

1. Smooth the charge density, without BCs (`fem_parproj`).
2. Solve the perpendicular Poisson problem (`fem_poisson_perp`).
3. In the core, apply TS BC to phi at the lower z-boundary, and a ghost-from-skin-surf BC at the upper z-boundary.
4. Smooth phi, with Dirichlet BCs that take the Dirichlet value from the ghost in the core, and witih Dirichlet BCs that take the Dirichlet value from the skin in the SOL.

#### Commentary on alternatives

- Regarding step 1: It's unclear whether BCs are needed here, or what they should be. In SOL simulations we don't apply BCs, and that's the option that makes the operator self-adjoint and preserves energy. Perhaps indirectly using TS BCs as done in steps 3-4 would be worthwhile, but doesn't seem necessary so far.
- Note that steps 3-4 only apply TS BC at the lower boundary. This is as it's done in main now. We tried, both in main and in this branch, but applying TS BC only at the upper boundary or in combination with the lower, does not work.

## Additional perks

We noticed that the LBO collision operator was not conserving energy in 2x. We fixed that in this branch.

This fix is not the final one, though. Through more testing I found scenarios (large gradients + normNu) in which the LBO still doesn't conserve energy. I think we need to compute nu-weighted moments to truly conserve energy, which we may pursue in another branch.

## Tests

This branch is valgrind and compute-sanitizer clean.

### Regression tests

Regression tests run, but they are expected to yield slightly different solutions since we are using a different algorithm now.

#### rt_gk_d3d_iwl_2x2v_p1

Screenshot 2026-01-07 at 10 20 06 AM

Screenshot 2026-01-07 at 10 18 13 AM

Screenshot 2026-01-07 at 10 20 59 AM

#### rt_gk_tcv_iwl_adapt_source_2x2v_p1

Screenshot 2026-01-07 at 10 23 17 AM

Screenshot 2026-01-07 at 10 24 20 AM

Screenshot 2026-01-07 at 10 25 08 AM

#### rt_gk_d3d_iwl_3x2v_p1

Screenshot 2026-01-07 at 10 26 20 AM

Screenshot 2026-01-07 at 10 28 07 AM

Screenshot 2026-01-07 at 10 29 35 AM

#### rt_gk_tcv_iwl_adapt_source_3x2v_p1

Screenshot 2026-01-07 at 10 30 38 AM

Screenshot 2026-01-07 at 10 32 36 AM

Screenshot 2026-01-07 at 10 31 29 AM

### Production simulation

We ran a TCV case provided by @Antoinehoff in both main and this branch. Here are some snapshots.

Screenshot 2026-01-07 at 9 35 48 PM

Screenshot 2026-01-07 at 9 35 08 PM

Screenshot 2026-01-07 at 9 34 09 PM

Screenshot 2026-01-07 at 9 33 33 PM

Screenshot 2026-01-07 at 9 33 00 PM

Screenshot 2026-01-07 at 9 31 53 PM

Screenshot 2026-01-07 at 9 31 35 PM


Antoinehoff:
**⚠️ The branch has changed since then, the results below might be obsolete ⚠️**

Here are one-to-one comparison of production like simulations between the current state of **main** (5506bbe9d39a) and this branch (8fee1f5bfb5d). The simulation setup is GK TCV iwl 3x2v, see [rt_gk_tcv_iwl_adapt_source_3x2v_p1.c](https://github.com/ammarhakim/gkeyll/blob/main/gyrokinetic/creg/rt_gk_tcv_iwl_adapt_source_3x2v_p1.c), for a coarse resolution (24x16x12x12x6) as presented in [Hoffmann et al. 2026](https://arxiv.org/abs/2510.11874). The only parameter alterations are an increase of the input power, 0.5MW instead of ~0.25MW, to accelerate turbulence development, and a reduction of the collisionality scaling factor, $\nu_{frac}=0.5$ instead of $1.0$, to increase the time step.
The results from the main and this branch are labelled **main** and **poissonperp**, respectively.

Edit: I updated the plots and analysis with longer runs as a noticeable difference rise when approaching the quasi-steady state.

Note: The poisson perp run crashed at t~1320mus, the main run did not reach that time yet so it is unclear if the crash is solely due to the solver. I've observed that reducing nu_frac can destabilize the simulation at longer time.

### General numerics comments
The poissonperp run presents a very stable time step ~4.6ns against a fluctuating one for the main run, ~3 to 4.6ns. The poissonperp simulation is consequently faster, achieving ~400mus against 330mus for main, in 6h and 4 GPUs.

### Energy
It seems that the poissonperp leads to a lower confined energy state. 
We see a strong reduction of the ion integrated Hamiltonian because of the global reduction of the potential value. This does not affect the total Hamiltonian. However, the electron Hamiltonian is lower in poissonperp than in main despite the stronger negativity of the potential, which indicates that the thermal energy of the electrons is strongly reduced in the poissonperp.
**main**
image
**poissonperp**
image

### Potential
Poisson perp saturates into a lower potential values everywhere, increasing the ExB shear at the inner radial boundary and decreasing tit at the outer one.
**main**
image
**poissonperp**
image

### Density
Fairly similar, no comments.
**main**
image
**poissonperp**
image

### Temperature
The electron quasi-steady state temperature is remarkably reduced with the Poisson perp solver as it is expected from the energy analysis above. The SOL electron temperature matches better with experiment but it is unclear for the core (recall that the initial conditions of this simulation is ~ the experimental profile measurement). For the ions, a stronger temperature "well" is observed in the Poisson perp simulation.
**main**
image
**poissonperp**
image

### Density fluctuations
We see that the fluctuations at the inner radial boundary are reduced in the poisson perp case, which may be due to the increase ExB shear in that region (potential well is larger in poisson perp).
**main**
ne_fluct_main__n_frames_400_to_499
**poissonperp**
ne_fluct_poiperp__n_frames_400_to_499

### Temperature fluctuations
It seems that the y-flows are stronger in the main simulation.
**main**
TeTi_fluct_main__Te_frames_400_to_499
**poissonperp**
TeTi_fluct_poiperp__Te_frames_400_to_499

Antoinehoff (approved):
Thanks for this great work!!

Antoinehoff (changes requested):
Longer runs show significant discrepancies in the quasi steady state, who's right? The above analysis is updated.

Antoinehoff:
> @Antoinehoff one concern you had is that your energy conservation time trace looked worse in this branch, right? But the adaptive sources aim to conserve M2 energy, not H energy, right?

The adaptive sources are connected to H energy, which is equal to M2 energy in many places since we have phi(x_lo)=0 and phi(x_up)=0.

Antoinehoff:
@manauref can you also the `ctest_fem_poisson_perp.c` so that it tests the Helmholtz solver fully? In particular, it should test that we can use `gkyl_fem_poisson_perp_update_lhs` to change `kSq`. I see that it only tests changing `epsilon` right now. See https://github.com/ammarhakim/gkeyll/issues/952 for example of testing.

ammarhakim (approved):
Ready to merge

Antoinehoff:
So the plots of the PR are from January 7 but many things have been changed in this branch since then. Do you know if the conclusions still hold? Can you please update the plots?
Also, can you show a plot that support the core issue you want to solve here which is: "[The poisson planar solve] has been shown to break energy conservation."? I think it is important to show that this PR resolves this issue.

@ammarhakim I don't think it is ready in the sense that we need to perform production-like clopen simulations to establish a baseline as this branch is removing entirely the planar Poisson solver from main, which was used in all clopen simulations until now. The smoothing was still undecided last week so we had to wait for this robustness.

@tnbernard would you agree to run a cheap DIII-D simulation with this branch and compare with existing data? I will do that for TCV as well.

Antoinehoff (changes requested):
We cannot merge this branch yet. The current version of this branch does not conserve particles where main was.

## Test
We use the regression test `gyrokinetic/creg/rt_gk_tcv_iwl_adapt_source_2x2v_p1.c` and postgkyl to investigate the error in particle conservation using this command line:
```
pgkyl gk-particle-balance --name rt_gk_tcv_iwl_adapt_source_2x2v_p1 --species elc --absy --logy -r --saveas  rt_gk_tcv_iwl_adapt_source_2x2v_p1_particle_balance_elc.png
```
The code is compiled on Perlmutter and run on 1 CPU, modules:
```
Currently Loaded Modules:
  1) conda/Miniforge3-25.11.0-1       4) craype/2.7.35    (c)   7) craype-network-ofi  (cpe)     10) cudatoolkit/13.0      (g)    13) nccl/2.29.2-cu13 (g)
  2) python/3.13-26.1.0               5) cray-dsmml/0.3.1       8) cray-libsci/25.09.0           11) craype-accel-nvidia80 (cpe)
  3) gcc-native/14              (c)   6) libfabric/1.22.0       9) PrgEnv-gnu/8.6.0    (prgenv)  12) cray-mpich/9.0.1      (mpi)

  Where:
   g:       built for GPU
   mpi:     MPI Providers
   cpe:     Cray Programming Environment Modules
   prgenv:  Programming Environment Modules
   c:       Compiler

```

### Result with `poisson_perp_bias`
To verify that the code environment is correct we use the `pgkyl info` command on the first ion dist. func. file:
```
(.venv) (nersc-python) ah1032@login13:~/gkeyll_poisson_perp/gkeyll/rt_gk_tcv_2x2v.o> pgkyl rt_gk_tcv_iwl_adapt_source_2x2v_p1-ion_0.gkyl info
(default#0)
├─ Time: 0.000000e+00
├─ Frame: 0
├─ Number of components: 24
├─ Number of dimensions: 4
├─ Grid: (uniform)
│  ├─ Dim 0: Num. cells: 9; Lower: 0.000000e+00; Upper: 1.200000e-01
│  ├─ Dim 1: Num. cells: 8; Lower: -3.141593e+00; Upper: 3.141593e+00
│  ├─ Dim 2: Num. cells: 8; Lower: -7.071068e-01; Upper: 7.071068e-01
│  └─ Dim 3: Num. cells: 8; Lower: 0.000000e+00; Upper: 1.000000e+00
├─ Maximum: 8.721618e-07 at (0, 6, 3, 2) component 0
├─ Minimum: -1.493391e-07 at (2, 1, 3, 2) component 1
├─ DG info:
│  ├─ Polynomial Order: 1
│  └─ Basis Type: gkhybrid (modal)
├─ Created with Gkeyll:
│  ├─ Changeset: b33037e7ef2e+
│  └─ Build Date: 2026-04-29 11:01:25 PDT
├─ Geometry info:
```
The commit is [b33037e](https://github.com/ammarhakim/gkeyll/pull/928/commits/b33037e7ef2e921a3a608709f52d7fe978ecb313)

Here is the particle balance:
Image

### Result with `main`
To verify that the code environment is correct we use the info command
```
(.venv) (nersc-python) ah1032@login13:~/gkeyll_main/gkeyll/rt_gk_tcv_iwl_adapt_source_2x2v_p1.o> pgkyl rt_gk_tcv_iwl_adapt_source_2x2v_p1-ion_0.gkyl info
(default#0)
├─ Time: 0.000000e+00
├─ Frame: 0
├─ Number of components: 24
├─ Number of dimensions: 4
├─ Grid: (uniform)
│  ├─ Dim 0: Num. cells: 9; Lower: 0.000000e+00; Upper: 1.200000e-01
│  ├─ Dim 1: Num. cells: 8; Lower: -3.141593e+00; Upper: 3.141593e+00
│  ├─ Dim 2: Num. cells: 8; Lower: -7.071068e-01; Upper: 7.071068e-01
│  └─ Dim 3: Num. cells: 8; Lower: 0.000000e+00; Upper: 1.000000e+00
├─ Maximum: 8.721618e-07 at (0, 6, 3, 2) component 0
├─ Minimum: -1.493391e-07 at (2, 1, 3, 2) component 1
├─ DG info:
│  ├─ Polynomial Order: 1
│  └─ Basis Type: gkhybrid (modal)
├─ Created with Gkeyll:
│  ├─ Changeset: 54375fde0853+
│  └─ Build Date: 2026-04-26 08:55:30 PDT
├─ Geometry info:
```
The commit is indeed the latest version of main [54375fde0](https://github.com/ammarhakim/gkeyll/commit/54375fde08532bb0c79a6c2636bc8ebcb3ec8bc8).
and the particles are conserved
Image

manauref:
There was a wrong assumption in the way periodicity of the smoother was enforced for 2x. It's been fixed now. See 
https://github.com/ammarhakim/gkeyll/commit/551bb72d49901189ad667cecb51c05a197ae6e31

Particle conservation in 2x IWL is machine precision now. Energy conservation is 10x better than in main (say in rt_tcv_iwl_adapt_source_3x2v_p1).

manauref:
TCV reg test conservation:
- Particles: machine precision for in 2x, 1e-12 in 3x.
- Energy: 1e-7 in 2x and 3x
for comparison, energy conservation in main is 1e-5, so this PR is 100x better. This proves the main point, switching to fem_poisson_perp improves energy conservation.

phi continuity seems to be no worse than in main. It's not good near the TS boundary. I think what we could explore in the future is 1D smoothing along x-y corners as @gwhammett has suggested.

Antoinehoff (approved):

Antoinehoff:
Did some last sanity check, everything looks good, thanks @manauref for this hard work! :)

PR #934, opened 2026-01-26
Extending the orbit average phase and pseudo orbit averaging to 2x, multispecies simulations, and tandem mirrors

### LINKED ISSUE #793: [DR] Pseudo-orbit average (POA) multiscale algorithm for magnetic mirrors
# Design Review

## Overview

Description: This DR aims to describe the implementation of the pseudo-orbit average (POA) algorithm. The idea is to have a full dynamics phase (FDP) and an orbit-averaged phase (OAP). We alternate between the two phases to reach steady state.

Motivation: The magnetic mirror simulations are slow and we want to make them faster. We have explored a few mutli-scale algorithms and this one works well to capture the dynamics of the expander and the center of the mirror.

Impact: Many files will be affected in this change. 

- A new timestepper will be added to the gyrokinetic/apps. This time stepper will be a superset of the SSPRK3, where this will pass parameters to the SSPRK3 stepper to toggle on and off certain parameters.
- Consideration must be taken when implementing the time stepper that we will aim to do a multi-species telescopic integraation scheme, freezing certain species at certain times.
- An updater is added to gyrokinetic/zero for determining the loss cone mask
- gk_species.c is modified to multiply the collisionless terms by a scalar number $\alpha$, so that the equation evolved is like $\partial_t f = \alpha \left\lbrace H, ~ f\right\rbrace + C + S + D$
- An app is added to multiply the whole GK equation by a multiplicative function $\mathcal{M}(x,v)$.
- An app is added to add a damping term to the RHS of the GK equation: $\partial_t f = \dots - \nu(x,v) f$.

## Proposed Design

Key Design Elements: To the user, this should present as an enum option in the first layer of the input struct `.time_stepper = { .id = GK_PSEUDO_ORBIT_AVERAGE'. An option for SSP_RK3 should be added as well.
This should be in a struct where we can specify a number $\alpha$. In theory, $\alpha$ should be allowed to be different for each species in the simulation (thinking ahead for multi-species integration). The structure to specify the mask to apply during the OAP should be in this struct and depends on the species.

The way frames are output needs to be modified for this time stepper. It makes sense that the user would want N frames per phase, so each OAP gives N frames, and each FDP gives N frames. This goes against what we are doing in the input file for frame generation, so we must embed the input file loop inside the time stepper. From a software engineering perspective, this is very tricky, and I'm not sure what the appropriate design is for this.

Interaction with Existing Code: SSPRK3 will be the default option and it should be the default. Existing code should work out-of-the-box.

Prototyping Efforts (if applicable): Branch [ `gk_dam_scale_hamil-prototyping`](https://github.com/ammarhakim/gkeyll/blob/gk_damp_scale_hamil-merge)

Sample Code:
```
  struct gkyl_gk app_inp = { 
    .time_stepper = {
      .id = GK_PSEUDO_ORBIT_AVERAGE,
      .frames_per_phase = 5,
      .scale_factor[2] = {alpha_ion, alpha_elc}, // Ions are index 0, electrons are index 1
      .damping[0] = { // The gk_damping_loss cone should distinguish where the trapped/passing boundaries are based on the charge of the species.
        .type = GKYL_GK_DAMPING_LOSS_CONE,
      }
      .damping[1] = { 
        .type = GKYL_GK_DAMPING_LOSS_CONE,
      }
    },
}

```

## Design Review Checklist 

Approval Criteria:

- [x] Is the design coherent and consistent with Gkeyll's existing architecture?
- [x] Does the design align with user-facing API standards?
- [x] Do functionalities overlap with existing features?
- [x] Are the algorithms robust and appropriate for the intended purpose?

## Additional Notes

More work that needs to be done before this branch is merged into main:

1) The location of the mirror throat is hardcoded right here https://github.com/ammarhakim/gkeyll/blob/gk_damp_scale_hamil-merge/gyrokinetic/apps/gk_species_fdot_multiplier.c#L165. To fix this, we can utilize code from the position map, perhaps moving it into a general method inside calc_bmag, to calculate the mirror throat automatically. This would be a straightforward task; however, we want this to generalize to 2x so that each field line has its own independent Bmax and location. I'm not sure what this structure should look like. Perhaps we find the extrema of the magnetic field at the nodes, then interpolate between.

2) We must identify the loss boundary with more accuracy. Scaling with a factor of 2 is not good enough. We can identify the "loss" cells through the corners of the cells, i.e., the nodal values, rather than the cell-centered coordinate. This may improve accuracy.

3) Calculate the loss boundary for electrons in the expander region. There is a confined region of electrons here that traces a circle in phase space. Following the loss criterion arguments, you arrive at an equation

$$ \mu_{max} = \frac{m_s}{2 B(x)} \left(  \frac{2 e \phi} {m_s} - v_{||}^2 \right) $$

This equation comes from thinking about 

$$ m v^2 /2 < e \phi, $$

 then separate the velocity into parallel and perpendicular, then solve for $\mu$.

4) Put all this work into a complete time-stepper app to complement SSP-RK3. At the input file level, I think a user should just specify $\alpha$ and identify the time stepper ID. This new time stepper will modulate between the orbit-averaged phase and the full dynamics phase. I would expect an input like "frames_per_phase = f" where it prints f frames per OAP or FDP phase. This contradicts the current loop structure in the input files, so I'm not sure how to implement this. 
4) For beams and localized sources, we must compute the orbit-averaged source for the OAP. If we do not do this, the results will be terrible. We must average the source by integrating along orbit contours and taking the average flux of the source. We must compute

 $$\mathring{S}(\vec x,\vec v) = \frac{\oint S(\vec\ell_x,\vec\ell_v)d\vec{\ell}}{\oint d\vec{\ell}}. $$

The orbit contours are 

$$    d\ell_z = \frac{\partial \mathcal{H}(z,v_\parallel)}{\partial v_\parallel},\\
    d\ell_{v_\parallel} = - \frac{\partial \mathcal{H}(z,v_\parallel)}{\partial z}.
$$

Greg Hammett suggested an alternative. We could transform the source into $(E, \mu)$ coordinates and the orbit average will be much easier to perform.

## Links

Additional Resources: Drafts of the paper that implements this scheme in some toy models.

[Application_of_a_multiscale_pseudo_orbit_averaging_algorithm_for_partial_differential_equations_to_gyrokinetic_simulations_of_magnetic_mirrors.pdf](https://github.com/user-attachments/files/21514198/Application_of_a_multiscale_pseudo_orbit_averaging_algorithm_for_partial_differential_equations_to_gyrokinetic_simulations_of_magnetic_mirrors.pdf)
[JCP__Part_1__A_multiscale_pseudo_orbit_averaging_algorithm_for_partial_differential_equations (1).pdf](https://github.com/user-attachments/files/21514199/JCP__Part_1__A_multiscale_pseudo_orbit_averaging_algorithm_for_partial_differential_equations.1.pdf)

Maxwell-Rosen:
# Feature

## Summary

Purpose: The orbit average phase and pseudo orbit average scheme were first implemented for single species simulations. This must be generalized to kinetic electrons simulations, 2x, and tandem mirrors. This PR adds these capabilities

fixes https://github.com/ammarhakim/gkeyll/issues/793

## Implementation Details

Key changes: 
`gk_species_fdot_multiplier.c` and `gk_species_damping.c` are modified to use an `array_dg_find_peaks` object. This is necessary to identify the locations of the walls, the tandem mirror peak, and the inner mirror peak. Critically, the maximum bmag and phi were previously defined at a single point, the mirror throat. Now, Bmag and phi are arrays in 2x, or a single-component array in 1x. These must be global arrays because peak finding is a global operation (each process must see global bmag and phi to find phi at the peaks of bmag).

`array_dg_find_peaks.c` is introduced as a new method inside core. This takes any DG array and finds the peaks. Currently, it is limited to 2D and 1D arrays, but this would be very complicated in 3D. It takes your 2D array and constructs N*1D arrays for the peaks. We can then project another 2D array onto the peaks of the initial 2D array. Comprehensive unit tests are added for this core method. GPU unit tests are included. The GPU implementation is a bit redundant since peak finding is recursive (and done on CPU) and the projection would be faster on CPU since it wouldn't have to launch a CUDA kernel, but Gkeyll is constructed in such a way that memory must lie on the GPU when running GPU code, so it's difficult to have this object always live on host and copy the neccisary information over. I tried this and it had difficulty interfacing with the loss cone mask module.

`loss_cone_mask_gyrokinetic` is adjusted for the changes necessary for 2x mirrors, tandem, and kinetic electrons. This meant a big overhaul of this updater so it's working with arrays, as well as extra arguments for tandem and kinetic electrons.

Dependencies: `core/zero/array_dg_find_peaks.c`

Automated testing: four regression tests are added

1. `gk_mirror_boltz_elc_poa_2x2v_p1`
2. `gk_mirror_kinetic_elc_poa_1x2v_p1`
3. `gk_mirror_tandem_boltz_elc_poa_1x2v`
4. `gk_wham_kinetic_poa_1x2v_p1`

LLM details: I relied heavily on LLMs for this PR.  The find_peaks method is almost entirely written by LLMs, given a few examples. It's thoroughly tested to machine precision in the unit test and performs well.

## Community Standards

## Testing:

## Additional Notes

Here is some evidence that the mask works correctly for the CFL condition and is different for the electrons and ions.
```
pgkyl rt_gk_mirror_kinetic_elc_poa_1x2v_p1-ion_cflrate_2.gkyl sel --z0 0.0 pl --title "Ion Z=0" --xlabel "vpar" --ylabel "mu" &
pgkyl rt_gk_mirror_kinetic_elc_poa_1x2v_p1-ion_cflrate_2.gkyl sel --z0 1.5 pl --title "Ion Z=1.5" --xlabel "vpar" --ylabel "mu" &
pgkyl rt_gk_mirror_kinetic_elc_poa_1x2v_p1-elc_cflrate_2.gkyl sel --z0 0.0 pl --title "Elc Z=0.0" --xlabel "vpar" --ylabel "mu" &
pgkyl rt_gk_mirror_kinetic_elc_poa_1x2v_p1-elc_cflrate_2.gkyl sel --z0 1.5 pl --title "Elc Z=1.5" --xlabel "vpar" --ylabel "mu" &
```
image

When comparing the 1x2v test of `gk_wham_poa_1x2v_p1`, it reproduces the exact same mask, even in parallel on GPU. All tests here are over 4 GPUs on perlmutter
```bash
pgkyl tests-main/gk_wham_poa_1x2v_p1-ion_fdot_multiplier_10.gkyl tests-branch/gk_wham_poa_1x2v_p1-ion_fdot_multiplier_10.gkyl ev "f[0] f[1] -" info
f[0] f[1] - (default#2)
├─ Time: 6.002000e-06
├─ Frame: 10
├─ Number of components: 1
├─ Number of dimensions: 3
├─ Grid: (uniform)
│  ├─ Dim 0: Num. cells: 288; Lower: -2.000000e+00; Upper: 2.000000e+00
│  ├─ Dim 1: Num. cells: 32; Lower: -1.000000e+00; Upper: 1.000000e+00
│  └─ Dim 2: Num. cells: 32; Lower: 0.000000e+00; Upper: 1.000000e+00
├─ Maximum: 0.000000e+00 at (0, 0, 0)
├─ Minimum: 0.000000e+00 at (0, 0, 0)
├─ geometry_type: 2
├─ geqdsk_sign_convention: 0

pgkyl tests-main/gk_wham_poa_1x2v_p1-ion_fdot_multiplier_10.gkyl tests-branch/gk_wham_poa_1x2v_p1-ion_fdot_multiplier_10.gkyl ev "f[0] f[1] -" info
f[0] f[1] - (default#2)
├─ Time: 6.002000e-06
├─ Frame: 10
├─ Number of components: 1
├─ Number of dimensions: 3
├─ Grid: (uniform)
│  ├─ Dim 0: Num. cells: 288; Lower: -2.000000e+00; Upper: 2.000000e+00
│  ├─ Dim 1: Num. cells: 32; Lower: -1.000000e+00; Upper: 1.000000e+00
│  └─ Dim 2: Num. cells: 32; Lower: 0.000000e+00; Upper: 1.000000e+00
├─ Maximum: 0.000000e+00 at (0, 0, 0)
├─ Minimum: 0.000000e+00 at (0, 0, 0)
├─ DG info:
│  ├─ Polynomial Order: 0
├─ Geometry info:
│  ├─ Type: GKYL_GEOMETRY_MIRROR
│  ├─ GEQDSK sign convention: 0
```

The tandem mirror loss mask is reproduced well.

$\mu_{\rm left}    = \max_{i<k}\left(\frac{\frac{1}{2}m_s v_\parallel^2 + q_s(\phi(z) - \phi_{mi})}{B_{mi} - B(z)}\right)$
$\mu_{\rm right} = \max_{i>k}\left(\frac{\frac{1}{2}m_s v_\parallel^2 + q_s(\phi(z) - \phi_{mi})}{B_{mi} - B(z)}\right)$
$\mu = \min\left(\mu_{\rm left} ,\mu_{\rm right}  \right)$

`pgkyl gk_mirror_tandem_boltz_elc_poa_1x2v-bmag.gkyl interp pl --title "|B| for tandem mirror" --xlabel "Z"`
image

`pgkyl gk_mirror_tandem_boltz_elc_poa_1x2v-ion_fdot_multiplier_10.gkyl sel --z2 12 pl --xlabel "Z" --ylabel "computational vpar" --title "Loss cone mask of tandem mirror"`
--c2p-vel does not work for this figure because it is c0.
image

Compute sanitizer clean on perlmutter
compute-sanitizer --tool memcheck --leak-check full ./cuda-build/gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_2x2v_p1 -g -s1
========= COMPUTE-SANITIZER

Running phase 0 @ t = 0.000000000e+00 ... 
Taking time-step 1 at t = 0 ... dt = 0.000132681

Number of update calls 1
Number of forward-Euler calls 3
Number of write calls 9
Timing: [removed]
========= LEAK SUMMARY: 0 bytes leaked in 0 allocations
========= ERROR SUMMARY: 0 errors

`rt_gk_mirror_boltz_elc_damped_1x2v_p1` runs as usual on 4 GPU.

This produces reasonable results for the 2x2v mirror
`pgkyl gk_mirror_boltz_elc_poa_2x2v_p1-ion_fdot_multiplier_2.gkyl sel --z1 0.0 --z3 14 pl --xlabel "psi" --ylabel "vpar"`
image
`pgkyl gk_mirror_boltz_elc_poa_2x2v_p1-ion_fdot_multiplier_2.gkyl sel --z0 0 --z3 14 pl --xlabel "Z" --ylabel "vpar"`
image


Maxwell-Rosen:
Runregression passes as of this time

rt_gk_mirror_boltz_elc_damped_1x2v_p1 fails, but the loss cone mask is depricated in another branch (low pass filter)

Maxwell-Rosen:
This branch works, but I'm working on an improvement based on the $\mu B + q \phi$ potential which will include arbitrary numbers of turning points, which is relevant for beam heating simulations

Issue #935, opened 2026-01-30
[DR] Change geo types in core/ and give .gkyl files produced by gyrokinetics/ the geo type


manauref:
I'd like to change this struct in core

```
enum gkyl_geometry_id {
  GKYL_TOKAMAK, // Tokamak Geometry from Efit
  GKYL_MIRROR, // Mirror Geometry from Efit
  GKYL_MAPC2P, // General geometry from user provided mapc2p
  GKYL_GEOMETRY_FROMFILE, // Geometry from file
};
```

to
```
enum gkyl_geometry_id {
  GKYL_GEOMETRY_NONE, // No special geo, defaults to Cartesian.
  GKYL_GEOMETRY_TOKAMAK, // Tokamak Geometry from Efit
  GKYL_GEOMETRY_MIRROR, // Mirror Geometry from Efit
  GKYL_GEOMETRY_MAPC2P, // General geometry from user provided mapc2p
  GKYL_GEOMETRY_FROMFILE, // Geometry from file
};
```

If people don't like it i can also make a new struct like this in gyrokinetics/ solely for GK.

Also, I want to give the .gkyl metadata this enum, so that it's easier to postprocess in postgkyl.

PR #936, opened 2026-01-30
Fix git ID and build date, and add geometry ID, to GK solver metadata

manauref:
see issue #935 

We change geo types to
```
// Identifiers for specific geometry types
enum gkyl_geometry_id {
  GKYL_GEOMETRY_NONE = 0, // No geometry, use Cartesian.
  GKYL_GEOMETRY_TOKAMAK, // Tokamak Geometry from Efit
  GKYL_GEOMETRY_MIRROR, // Mirror Geometry from Efit
  GKYL_GEOMETRY_MAPC2P, // General geometry from user provided mapc2p
  GKYL_GEOMETRY_FROMFILE, // Geometry from file
};
```

We also enrich the metadata to have this info. Now doing info on a mapc2p sim file gives something like this
image

and doing info on a sim with numerical equilibrium gives
image

this additional metadata will help us do postprocessing more easily.

### DIFF
--- a/gyrokinetic/apps/gk_species_damping.c
+++ b/gyrokinetic/apps/gk_species_damping.c
@@ -158,6 +158,7 @@ gk_species_damping_init(struct gkyl_gyrokinetic_app *app, struct gk_species *gks
         .vel_map = gks->vel_map,
         .bmag = app->gk_geom->geo_int.bmag,
         .bmag_max = damp->bmag_max,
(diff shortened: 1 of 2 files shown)


manauref:
Several key regression tests, using MAPC2P, TOKAMAK, MIRROR and FROMFILE geometry are valgrind and compute-sanitizer clean.

Restarts also seem to work fine.

PR #943, opened 2026-02-04
Write Psi produced by EFIT and metadata 

akashukla:
This PR is a small addition: we write the psi.gykl file from psi(R,Z) produced during geometry generation along with some metadata (like psisep). Also add the efit name to the geometry metadata so that we can use pgkyl commands to easily overlay nodes on flux surfaces.

### DIFF
--- a/gyrokinetic/zero/gk_geometry.c
+++ b/gyrokinetic/zero/gk_geometry.c
@@ -567,27 +567,6 @@ gkyl_gk_geometry_deflate(const struct gk_geometry* up_3d, struct gkyl_gk_geometr
   return up;
 }
 
-bool gkyl_gk_geometry_check_name(char *name){
-  size_t len = strlen(name);
-  int i = len - 1;
-  int digit_count = 0;
-  while (i >= 0 && isdigit((unsigned char)name[i])) {
-    i--;
-    digit_count++;
-  }
-  if (digit_count > 0 && i >= 1 && name[i] == 'b' && name[i-1] == '_') {
-    const char *num_str = &name[i + 1];
-    int num = atoi(num_str);
-    if ( num == 0)
-      return true;
-    else
-      return false;
-  }
-  else {
-    return true;
-  }
-}
-
 void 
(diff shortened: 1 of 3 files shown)


gyrokinetic/zero/gk_geometry.c
```diff
@@ -13,6 +13,7 @@
 #include <gkyl_alloc_flags_priv.h>
 #include <assert.h>
 #include <float.h>
+#include <mpack.h>
```
manauref: don't include mpack, include gkyl_util.h (unless it's already included indirectly)

gyrokinetic/zero/gk_geometry.c
```diff
@@ -566,6 +567,34 @@ gkyl_gk_geometry_deflate(const struct gk_geometry* up_3d, struct gkyl_gk_geometr
...
+    { .key = "simag", .elem_type = GKYL_MP_DOUBLE, .dval = efit->simag},
+  };
+  int io_meta_rz_len = sizeof(io_meta_rz)/sizeof(io_meta_rz[0]);
+  int io_meta_len[] = {io_meta_basic_len, io_meta_rz_len};
+  const struct gkyl_msgpack_map_elem* io_meta[] = {io_meta_basic, io_meta_rz};
+  struct gkyl_msgpack_data *mt = gkyl_msgpack_create_union(sizeof(io_meta_len)/sizeof(int), io_meta_len, io_meta);
+  gkyl_grid_sub_array_write(&efit->rzgrid, &efit->rzlocal, mt, efit->psizr, fileNm);
+
+  MPACK_FREE(mt->meta);
+  gkyl_free(mt);
```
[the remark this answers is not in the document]
akashukla: Fixed as discussed. And fixed other calls to use the gkyl_msgpack_data_release function.

gyrokinetic/zero/gk_geometry.c
akashukla: fixed

Antoinehoff (approved):

PR #946, opened 2026-02-10
Time dilation methods to suppress CFL

### LINKED ISSUE #895: [DR] Time dilation methods to control CFL in inaccessible regions of phase space
# Design Review

## Summary

Late on a Friday night, I had a research meeting with Ammar, Mana and Jimmy in my sleep. It was mostly with Ammar, but Jimmy and Mana were there. I explained to them the issues we had in G2 when we implemented force softening. In the 2023 paper, we published the results below.
image
image
The issue is that the force softening introduced higher temperatures in the expander. Expander heating is something we are incredibly concerned about right now because of the positivity hack, so this scheme is not ideal. I explained in my dream that the issue was that we are modifying the Hamiltonian, therfore modifying the orbit contours, and the orbit contours are essential for particle confinement. We really should not modify the Hamiltoninan for an appropriate scheme.

Instead, I was reminded of the work in this article (“Time-Dilation Methods for Extreme Multiscale Timestepping Problems” https://arxiv.org/html/2510.09756). They essentially discuss slowing down time near the extreme regions of the problem, effectively what we have done for the orbit average phase of the pseudo orbit averaging integrator. This method preserves the orbit contours, masks the CFL, and will speed up our simulations. Ammar was doubtful this would work and dismissed my idea, however I was very confident that this would work, so he wanted to see some evidence. I awoke to program in the middle of the night, teeming with excitement and motivation.

I realize another issue is that we used an overly harsh mask in the 2023 paper. It is not neccisary to create a completely logical mask, going all the way from 1 to 0. Instead, the mask should optimally choose its strength based on the CFL frequency and rate. It should modify the problem the bare minimum in order to get the optimal CFL rate. That lead me to this modification of the GK vlasov equation.

$$ \frac{\partial f}{\partial t} =\rm{min}(1, \Delta t_{\Omega_H} / \Delta t_{\rm CFL}) [H, f ] + C + S  $$

This slows down the advective Poisson bracket by the appropriate ratio when $\Delta t_{\mathrm{CFL}} < \Delta t_{\Omega_H}$. This numerical approach to the mask is also more flexible, as the code or user does not need to know about the mirror loss cone, therefore it is a more general method of time-dilation in inaccessible regions of phase space.

Prototyping branch: https://github.com/ammarhakim/gkeyll/tree/slowdown-to-omega-cfl

## Implementation Details

Key changes: To calculate $\rm{min}(1, \Delta t_{\Omega_H} / \Delta t_{\rm CFL}) $, some modifications are needed inside the array_ops and dg_bin_ops methods.

We need an inversion operation to invert a P0 array. This should use the dg_inv method, which is already written. However, there is no kernel for the P0 array. Division is tricky with DG arrays, so it should not be an array_op, even though this is a P0 array. I handwrote a P0 kernel for this method because the Maxima generators do not work for P0. It just does 1/f[0], which is a straightforward operation and I really don't want to make the operation 1/f[0] more complex then neccisary.

This is followed by a $\rm min()$ operation, so that also needed to be implemented. I decided on the location for this function inside array_ops for a few reasons. One, this is an operation on a single array. Second, it is a simple operation, not requiring a basis. Third, I really don't want to think about how this generalizes to a P1 array and implement a full kernel operation. I see the issue that this method is only intended for P0 arrays, however I think that is the purpose of these array_ops methods. It is for single array manipulation, independent of basis. I really don't want to make this method implementation more complex then neccisary and do P1,P2 kernels and write Maxima scripts when the operation is just `fmin( coeff, array)' element wise.

The array methods are determined within gk_species_collisionles.c inside gk_species_collisionless_fdot_scaling_enabled. We could create a seperate function for this method, but this works well.

Image

Dependencies: None

Automated testing: New unit tests are added to ctest_array_ops.c. A regression test is used to test the mask of the CFL.

## Example Use

```    
    .collisionless = {
      .cfl_dt_min_value = 1e-8, // User specified minimum dt
      .write_diagnostics = true,
      .cfl_dt_min_omegaH = true, // To limit the CFL based on the omegaH mode.
    },
```

## Preliminary testing

Ammar from my dream wanted testing results to see if this is a good idea. This figure compares two simulations, one with the time dilation feature, the other without it. This is a non-uniform z simulation with 216 cells using the positivity hack. The simulation executed 10 POA cycles, with a 20 μs FDP and 2 s OAP.  The simulation ends around 20 s. The simulations compare nearly identically, except for minor deviations in Tpar in the mirror throat. The time dilation saves 20x during the FDP and the simulation goes from 22 hours to around 5-6 hours. It's a big cost saving,
image

Maxwell-Rosen:
# Feature

## Summary

Purpose: This PR adds methods to dilate time in extreme regions of phase space to control CFL.

Issue link: Based on #894, fixes #895.

## Implementation Details

Math basis: [Time_dilation_notes (1).pdf](https://github.com/user-attachments/files/25216095/Time_dilation_notes.1.pdf)
Essentially, we multiply fdot by a small factor beta, determined by the CFL condition.

Key changes: The time dilation is contained inside the fdot multiplier. Parameters are added to the `time_rate_multiplier` struct in the input file, which contains everything relevant for time dilation.

New operations are added to array ops. `gkyl_array_min_by_cell`, `gkyl_array_min_by_cell_range`, `gkyl_array_invert_by_cell` with appropriate unit tests. These are neccisary to compute 

$$
{\rm min} (1, \Omega_{max} / \Omega_{cfl} )
$$

Utilizes the masks introduced in #889.

Dependencies: Mask, array ops, fdot_multiplier

Automated testing: This feature is added to multiple mirror regression tests, which include every combination of time dilation configurations

## Example Use

```
...
    .time_rate_multipliers = {
      .num_multipliers = 1,
      .multiplier[0] = {
        .type = GKYL_GK_FDOT_MULTIPLIER_FIXED_FACTOR_TIMES_OMEGA_MAX, // Many types, located in gyrokinetic.h
        .cellwise_const = true,
        .write_diagnostics = true,
        .time_dilation_scale_const = 0.05, // Scales time overall by a factor 0.05
        .cfl_factor_times_omega_max = 0.1, // Dilates time for cells with a timestep smaller than  10x the CFL-restricted timestep
      },
    },
}
```

## Community Standards

## Testing:

## Additional Notes

I'm comparing two simulations here. One is regular, the other with time dilation. It uses `GKYL_GK_FDOT_MULTIPLIER_MASK_F_FRAC_LOCAL` with `f_threshold=1e-3`. The first figure is for a maxwellian source R=32 mirror simulation with the pseudo orbit averaging scheme.
image
Subtracting the two lines, we see the percent error (f_time-dilated - f_orig)/f_orig
image

The two lines are nearly identical, but the timestep in the FDP is 3.37466e-10 versus 5.24247e-11. I have found this to be a stable amount of speed up for this problem.

Here are the same two figures, but with the beam source.
image
There are more errors in the beam simulation, but other than right at the mirror throat, they are small.


JunoRavin:
I think this PR is basically ready-to-go. If you can just do a couple quick checks that with the merge in of recent pushes to main that nothing is behaving differently than you expect (and where appropriate utilizing the runregression system to compare the output of gyrokinetic simulations in this branch to those in main), then I think this is ready to merge.

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -351,18 +351,47 @@ struct gkyl_gyrokinetic_damping {
...
+  double time_dilation_scale_const; // A constant which multiplies all of fdot and cfl to dilate time. Small number (0,1] so that dt *= 1/time_dilation_scale_const
+  double cfl_factor_times_omega_max; // For FIXED_FACTOR_TIMES_OMEGA_MAX: the factor multiplied by the maximum characteristic frequency to get the minimum dt value.
+
+  char dt_set_by_species[32]; // For DT_SET_BY_SPECIES: the name of the species whose dt we are using to set the maximum dt for this species. e.g. "elc"
+};
+
+// Parameters for a chain of df/dt multipliers. Components are applied in order.
+struct gkyl_gyrokinetic_fdot_multipliers {
+  int num_multipliers;
+  double omega_H_scale_const; // A constant which multipliers the omegaH frequency
```
manauref: I think this is the wrong place to put this.
1. If the intention is to simply relax the omega_H restriction, you can use the cfl_frac_omegaH input.
2. If the intention is to modify the contribution from this species to the omega_H restriction, we simply need to multiply such contribution by the fdot multiplier. I did this to slow down the electrons in LTX simulations, see the use of `gk_species_fdot_multiplier_advance_times_cfl_conf` in `gk_species_omegaH_dt` in the branch `slow_elc` i just pushed, and feel free to bring that stuff into this PR.

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -351,18 +351,47 @@ struct gkyl_gyrokinetic_damping {
...
+  GKYL_GK_FDOT_MULTIPLIER_FIXED_DT,           // dt floor from user-specified cfl_dt_min_value. Specify cfl_dt_min_value
+  GKYL_GK_FDOT_MULTIPLIER_FIXED_FACTOR_TIMES_OMEGA_MAX, // dt floor from user-specified factor times maximum characteristic frequency. Specify cfl_factor_times_omega_max.
+  GKYL_GK_FDOT_MULTIPLIER_FIXED_DT_OMEGAH,    // dt floor from cyclotron frequency omega_H. Dilates time. No need to specify anything else
+  GKYL_GK_FDOT_MULTIPLIER_DT_SET_BY_SPECIES,  // Set the dt floor based on the dt from another species. That species must be updated before this species. i.e. if the ions scale dt based on electrons, you must specify in the app `.species = {elc, ion},` so the electrons are updated first.
+  GKYL_GK_FDOT_MULTIPLIER_MASK_F_THRESHOLD,   // Dilates time in cells where |J_tot*f| < threshold. Specify f_threshold
+  GKYL_GK_FDOT_MULTIPLIER_MASK_F_FRAC_LOCAL,  // Dilates time in cells where |J_tot*f| < threshold * local_max. Spatially dependent mask. Specify f_threshold
+  GKYL_GK_FDOT_MULTIPLIER_MASK_F_FRAC_GLOBAL, // Dilates time in cells where |J_tot*f| < threshold * global_max. Specify f_threshold
```
manauref, line 378: after much experimentation, do we still want to keep all these options? I only see 1-2 being commonly useful, yet you overwhelm the user with so many options, and make the code harder to maintain. If we are only going to use 1-2 of these, I think we should delete the rest.

also, comments above refer to omega_H as the cyclotron frequency. The cyclotron frequency, omega_c = q*B/m, is something else.

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -351,18 +351,47 @@ struct gkyl_gyrokinetic_damping {
...
+  double f_threshold; // For MASK_F_* types: absolute value (THRESHOLD)
+    // or fraction 0-1 (FRAC_LOCAL, FRAC_GLOBAL).
+  double time_dilation_scale_const; // A constant which multiplies all of fdot and cfl to dilate time. Small number (0,1] so that dt *= 1/time_dilation_scale_const
+  double cfl_factor_times_omega_max; // For FIXED_FACTOR_TIMES_OMEGA_MAX: the factor multiplied by the maximum characteristic frequency to get the minimum dt value.
+
+  char dt_set_by_species[32]; // For DT_SET_BY_SPECIES: the name of the species whose dt we are using to set the maximum dt for this species. e.g. "elc"
+};
+
+// Parameters for a chain of df/dt multipliers. Components are applied in order.
+struct gkyl_gyrokinetic_fdot_multipliers {
```
manauref: this name differs from that of the previous struct by a single letter. Please change it to something like `gkyl_gyrokinetic_fdot_multiplier_array`. Or perhaps is better to simply call this one `gkyl_gyrokinetic_fdot_multiplier`, which is simpler and is the user-facing one, and rename the one above it `gkyl_gyrokinetic_fdot_multiplier_inp` or `gkyl_gyrokinetic_fdot_multiplier_op`.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1039,7 +1061,8 @@ struct gk_species {
 
   struct gk_damping damping; // Damping term: -nu(z)*f.
 
-  struct gk_fdot_multiplier fdot_mult; // Function multiplying df/dt.
+  int num_fdot_mult; // Number of active df/dt multipliers in chain.
+  struct gk_fdot_multiplier fdot_mult[GKYL_MAX_FDOT_MUL]; // Functions multiplying df/dt.
```
manauref: I think the way multiple fdot multipliers is implemented is incorrect. You are making the species aware of having multiple fdot_mult ops, and in some places asking the species to loop. That is not right. The correct way to do this is for the multiple fdot_mult ops to be managed internally within a single fdot_multiplier object. That is what we do with sources; for example, instead of
```
  for (int k=0; k<app->gk_species[i].num_sources; ++k)
      gk_species_source_rhs( ..., k )
```
we do
```
      gk_species_source_rhs( ...)
```
and internally, inside `gk_species_source`, there are loops over sources.

Please change this.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1094,7 +1117,10 @@ struct gk_species {
   void (*gyroaverage)(gkyl_gyrokinetic_app *app, struct gk_species *species,
     struct gkyl_array *field_in, struct gkyl_array *field_gyroavg);
 
+  double dt_omegaH; // dt_omegaH. Recorded at the end of the rhs evaluation.
+  double time_dilation_scale_const; // A constant which multiplies all of fdot and cfl to dilate time.
```
manauref: this shouldn't be here. It's specific to time dilation, so it should live in the time dilation app/struct/object.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1094,7 +1117,10 @@ struct gk_species {
...
   double *omega_cfl; // Maximum Omega_CFL in this MPI process.
+  double *dt_cfl_global_ho; // Global maximum Omega_CFL across all MPI processes.
```
manauref: same here, this is used by the fdot multiplier app, so it shouldn't live in the species struct.

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -351,18 +351,47 @@ struct gkyl_gyrokinetic_damping {
...
-  GKYL_GK_FDOT_MULTIPLIER_USER_INPUT,
-  GKYL_GK_FDOT_MULTIPLIER_LOSS_CONE,
+  GKYL_GK_FDOT_MULTIPLIER_NONE = 0,           // No multiplier applied.
+  GKYL_GK_FDOT_MULTIPLIER_USER_INPUT,         // User-provided static profile M(z) via function pointer.
+  GKYL_GK_FDOT_MULTIPLIER_LOSS_CONE,          // M=1 in loss cone, M=0 in confined region.
+  GKYL_GK_FDOT_MULTIPLIER_CONSTANT,           // Dilates time by a fixed constant, time_dilation_scale_const.
+  GKYL_GK_FDOT_MULTIPLIER_FIXED_DT,           // dt floor from user-specified cfl_dt_min_value. Specify cfl_dt_min_value
+  GKYL_GK_FDOT_MULTIPLIER_FIXED_FACTOR_TIMES_OMEGA_MAX, // dt floor from user-specified factor times maximum characteristic frequency. Specify cfl_factor_times_omega_max.
+  GKYL_GK_FDOT_MULTIPLIER_FIXED_DT_OMEGAH,    // dt floor from cyclotron frequency omega_H. Dilates time. No need to specify anything else
+  GKYL_GK_FDOT_MULTIPLIER_DT_SET_BY_SPECIES,  // Set the dt floor based on the dt from another species. That species must be updated before this species. i.e. if the ions scale dt based on electrons, you must specify in the app `.species = {elc, ion},` so the electrons are updated first.
```
manauref: this is not viable. It's extremely hard to guarantee species update order, especially if we parallelize over species. Please remove this option.

manauref (changes requested):
1. I don't see where `gkyl_array_min_by_cell` is used. Please delete if not used.
2. Same for `gkyl_array_invert_by_cell`. Also keep in mind that there's a more general array_ops in the sundials PR that can do this.
3. You are assigning `advance_threshold_func_cu` in dg_array_mask, but where is it used? if not used, delete it.
4. I think the multiple fdot_mult capability is not implemented correctly, see detailed comment below.

Maxwell-Rosen:
> * I don't see where `gkyl_array_min_by_cell` is used. Please delete if not used.

https://github.com/ammarhakim/gkeyll/blob/time-dilation-reorg/gyrokinetic/apps/gk_species_fdot_multiplier.c#L131

> * Same for `gkyl_array_invert_by_cell`. Also keep in mind that there's a more general array_ops in the sundials PR that can do this.

https://github.com/ammarhakim/gkeyll/blob/time-dilation-reorg/gyrokinetic/apps/gk_species_fdot_multiplier.c#L129

Sounds like the sundials branch should have been PRd as smaller functions.

> * You are assigning `advance_threshold_func_cu` in dg_array_mask, but where is it used? if not used, delete it.

Removed. It doesn't look lit it was being used. It must have been left over from another version.

> * I think the multiple fdot_mult capability is not implemented correctly, see detailed comment below.

I disagree. It is all implemented correctly. It's good to support multiple types of fdot multipliers, as each situation requires something different. I already discussed with @JunoRavin  that he is interested in some of these other options that I'm not using for production simulations.

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Removed

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Yes. They are all useful.

Comments updated. Good catch!

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Changed to `gkyl_gyrokinetic_fdot_multiplier_array`

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: Species can have dt_omegaH. This is a property of the species, not the time dilation. I needed to add this as a species property so species can know the omegaH frequency of another species.

Removed time_dilation_scale_const

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: Again, the CFL rate is a property of the species, not the fdot multiplier. The fdot multiplier is not enforcing a CFL rate. It wouldn't make sense for the species to store its timestep in the fdot multiplier object.

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: That's why the comment is specified. I'm using it in kinetic electron simulations so that we only time-dilate the electrons and not the ions.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: I'll reorganize this. The array method is necessary for kinetic electron simulations, where we need to put multiple multipliers on the electrons relative to the ions.

Maxwell-Rosen:
I have been asked by @manauref how severe the ordering restriction is for the `GKYL_GK_FDOT_MULTIPLIER_DT_SET_BY_SPECIES` option. Gkeyll aims for the stack to be independent of species order, so I understand why this is a bad design choice. I left the comment

```
  GKYL_GK_FDOT_MULTIPLIER_DT_SET_BY_SPECIES,  // Set the dt floor based on the dt from another species. That species must be updated before this species. i.e. if the ions scale dt based on electrons, you must specify in the app `.species = {elc, ion},` so the electrons are updated first.
```

While I think a good comment in the right place makes it clear how users should use this feature, it does require the user to set a specific order for the species, which is bad design. However, this is not a severe restriction. If species are assigned in the opposite order (i.e. ions are time dilated to electron speeds, but we set `.species = {ion, elc}`), then the effect is that the ions are time-dilated according to the dt (as set by the species_rhs update) of the electrons from the previous RK stage. This is not a problem and results in either a slightly more restricted time step than optimal or slightly enhanced time dilation. This does not affect stability.

To show that this is not an issue, here is a kinetic electron simulation in which the ions are time-dilated to the electron time step. The initial condition is the final result of a POA equilibrium. Here, an FDP is evolved for 10 $\mu s$. No meaningful change in the BiMaxwellian moments occurs, indicating that misordering the species has no deleterious effects and is fine. 

image

Leaving a good comment to explain the limitations and usage of features is alright. It would be an even worse design choice to implement a species-update-order shuffler. Implementing such a feature communicates to developers that it is okay to have order-dependent updates.

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -385,8 +413,8 @@ struct gkyl_gyrokinetic_species {
   // Initial conditions from a file.
   struct gkyl_gyrokinetic_ic_import init_from_file;
 
-  // Phase-space field multiplying df/dt.
-  struct gkyl_gyrokinetic_fdot_multiplier time_rate_multiplier;
+  // Ordered list of phase-space fields multiplying df/dt.
+  struct gkyl_gyrokinetic_fdot_multiplier_array time_rate_multipliers;
```
manauref: rename `time_rate_multiplier`
just like we have `source`, not sources

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -154,51 +286,61 @@ gk_species_fdot_multiplier_init(struct gkyl_gyrokinetic_app *app, struct gk_spec
       gkyl_array_copy(fdmul->multiplier, fdmul->multiplier_host);
 
       fdmul->advance_times_cfl_func = gk_species_fdot_multiplier_advance_mult;
+      // if (gks->collisionless.scale_fac > 0.0 && gks->collisionless.scale_fac < 1.0) {
+      //   fdmul->advance_times_omegaH_func = gk_species_fdot_multiplier_advance_omegaH_mult;
+      // }
```
manauref: why is this commented out?

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -848,8 +849,9 @@ struct gk_damping {
   void (*write_func)(gkyl_gyrokinetic_app* app, struct gk_species *gks, double tm, int frame);
 };
 
-struct gk_fdot_multiplier {
+struct gk_fdot_multiplier_comp {
   enum gkyl_gyrokinetic_fdot_multiplier_type type; // Type of multiplicative function term.
+  int component_id; // Index of this multiplier in the chain.
```
manauref: delete this
the only reason you have it so that you can put an index in the name of the file written for each multiplier. Just pass the index as an argument in this call
```
fdmul->write_func(app, gks, fdmul, tm, frame);
```

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -848,8 +849,9 @@ struct gk_damping {
...
   bool write_diagnostics; // Whether to write diagnostics out.
   bool evolve; // Whether the multiplicative function is time dependent.
   struct gkyl_array *multiplier; // Damping rate.
```
manauref: each multiplier has a separate array. But can't we have a single array for all the multipliers, and multiply all the multipliers into it?

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -351,18 +351,46 @@ struct gkyl_gyrokinetic_damping {
...
+  double f_threshold; // For MASK_F_* types: absolute value (THRESHOLD)
+    // or fraction 0-1 (FRAC_LOCAL, FRAC_GLOBAL).
+  double time_dilation_scale_const; // A constant which multiplies all of fdot and cfl to dilate time. Small number (0,1] so that dt *= 1/time_dilation_scale_const
+  double cfl_factor_times_omega_max; // For FIXED_FACTOR_TIMES_OMEGA_MAX: the factor multiplied by the maximum characteristic frequency to get the minimum dt value.
+
+  char dt_set_by_species[32]; // For DT_SET_BY_SPECIES: the name of the species whose dt we are using to set the maximum dt for this species. e.g. "elc"
+};
+
+// Parameters for a chain of df/dt multipliers. Components are applied in order.
+struct gkyl_gyrokinetic_fdot_multiplier_array {
```
manauref: i think in keeping with the names used inside the gkeyll solver it'd be best to rename
`gkyl_gyrokinetic_fdot_multiplier` -> `gkyl_gyrokinetic_fdot_multiplier_comp`
`gkyl_gyrokinetic_fdot_multiplier_array` -> `gkyl_gyrokinetic_fdot_multiplier`

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -232,54 +374,194 @@ gk_species_fdot_multiplier_init(struct gkyl_gyrokinetic_app *app, struct gk_spec
...
+    fdmul->advance_times_rate_func(app, gks, fdmul, phi, f, cflrate);
+  }
 
   app->stat.species_fdot_mult_tm += gkyl_time_diff_now_sec(wst);
-  
 }
 
 void
-gk_species_fdot_multiplier_write(gkyl_gyrokinetic_app* app, struct gk_species *gks, double tm, int frame)
+gk_fdot_multiplier_get_time_dilation_scale_const(gkyl_gyrokinetic_app *app,
```
manauref: just make this a double and return the value, rather than doing it by argument

gyrokinetic/apps/gk_species.c
```diff
@@ -132,11 +137,13 @@ gk_species_rhs_dynamic(gkyl_gyrokinetic_app *app, struct gk_species *species,
   else {
     omega_cfl_ho[0] = species->omega_cfl[0];
   }
+
   double dt_out = app->cfl/omega_cfl_ho[0];
   
-  // Enforce the omega_H constraint on dt.
```
manauref, line 137: restore the comment

gyrokinetic/apps/gk_species.c
```diff
@@ -743,6 +754,7 @@ gk_species_init_dynamic(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app
     gks->omega_cfl = gkyl_malloc(sizeof(double));
     gks->m0_max = gkyl_malloc(app->basis.num_basis*sizeof(double));
   }
+  gks->dt_cfl_global_ho = gkyl_malloc(sizeof(double));
```
manauref: why does this have to be a dynamically allocated variable?
it'd be simpler/safer if it wasn't

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -72,14 +91,119 @@ gk_species_fdot_multiplier_advance_loss_cone_mult(gkyl_gyrokinetic_app *app, con
...
-  gkyl_array_scale_by_cell(out, fdmul->multiplier);
+  // Offset by a constant factor
+  gkyl_array_scale(fdmul->multiplier, fdmul->time_dilation_scale_const);
+
+  // Multiply cflrate by the multplier.
+  gkyl_array_scale_by_cell(cflrate, fdmul->multiplier);
+}
+
+// Set the multiplier array based on cflrate and time dilation parameters.
+// See https://arxiv.org/html/2510.09756
```
manauref: best to put a short citation to the paper (e.g. author, paper title, year), rather than the arxiv URL.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -859,12 +861,37 @@ struct gk_fdot_multiplier {
   double *bmag_max; // Maximum magnetic field amplitude.
   double *bmag_max_coord; // Location of bmag_max.
   double *phi_m, *phi_m_global; // Electrostatic potential at bmag_max.
+
+  // Time dilation parameters (from input).
+  double cfl_dt_min_value; // User-specified minimum dt value.
+  double cfl_factor_times_omega_max; // User-specified factor multiplied by maximum characteristic frequency to get minimum dt.
+  double f_threshold; // Threshold for mask-based time dilation.
+  double time_dilation_scale_const;
```
manauref: missing description

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -859,12 +861,37 @@ struct gk_fdot_multiplier {
...
+    struct gk_fdot_multiplier_comp *fdmul, const struct gkyl_array *phi, const struct gkyl_array *f,
+    struct gkyl_array *cflrate);
   void (*advance_times_cfl_func)(gkyl_gyrokinetic_app *app, const struct gk_species *gks,
-    struct gk_fdot_multiplier *fdmul, const struct gkyl_array *phi, struct gkyl_array *out);
+    struct gk_fdot_multiplier_comp *fdmul, const struct gkyl_array *phi, const struct gkyl_array *f,
+    struct gkyl_array *cflrate);
+};
+
+struct gk_fdot_multiplier {
+  int num_multipliers; // Number of active df/dt multipliers in chain.
```
manauref: delete "active". It's not clear

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -72,14 +91,119 @@ gk_species_fdot_multiplier_advance_loss_cone_mult(gkyl_gyrokinetic_app *app, con
   gkyl_loss_cone_mask_gyrokinetic_advance(fdmul->lcm_proj_op, &gks->local, &app->local,
     phi, fdmul->phi_m_global, fdmul->multiplier);
 
-  // Multiply out by the multplier.
-  gkyl_array_scale_by_cell(out, fdmul->multiplier);
+  // Offset by a constant factor
+  gkyl_array_scale(fdmul->multiplier, fdmul->time_dilation_scale_const);
+
+  // Multiply cflrate by the multplier.
+  gkyl_array_scale_by_cell(cflrate, fdmul->multiplier);
```
manauref: I thought we were scaling the cflrate in a separate method, i.e. `gk_species_fdot_multiplier_advance_times_cfl`. Why is this being done here now?

core/zero/gkyl_array_ops.h
```diff
@@ -130,6 +130,14 @@ struct gkyl_array* gkyl_array_scale_by_cell(struct gkyl_array *out, const struct
  */
 struct gkyl_array* gkyl_array_divide_by_cell(struct gkyl_array *out, const struct gkyl_array *a);
 
+/**
+ * Compute element-wise reciprocal, out = 1/out. Returns out.
+ *
+ * @param out Output array.
+ * @return out array.
+ */
+struct gkyl_array* gkyl_array_invert_by_cell(struct gkyl_array *out);
```
manauref: this method should have an `inp` and an `out`. If the user wants to do an in-place operation, that's up to them, but this should also allow out-of-place operation

core/zero/gkyl_array_ops.h
```diff
@@ -140,6 +148,15 @@ struct gkyl_array* gkyl_array_divide_by_cell(struct gkyl_array *out, const struc
...
 struct gkyl_array* gkyl_array_shiftc(struct gkyl_array *out, double a, unsigned k);
 
+/**
+ * Take element-wise minimum: out = min(out, a). Returns out.
+ *
+ * @param out Output array.
+ * @param a Value to compare against each element.
+ * @return out array.
+ */
+struct gkyl_array* gkyl_array_min_by_cell(struct gkyl_array *out, double a);
```
manauref: this method should have an `inp` and an `out`. If the user wants to do an in-place operation, that's up to them, but this should also allow out-of-place operation

same for its `_range` equivalent

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -99,52 +223,60 @@ proj_on_basis_c2p_position_func(const double *xcomp, double *xphys, void *ctx)
 }
 
 void
-gk_species_fdot_multiplier_init(struct gkyl_gyrokinetic_app *app, struct gk_species *gks,
-  struct gk_fdot_multiplier *fdmul)
+gk_species_fdot_multiplier_init_comp(gkyl_gyrokinetic_app *app, struct gk_species *gks,
```
manauref, line 267: i think this can be static

gyrokinetic/apps/gk_species_fdot_multiplier.c
```diff
@@ -232,54 +374,194 @@ gk_species_fdot_multiplier_init(struct gkyl_gyrokinetic_app *app, struct gk_spec
...
+          fdmul->advance_times_cfl_func =
+            gk_species_fdot_multiplier_advance_time_dilation_cfl_f_frac_local;
+          break;
+        case GKYL_GK_FDOT_MULTIPLIER_MASK_F_FRAC_GLOBAL:
+          mask_type = GKYL_DG_ARRAY_MASK_C0_GREATER_FRAC;
+          fdmul->advance_times_cfl_func =
+            gk_species_fdot_multiplier_advance_time_dilation_cfl_f_frac_global;
+          break;
+        default:
+          mask_type = GKYL_DG_ARRAY_MASK_NONE;
```
manauref: shouldn't you throw an error here? my impression is the code shouldn't get here, if it does, it's an error

gyrokinetic/creg/rt_gk_mirror_boltz_elc_1x2v_p1.c
```diff
@@ -506,7 +506,7 @@ create_ctx(void)
   double mu_max_elc = me * pow(3. * vte, 2.) / (2. * B_p);
   int Nz = 32;
   int Nvpar = 32; // Number of cells in the paralell velocity direction 96
-  int Nmu = 48;  // Number of cells in the mu direction 192
+  int Nmu = 16;  // Number of cells in the mu direction 192
```
manauref, line 509: can you comment on the impact of the two changes made to this regression test? can you show a comparison of results? why are you making this change?

manauref (changes requested):
please check valgrind/compute sanitizer again when done

gyrokinetic/creg/rt_gk_mirror_boltz_elc_1x2v_p1.c
Maxwell-Rosen: It just makes the test faster. My production simulations don't even use 48 cells in mu

gyrokinetic/apps/gk_species_fdot_multiplier.c
Maxwell-Rosen: Error thrown. Fixed

gyrokinetic/apps/gk_species_fdot_multiplier.c
Maxwell-Rosen: Made static. Fixed

core/zero/gkyl_array_ops.h
Maxwell-Rosen: Updated. Fixed

core/zero/gkyl_array_ops.h
Maxwell-Rosen: Updated. Fixed

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Updated all fields to reflect the distinction between multiplier and multiplier_comp

gyrokinetic/apps/gk_species_fdot_multiplier.c
Maxwell-Rosen: Removed

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: Changed by using a larger scale "multiplier" which presides over all the components of the multiplier. Then, the write method doesn't need the component_id

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: Changed so that the multiplier only has 1 big array that multiplies everything, but we do need some scratch working storage for internal calculations in the advance methods.

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Changed so that we aren't using the `gkyl_gyrokinetic_fdot_multiplier_array` name anymore. We just have `multiplier` and `multiplier_comp`.

gyrokinetic/apps/gk_species_fdot_multiplier.c
Maxwell-Rosen: Done

gyrokinetic/apps/gk_species.c
Maxwell-Rosen: Restored

gyrokinetic/apps/gk_species.c
Maxwell-Rosen: Done. There are a few other variables in the header that are used similarly, which are dynamically  allocated, which is why I malloc-d this memory.

gyrokinetic/apps/gk_species_fdot_multiplier.c
Maxwell-Rosen: Done

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: Added description

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Maxwell-Rosen: Removed

gyrokinetic/apps/gk_species_fdot_multiplier.c
Maxwell-Rosen: Refactored to use the separate advance_times_cfl method

manauref (approved):

JunoRavin (approved):

Issue #950, opened 2026-02-17
Bug: old initial conditions aren't loading after the GK metadata was restructured


Maxwell-Rosen:
In an input file, I'm loading an old initial condition before the metadata restructure
```
    .init_from_file = {
      .type = GKYL_IC_IMPORT_F,
      .file_name = "initial-condition/gk_lorentzian_mirror-ion_200.gkyl",
    },
```

However, I get the following GDB error
```
#1  0x000015553f000d0d in gkyl_msgpack_to_map_elem_list (mpack_in=mpack_in@entry=0x7fffffdac180, nvals=nvals@entry=2, elist=elist@entry=0x7fffffdacd30) at zero/util.c:706
#2  0x00001555436a42cb in gk_species_file_import_init (app=app@entry=0xafbc00, gks=gks@entry=0x11300c0, inp=...) at apps/gk_species.c:1082
#3  0x00001555436a6ad4 in gk_species_init (gk_app_inp=gk_app_inp@entry=0x7fffffdc81c0, app=app@entry=0xafbc00, gks=0x11300c0) at apps/gk_species.c:1531
#4  0x00001555436b8a9c in gkyl_gyrokinetic_app_new_solver (gk=gk@entry=0x7fffffdc81c0, app=app@entry=0xafbc00) at apps/gyrokinetic.c:679
#5  0x00001555436c42bc in gkyl_gyrokinetic_app_new (gk=gk@entry=0x7fffffdc81c0) at apps/gyrokinetic.c:775
#6  0x00000000004022e6 in main (argc=<optimized out>, argv=<optimized out>) at sim.c:934
```

There is something wrong with the way the metadata is handled in the distribution functions.
The error is in line 1082 here.
<img width="1972" height="1532" alt="Image" src="https://github.com/user-attachments/assets/38fdaa69-2046-4023-8856-fc499b412998" />

<img width="1284" height="1900" alt="Image" src="https://github.com/user-attachments/assets/9c931585-6afe-4de4-9d18-c54779e54558" />

This issue was also brought up by @SjoerdPR

Maxwell-Rosen:
I ran on my laptop the regression test `./build/gyrokinetic/creg/rt_gk_wham_1xIC_2x2v_p1` and this reproduces the segfault exactly. Issues like this would be prevented with CI that works.

PR #951, opened 2026-02-19
Fix reading string from metadata

manauref:
in response to issue #950


Maxwell-Rosen (approved):
I tested the regression test that was failing and now it passes

Antoinehoff (approved):

PR #954, opened 2026-02-23
Porting over Makefile experimentation into a separate branch to PR for testing

JunoRavin:
We now manually export all variables needed for the environment when running make because the use of the EXPORT_ALL_VARIABLES make rule seemed to add in the entries of the .d files so that make propagates down to sub-make processes when it needs to recompile edited files. This bug was affecting certain branches such as the https://github.com/ammarhakim/gkeyll/tree/kitp-hamil-vlasov and the https://github.com/ammarhakim/gkeyll/tree/fpo-john branches on GPUs because both of these branches have added a number of new kernels in the addition of functionality such as the refactored Vlasov solver and implementation of the full Fokker-Planck operator. 

I am porting these changes over from these branches to bring into main first in the hopes that we can catch anything we might have missed. @johnson452 @jmrodman and @VigneshKrsnaKumar have all built their branches with these Makefile fixes, with special thanks to @VigneshKrsnaKumar for helping me catch that the way the SQL CFLAGS were being exported now had an issue that was only present on CPU builds. 

I request folks to build this branch on all of our standard machines (graduate students have been included if they have the bandwidth, but so long as multiple people check machines like Perlmutter for the different projects like gyrokinetics vs. Vlasov, it's fine). 

Edit: Note that right now we are exporting these various environment variables by hand. There may be a way to directly utilize every environment variable set in the config.mak or create some sort of shell script to automatically export everything, but for now I'm just specifying things by hand.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,10 @@ endif
 
 # CUDA flags
 USING_NVCC =
-NVCC_FLAGS = 
+NVCC_FLAGS =
 CUDA_LIBS =
+# Default SQL flags (nvcc block below overrides this for GPU builds)
+SQL_CFLAGS ?= -fPIC -Wno-implicit-int-float-conversion
 ifeq ($(CC), nvcc)
 	USING_NVCC = yes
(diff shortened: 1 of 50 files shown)


Maxwell-Rosen (approved):
Looks like a great PR. It builds on my local Linux CPU laptop. Jimmy confirms that it builds on Mac CPU, Perlmutter/Stellar GPU, and WSL CPU. I'm happy to approve since it doesn't break anything. Valgrind works fine on my laptop too.

johnson452 (approved):
I tested this on CPU (Mac) and GPU (Perlmutter) a ran Vlasov checks and some regression tests and simulations. This branch also successfully merges into gr_triads (an extension of kip_hamil_vlasov).

Antoinehoff (approved):

PR #964, opened 2026-03-04
Fix for memory leak in the implicit source solver for fluids

kltrbrds:
Released an array that is allocated in the implicit collision source updater was never being deallocated, leading to memory issues.


JonathanGorard (approved):

JonathanGorard:
Sorry!

Issue #967, opened 2026-03-07
[DR] Adding ion-electron collisions in the Boltzmann electron model


manauref:
We'd like to improve the Boltzmann electron model by adding ion-electron collisions. There are two stages in which we could do this, stage 1 is a short-term solution, and stage 2 is the preferred longer term option that we don't have time for right now.

Note that we wish to make this possible with $\nu = \nu(x,t)$.

Stage 1
======

The simplest way to do this is to add an electron species to the boltz_elc input files and make two changes in the source code:

Change 1
----------
For ions to collide with electrons, presently the electrons need to have their collisions app (LBO or BGK) initialized. This is because, for example, ions will fetch $\nu_{ei}$ in order to compute $u_{\parallel ie}$ and $v_{t,ie}$. So we need to initialize the electron collisions app. But we don't want to apply the collision contribution to df/dt (we want df/dt to stay 0), so we will simply add an option in the collisions table like
```
.elc = {
  .collisions =  {
      .collision_id = GKYL_LBO_COLLISIONS,
      ...
      .not_in_dfdt = true,
    },
},
```
this new `not_in_dfdt` option (defaults to false) prevents the collision contribution from being added to `df/dt`.

Change 2
----------

Assuming the electron temperature stays constant, we still want the electron density to follow the ions' (because our boltz elc model assumes quasineutrality). So we'd like to set `n_e = n_i` at every step.

For this purpose we'll use work in the `fixed_fraction` branch (originally meant for running simulations with a fixed impurity fraction), which can rescale a species `s` to have the density of another species `r` times a fraction `g`, e.g. `n_s = g\cdot n_r`. It uses the following table in the electron species
```
.elc = {
  .scaling = {
    .type = GKYL_GK_SPECIES_SCALING_FIXED_FRACTION,
    .reference_species = {'ion'},
    .fixed_fraction = 1.0,
  },
},
```

Stage 2
======

Ideally we revert back to implementing the Boltzmann electron model how we did in the old `gkyl` (g2) code: we use a fluid electron species. This means not specifying `vdim` or the velocity grid parameters in the electron species table of the input file. This will follow changes in the neutral species to support fluid and kinetic neutrals. But we don't have time for this at the moment so we'll test stage 1 first.

manauref:
implemented in PR #974

Maxwell-Rosen:
Did we implement stage 2? I think not, so maybe we should keep this issue open or re-file the issue in a new form.

PR #974, opened 2026-03-13
Ion-electron collisions in Boltzmann model & option to run with fixed fraction

manauref:
(see issue #967 for more discussion and future plans)

In this work we add ion-electron collisions to the Boltzmann electron model. In order to support spatially and time-dependent collision frequency $\nu_{ie}=\nu_{ie}(x,t)$ we:

1. Need the electron density to be updated in space-time. This is accomplished by including an electron species in the Boltzmann electron input file, but without any terms, and adding a scaling operation:
```
.elc = {
  .lower = ...,
  .upper = ...,
  .cells = ...,

  .scaling = {
    .type = GKYL_GK_SPECIES_SCALING_BOLTZMANN,
  },
},
```
which simply sets, every time step, the electron density as
```math
n_e = n_{i,sheath} \exp\left[-e\left(\phi - \phi_{sheath}\right)/T_e\right]
```
assuming quasineutrality at the sheath entrance.

2. Need the electron collisions app to be initialized (because ions get $nu_{ei}$ from it), so we add the electron collision table but we add the option `do_not_add_to_dfdt` so that electrons are not updated by electron-ion collisions. That is, electron table in the input file would look like:
```
.elc = {
  .lower = ...,
  .upper = ...,
  .cells = ...,

  .scaling = {
    .type = GKYL_GK_SPECIES_SCALING_BOLTZMANN,
  },

  .collisions = {
    ...
    .do_not_add_to_dfdt = true,
  },
},
```

New fixed fraction
-------------------

We upgraded the previously existing `.recycle_react_scaling` operation that we implemented for scaling neutrals to simply `.scaling`, and so we can scale the electrons as mentioned before. In the process we also added an option to scale any species to a fixed fraction of another species. For example, we may want to run with a fixed fraction carbon species like
```
.carbon = {
  .scaling = {
    .type = GKYL_GK_SPECIES_SCALING_FIXED_FRACTION,
    .reference_species = 'ion',
    .fixed_fraction = 0.05,
  },
},
```
so that every time step it sets the carbon density to be 5% of the main `ion` density.

Tests
------
@Maxwell-Rosen ran mirror simulations with and w/o ion-elc collisions, and below we can see that they have the intended effect:
- Lower density because we increase the rate at which we scatter into the loss cone
- Lower temperature because beam ions are more quickly slowed down
image (13)
image (14)
image (15)

We also ran all the GK regression tests with the new runregression tool @JunoRavin implemented. They all pass, except for `rt_gk_multib_sheath_1x2v_p1`, but that should be unaffected by this work and I think there's an issue in the way that test is constructed.


Antoinehoff:
Can you please valgrind and compute sanitizer check this branch with a regression test that uses this new feature and another one that does not but use many others like `rt_gk_tcv_iwl_adapt_source_3x2v_p1.c`?

manauref:
@Antoinehoff the 3x2v TCV test is clean, and so are another couple of tests that would be affected by these changes.

The regression tests pass, as mentioned in the original message.

manauref:
also, as annoying as it may be to get more work, a reviewer is always right to ask whether a PR is valgrind clean. @Antoinehoff 's request prompted a couple of fixes

Maxwell-Rosen:
Does this branch pass the runregression testing suite?

manauref:
Yes, it passed reg tests using the new reg testing system @JunoRavin created.
And I memory checked both reg tests that don't use any of the new stuff and reg tests that use the new stuff, and they are clean.

ammarhakim (approved):
Ready to merge.

tnbernard (approved):

PR #979, opened 2026-03-19
Far-SOL gridding for IWL.

akashukla:
Change IWL bound setting to accommodate flux surfaces in an IWL SOL that have no turning point or strange shapes. This fixes a high recycling case where Mana was having problems generating  IWL grids.

Below I am plotting an LTX grid that was previously problematic on the purple and yellow flux surfaces, which have somewhat strange shapes (at least compared with what I originally assumed IWL would look like). Now it looks good. Very small change in the code.

image


ammarhakim (approved):
Ready to merge.

PR #980, opened 2026-03-19
Fixing a couple of bugs which can affect Braginskii and cold fluid simulations

JunoRavin:
The explicit source coupling not correctly using z momentum sources when including transport terms and the cold fluid explicit EM update not correctly updating the electric field (needed to use q/m instead of just q so mass factor is not folded into current accumulation).


JonathanGorard (approved):
Sorry for the delay! (And for the typos in the implementation). Looks great to me.

johnson452 (approved):

PR #982, opened 2026-03-23
Coupling with an AI model for sheath BC in GK simulations

Antoinehoff:
This PR follows DR #957 and introduces $\mu$-dependent sheath boundary conditions in the gyrokinetic solver using a neural network surrogate of the [GYRAZE code](https://github.com/alessandrogeraldini/GYRAZE).

It is now using the KANN Gkeyll interface developed in #1056. In particular, this branch demonstrates the use of KANN model inference on multi GPU in production simulations.

## Motivation

The standard conducting sheath BC in Gkeyll uses a constant cutting velocity $v_{\parallel\text{cut}}$ determined by the potential drop $\Delta\phi = \phi_\text{mpe} - \phi_w$ between the magnetic presheath entrance (MPE) and the wall:

$$\frac{1}{2} m_s v_{\parallel\text{cut}}^2 = q_s \Delta\phi$$

This ignores the effect of the particle magnetic moment $\mu$ on the absorption/reflection condition.

The main changes in this PR reside in a new `gkyl_array vcutsq` which stores the $v_{\parallel\text{cut}}(\mu)$ curve in the `gkyl_bc_sheath_gyrokinetic` structure. The evaluation of the sheath BC is now done in two steps:
1. update of `vcutsq`, if we don't use the NN model, it will just contain the conducting sheath constant value.
2. set the ghost values with the reflection algorithm according to `vcutsq`.

The inference of the NN model in step 1. above is done in three stages:
1. Fill the KANN input vector with all parameter triplet $\alpha,\gamma,\hat\phi$ evaluated on each perpendicular surface nodes.
1. Call `gkyl_kann_net_apply` storing the output in `kann_out`. The KANN Gkeyll interface (#1056) makes this call on CPU or GPU.
1. Convert back the nodal values produced by the NN to a DG representation in `vcutsq`. Note, this step requires to interpolate between the NN model hardcoded 20 points $\mu$ grid and the simulation $\mu$ grid. The grid points are expected to be output by the NN with an extension of the output layer so we do not have any hardcoded features now.

## New files

- ~**`gyrokinetic/ker/bc_sheath_gyrokinetic/bc_sheath_gyrokinetic_gyraze_surrogate.c`**~
~Auto-generated GYRAZE surrogate model (NN weights, SVM classifier, forward pass, interpolation, projection), from [gkeyll_sheath_ai](https://github.com/Antoinehoff/gyraze_surrogate_c_interface), the commit is indicated in the file.~ This is now replaced by the KANN Gkeyll interface.

- ~**`gyrokinetic/ker/bc_sheath_gyrokinetic/gkyl_bc_sheath_gyrokinetic_gyraze_surrogate.h`**~
~Public API for the surrogate.~ This is now replaced by the KANN Gkeyll interface.

- ~**`gyrokinetic/creg/rt_gk_tcv_sheath_surrogate_2x2v_p1.c`**~
~Regression test, TCV clopen IWL GK simulation with surrogate sheath BC.~ The original 2x2v tcv reg test is using the surrogate model now.

- **`gyrokinetic/data/nn_model/nn_model_sheath_bc_conv_MPE.kann`**
This is the NN model in a KANN readable format, generate using [a python routine](https://github.com/Antoinehoff/gyraze_surrogate_c_interface/blob/main/src_py/to_kann.py) to convert from pytorch. Nomenclature: "conv" indicates that the model was trained on converged region of GYRAZE, "MPE" means it is expecting values at the magnetic presheath entrance. One could think of simplifying this name...

### Modified files

- **`gyrokinetic/zero/gkyl_bc_sheath_gyrokinetic.h`**
New constructor parameters `use_surrogate`, `surrogate_model_path`, and `phase_grid`/`phase_global` (the latter two for the `vcutsq` diagnostic output). New public methods `gkyl_bc_sheath_gyrokinetic_update_vcutsq` (computes `vcutsq`), `gkyl_bc_sheath_gyrokinetic_set_vcutsq`, and `gkyl_bc_sheath_gyrokinetic_write_vcutsq` (writes `vcutsq` as a diagnostic).

- **`gyrokinetic/zero/gkyl_bc_sheath_gyrokinetic_priv.h`**
`gkyl_bc_sheath_gyrokinetic` struct now stores: `vcutsq` DG array over perpendicular config space $\times\,\mu$, its basis and range, `use_surrogate` flag, a `update_vcutsq` function pointer for switching between conducting sheath and NN inference, and the `gkyl_kn_vec` input/output buffers and `gkyl_kann_net` of the KANN interface.

- **`gyrokinetic/zero/bc_sheath_gyrokinetic.c`**
See git diff.

- **`gyrokinetic/zero/bc_sheath_gyrokinetic_cu.cu`**
See git diff.

- **`gyrokinetic/ker/bc_sheath_gyrokinetic/bc_sheath_gyrokinetic_ser_p1.c`**
Reflection kernels now take the `vcutsq` DG expansion and do not compute it on the fly anymore. New `bc_sheath_gyrokinetic_vcutsq_const_{lower,upper}_{1x2v,2x2v,3x2v}_ser_p1` kernels to compute the constant conducting sheath model. New `bc_sheath_gyrokinetic_vcutsq_surr_{lower,upper}_{1x2v,2x2v,3x2v}_ser_p1` kernels to compute the surrogate vcut.

- **`gyrokinetic/apps/gkyl_gyrokinetic.h`**
New `use_sheath_surrogate` boolean to activate surrogate. New `surrogate_model_path` to store the path to the `.kann` file.

- **`gyrokinetic/apps/gkyl_gyrokinetic_priv.h`**
Species struct gains `alloc_surr_aux_var`, `maxwell_mom`, `maxmom`, `dens_sheath`, `temp_sheath` for surrogate moment computation.

- **`gyrokinetic/apps/gk_species.c`** 
Allocates the Maxwellian moment calculator when the surrogate is enabled, computes density and temperature before each BC application, and calls `gkyl_bc_sheath_gyrokinetic_update_vcutsq` at both lower and upper sheath/IWL boundaries. Also writes the `vcutsq` diagnostic alongside the phase-space data.

- **`gyrokinetic/unit/ctest_bc_sheath_gyrokinetic.c`** Expanded with surrogate tests: verifies `vcutsq` evaluation and end-to-end BC application with the surrogate for 1x2v, 2x2v, and 3x2v on both CPU and GPU.

### Correcting factor approach

Unlike DR #957, we gave up on the $\mu$-dependent multiplying factor $\alpha(\mu)$ which would modify the cutting velocity:

$$v_{\parallel\text{cut}}^2(\mu) = \alpha(\mu) \,\frac{2\,q_s}{m_s}\,\Delta\phi$$

because of computing it requires to divide by $\Delta\phi$ that can be 0.

Instead, the `vcutsq` array stores the full cutting velocity as a DG expansion over perpendicular configuration space $\times\,\mu$.

### Safety assertions

- An `assert` enforces that the surrogate is only used for electrons (`q2Dm == -2e/m_e`), since ion reflection is not expected.
- The surrogate requires `vdim > 1` (i.e. a $\mu$ dimension must exist).

### User interface

Enabling the surrogate requires a single flag in the BC specification:
```c
.bc_z = {
  { .dir = 1, 
    .edge = GKYL_LOWER_EDGE, 
    .type = GKYL_BC_GK_SPECIES_IWL,
    .use_sheath_surrogate = true,
    .surrogate_model_path = "gyrokinetic/data/nn_model/nn_model_sheath_bc_conv_MPE.kann",
  },
  },
  { .dir = 1, 
    .edge = GKYL_UPPER_EDGE, 
    .type = GKYL_BC_GK_SPECIES_IWL, 
    .use_sheath_surrogate = true,
    .surrogate_model_path = "gyrokinetic/data/nn_model/nn_model_sheath_bc_conv_MPE.kann",
  },
},
```

### Unit test

The expanded unit test `gyrokinetic/unit/ctest_bc_sheath_gyrokinetic.c` validates:
- Standard BC application (without surrogate) for 1x2v, 2x2v, 3x2v at both edges, on CPU and GPU.
- Surrogate `vcutsq` evaluation and end-to-end BC application with the surrogate.

Below are outputs with `write_field = true`, showing the reflected distribution and the $v_\text{cut}(\mu)$ curve from the surrogate:

<table>
<tr>
<td>Unit test: reflected distribution with surrogate vcut(mu)</td>
<td>Unit test: vcutsq from surrogate</td>
</tr>
</table>

The $v_\text{cut}(\mu)$ curve is now provided by the GYRAZE surrogate rather than the ad-hoc function in #960.

### TCV 2x2v regression test

The new regression test `rt_gk_tcv_sheath_surrogate_2x2v_p1.c` runs a TCV clopen IWL GK simulation. At higher resolution (24 $\times$ 16 $\times$ 12 $\times$ 8) we obtain the following results with and without the surrogate:

<table>
<tr>
<th>Without surrogate</th>
<th>With surrogate</th>
</tr>
<tr>
<td>TCV 2x2v without surrogate</td>
<td>TCV 2x2v with surrogate</td>
</tr>
</table>

We can see that the electron distribution functions close to the magnetic presheath entrance ($z=\pm \pi$) is affected by the surrogate as expected. More electrons are absorbed with the surrogate which creates a large region of $f_e=0$ in the $v_\parallel > 0$ region for $z=-\pi$ and $v_\parallel < 0$ region for $z=\pi$. This also show that the surrogate is active in the simulation and is not always reaching the non converged state which would bypass it. The plots are zoomed in the region of interest. 

<table>
<tr>
<th>Without surrogate</th>
<th>With surrogate</th>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

The SOL potential is significantly higher with the surrogate. This is expected: FLR effects allow electrons to be absorbed at a gyroradius distance from the wall, lowering the effective potential barrier. The increased negative charge depletion raises the potential.

### TCV 3x2v production like turbulence simulation
We consider now *my favorite* 3x2v low cost production case ([see input file](https://github.com/user-attachments/files/26236297/aibc_tcv_3x2v.c)).

<table>
<tr>
<th>Without surrogate</th>
<th>With surrogate</th>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

We also check the distribution function at the magnetic presheath entrance for the 3x2v case. The effect of the surrogate is less visible on this snapshot. It is possible that the surrogate is often bypassed though an effect is observed in the potential and macroscopic quantities.

<table>
<tr>
<th>Without surrogate</th>
<th>With surrogate</th>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

#### Performance comparison
For production like run as the 3x2v simulation presented above, the computational cost is limited. Averaged over a 6h restart, we have a cost of forward Euler evaluation of:
- 54.7547 ms for CSBC (see [csbc_timing.txt](https://github.com/user-attachments/files/26312722/csbc_timing.txt))
- 56.4295 ms for AIBC (see [aibc_timing.txt](https://github.com/user-attachments/files/26312724/aibc_timing.txt))

## Known limitations

- **Small impact angle assumption**: The GYRAZE code (and hence the surrogate) assumes a small magnetic field impact angle with the wall. The surrogate should **not** be used for c onfigurations like LAPD where $\alpha \sim \pi/2$. A runtime check on the impact angle could be added.
- **Training domain**: The surrogate is trained on $\alpha \in [2°, 10°]$, $\gamma \in [0.5, 4]$, $\hat\phi \in [1, 10]$. Outside this domain, we still evaluate the surrogate but we don't know how physical it is.
- **Electron-only**: Currently restricted to electrons by assertion. There is no need of a surrogate for ions since they are mostly absorbed in the applications of interest, but this could be relaxed in the future if needed.


Antoinehoff:
Here are a few examples in 1x2v TCV simulation

image

image

We can see that the amount of reflected electrons at the magnetic presheath entrance is much bigger with the surrogate model.

image

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -219,6 +219,8 @@ struct gkyl_gyrokinetic_bc {
   struct gkyl_gyrokinetic_emission_inp emission; 
   bool write_diagnostics; // Whether to output diagnostics.
   int bidx; // Block index (for multiblock solver).
+  bool use_sheath_surrogate; // Whether to use surrogate sheath BC to determine vcut (only used if type is SHEATH).
+  const char *surrogate_model_path; // Path to the .kann surrogate model file.
```
manauref: I don't think you need a single-purpose boolean here. Use `type` by introducing a new bc_type like ...SHEATH_SURROGATE.

similarly rename `surrogate_model_path` to something like `aux_str` so that other models can likewise pass a string (see aux_profile, aux_ctx). By the way, I think, though not 100% sure that
```
char aux_str[128];
```
is (memory) safer than
```
const char aux_str*;
```
hence why we don't have any strings passed by addressed anywhere.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1039,7 +1039,13 @@ struct gk_species {
   struct gkyl_array *f_host; // Host copy for IO and initialization.
 
   struct gkyl_array *gyro_phi; // Gyroaveraged electrostatic potential.
-  
+
+  bool alloc_surr_aux_var; // To know whether surrogate auxiliary variables are allocated.
+  struct gkyl_gk_maxwellian_moments *maxwell_mom; // Maxwellian moment array to get temperature and density for surrogate BCs.
+  struct gkyl_array *maxmom; // Array for maxwellian moments.
```
manauref: Why add these to operators species when gk_species_moment can compute the Maxwellian moments?

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1039,7 +1039,13 @@ struct gk_species {
   struct gkyl_array *f_host; // Host copy for IO and initialization.
 
   struct gkyl_array *gyro_phi; // Gyroaveraged electrostatic potential.
-  
+
+  bool alloc_surr_aux_var; // To know whether surrogate auxiliary variables are allocated.
```
manauref: I think this boolean, and the way it is used in the app, is not needed.

If you define a SHEATH_SURROGATE enum option, each place in the code where you currently check `alloc_surr_aux_var` can move to a block belonging to that `SHEATH_SURROGATE` option

gyrokinetic/apps/gk_species.c
```diff
@@ -204,14 +205,24 @@ gk_species_apply_bc_dynamic(gkyl_gyrokinetic_app *app, const struct gk_species *
...
   for (int d=0; d<cdim; ++d) {
     if (species->bc_is_np[d]) {
 
       switch (species->lower_bc[d].type) {
         case GKYL_BC_GK_SPECIES_SHEATH:
-          gkyl_bc_sheath_gyrokinetic_advance(species->bc_sheath_lo, app->field->phi_smooth, 
-            app->field->phi_wall_lo, f, &app->local);
+          gkyl_bc_sheath_gyrokinetic_update_vcutsq(species->bc_sheath_lo, 
+            app->field->phi_smooth, app->field->phi_wall_lo, species->dens_sheath, species->temp_sheath, 
+            app->gk_geom->geo_int.bmag, app->gk_geom->geo_surf[cdim-1].bimpactangle, &app->local);
```
manauref: den_sheath and temp_sheath are not allocated for GKYL_BC_GK_SPECIES_SHEATH. Best to pass 0 here.

Also, since the loop is written in terms of `d`, best to use `app->gk_geom->geo_surf[d]`.

Somewhere else there should be (hopefully) an assert checking that the user didn't request sheath in a direction other than `cdim-1`

gyrokinetic/apps/gk_species.c
```diff
@@ -204,14 +205,24 @@ gk_species_apply_bc_dynamic(gkyl_gyrokinetic_app *app, const struct gk_species *
   int num_periodic_dir = species->num_periodic_dir, cdim = app->cdim;
   gkyl_comm_array_per_sync(species->comm, &species->local, &species->local_ext,
     num_periodic_dir, species->periodic_dirs, f); 
+
+  if (species->alloc_surr_aux_var) {
```
manauref: this bloc can move into a `GKYL_BC_GK_SPECIES_SHEATH_SURROGATE` option in the loop below

gyrokinetic/apps/gk_species.c
```diff
@@ -348,6 +394,13 @@ gk_species_write_dynamic(gkyl_gyrokinetic_app* app, struct gk_species *gks, doub
 
   app->stat.species_io_tm += gkyl_time_diff_now_sec(wst);
   app->stat.n_io += 1;
+
+  // Write out the sheath BC velocity cutoff (vcutsq).
+  int par_dir = app->cdim-1; // Sheath BC acts in the parallel direction.
+  if (gks->lower_bc[par_dir].type == GKYL_BC_GK_SPECIES_SHEATH)
```
manauref: make this dependent on `GKYL_BC_GK_SPECIES_SHEATH_SURROGATE`, not `GKYL_BC_GK_SPECIES_SHEATH`

unclear whether this should always be written out (instead of depending on some `.write_diagnostics = true` option, but you know better

gyrokinetic/apps/gk_species.c
```diff
@@ -316,6 +329,39 @@ gk_species_copy_range_static(struct gkyl_array *out,
...
+    { .key = "basis_type", .elem_type = GKYL_MP_STRING, .cval = "serendipity" },
+    { .key = "Description", .elem_type = GKYL_MP_STRING,
+      .cval = "Square of the velocity cutoff used in the gyrokinetic sheath BC." }
+  };
+  int mpe_vcutsq_len = sizeof(mpe_vcutsq)/sizeof(mpe_vcutsq[0]);
+  int io_meta_len[] = {app->io_meta_dg_len, mpe_vcutsq_len};
+  const struct gkyl_msgpack_map_elem* io_meta[] = {app->io_meta_dg, mpe_vcutsq};
+  struct gkyl_msgpack_data *mt = gkyl_msgpack_create_union(sizeof(io_meta_len)/sizeof(int), io_meta_len, io_meta);
+
+  const char *fmt = "%s-%s_vcutsq_%s_%d.gkyl";
```
manauref: 1. please change the format to follow some other boundary bc files, i.e.
`<sim_name>-<species>_bc_<dir><edge>_vcutsq_<frame>.gkyl`
where <dir> is x,y, or z and <edge> is lower or upper.

I'd like to change x,y,z to 0,1,2 instead so there's less confusion for 2x2v. But if you use either is fine.

2. Can you add this quantity to 
https://www.overleaf.com/read/zxxncdvcrxwz#534f3a
please?

gyrokinetic/apps/gk_species.c
```diff
@@ -316,6 +329,39 @@ gk_species_copy_range_static(struct gkyl_array *out,
...
+// Write out the sheath BC velocity cutoff (vcutsq) at one boundary.
+static void
+gk_species_write_vcutsq(gkyl_gyrokinetic_app* app, struct gk_species *gks,
+  struct gkyl_bc_sheath_gyrokinetic *bc_sheath, const char *edge_str, double tm, int frame)
+{
+  struct timespec wst = gkyl_wall_clock();
+
+  // Package metadata (time/frame + DG basis info).
+  gkyl_msgpack_map_elem_set_double(app->io_meta_dg_len, app->io_meta_dg, "time", tm);
+  gkyl_msgpack_map_elem_set_uint(app->io_meta_dg_len, app->io_meta_dg, "frame", frame);
```
manauref, line 347: I think but i'm not 100% that this should use `io_meta_basic` instead of `io_meta_dg`, because the latter has the conf-space basis and poly order, which is not what you want here.

gyrokinetic/creg/rt_gk_ltx_1x2v_p1.c
```diff
@@ -128,6 +128,16 @@ double integrand(double t, void *int_ctx)
   struct gk_app_ctx *app = inctx->app_ctx;
   return Jr(r,t,app) / pow(R_rtheta(r,t,app),2);
 }
+
+double Bphi(double R, void *ctx)
```
manauref, line 132: i'd like to merge the PR fixing this file first before we accept this

gyrokinetic/creg/rt_gk_tcv_iwl_1x2v_p1.c
```diff
@@ -0,0 +1,688 @@
...
+  double floor_src;
+  // Grid parameters
+  double x;
+  double Lz;
+  double z_min, z_max;
+  int num_cell_z, num_cell_vpar, num_cell_mu;
+  int cells[GKYL_MAX_DIM], poly_order;
+  double vpar_max_elc, mu_max_elc, vpar_max_ion, mu_max_ion;
+  // Simulation control parameters
+  double final_time, write_phase_freq;
```
manauref: please rename 
num_cell_ -> N
final_time -> t_end
because most input files follow this pattern, and it's useful to have all input files have these be the same (sometimes I edit them all in mass)

gyrokinetic/ker/bc_sheath_gyrokinetic/gkyl_bc_sheath_gyrokinetic_kernels.h
```diff
@@ -10,16 +12,57 @@ static inline double invL(double x) {
   return (3.*x-x*x*x*(6. + x*x - 2.*x*x*x*x)/5.)/(1.-x*x); 
 }
 
+// Linear interpolation of the surrogate output vcut[ng] (defined on the mu-grid
```
manauref, line 15: This header file is Maxima-generated. Either add this to the maxima or create a separate header just for this and include it in the surrogate kernels.

gyrokinetic/zero/bc_sheath_gyrokinetic.c
```diff
@@ -1,12 +1,194 @@
...
+
+    up->kernels->vcutsq_const(phi_p, phi_wall_p, up->q2Dm, vcutsq_p);
+  }
+}
 struct gkyl_bc_sheath_gyrokinetic*
 gkyl_bc_sheath_gyrokinetic_new(int dir, enum gkyl_edge_loc edge, const struct gkyl_basis *basis,
   const struct gkyl_range *skin_r, const struct gkyl_range *ghost_r, const struct gkyl_velocity_map *vel_map,
-  int cdim, double q2Dm, bool use_gpu)
+  int cdim, double q2Dm, bool use_surrogate, const char *surrogate_model_path,
+  const struct gkyl_rect_grid *phase_grid, const struct gkyl_range *phase_global, bool use_gpu)
```
manauref: you are passing phase_global? that range is almost never used. Do you mean local_ext?

gyrokinetic/zero/gkyl_bc_sheath_gyrokinetic.h
```diff
@@ -20,28 +25,77 @@ typedef struct gkyl_bc_sheath_gyrokinetic gkyl_bc_sheath_gyrokinetic;
  * @param vel_map Velocity space mapping object.
  * @param cdim Configuration space dimensions.
  * @param q2Dm charge-to-mass ratio times 2.
+ * @param use_surrogate Boolean to indicate whether to use surrogate model for vcut.
```
manauref: just pass the bc_type enum instead, after you've created the SHEATH_SURROGATE option

gyrokinetic/zero/gkyl_bc_sheath_gyrokinetic.h
```diff
@@ -20,28 +25,77 @@ typedef struct gkyl_bc_sheath_gyrokinetic gkyl_bc_sheath_gyrokinetic;
...
+  struct gkyl_msgpack_data *meta, const char *fname);
+
+/**
+ * Get the pointer to the KANN surrogate model used in the sheath BC updater.
+ * Needs to be released by the caller using gkyl_kann_net_release after use.
+ * 
+ * @param up BC updater.
+ * @return Pointer to KANN model.
+ */
+struct gkyl_kann_net *gkyl_bc_sheath_gyrokinetic_acquire_model(struct gkyl_bc_sheath_gyrokinetic *up);
```
manauref: rename _acquire_kann or _acquire_kann_model

gyrokinetic/zero/bc_sheath_gyrokinetic.c
```diff
@@ -15,48 +197,133 @@ gkyl_bc_sheath_gyrokinetic_new(int dir, enum gkyl_edge_loc edge, const struct gk
...
+  up->vcutsq_write = false;
+  up->vcutsq_ho = NULL;
 
-  // Choose the kernel that does the reflection/no reflection/partial
-  // reflection.
+  // Get polynomial order
+  int poly_order;
+  if (use_gpu) {
+    struct gkyl_basis *basis_ho;
+    basis_ho = gkyl_malloc(sizeof(struct gkyl_basis));
```
manauref: merge lines

gyrokinetic/zero/bc_sheath_gyrokinetic.c
```diff
@@ -15,48 +197,133 @@ gkyl_bc_sheath_gyrokinetic_new(int dir, enum gkyl_edge_loc edge, const struct gk
...
+    gkyl_cu_memcpy(basis_ho, basis, sizeof(struct gkyl_basis), GKYL_CU_MEMCPY_D2H);
+    poly_order = basis_ho->poly_order;
+    gkyl_free(basis_ho);
+  } else {
+    poly_order = basis->poly_order;
+  }
```
manauref: this is a lot of juggling just to get poly_order of the vcut discretization. Just make vcutsq_poly_order an input to the updater, like we have a poly_order input in bc_twistshift. In both cases one could use the same order as f to define these curves, but you could also use a higher order representation.

manauref (changes requested):
Very nice general description. It covers all the basis and offers useful plots. I also very much like that this is now using the kann code.

Here's some feedback on what I've had time to review. I'll try to look at the actual kann part later today/tomorrow. In addition to the minor things below, here are some higher level questions:
1. One downside I see here is that another person (maybe even yourself) could not extend the surrogate because we don't have enough material to replicate the construction of the network. Would it be possible to create a directory (say in gyrokinetic/data/) with the GYRAZE input files and network training scripts? I see this as necessary, much in the same way that we have a folder for creating the parameters of the radiation operator (which depend on other external analysis, but which can be replicated especially with what's in PR #1060).
2. I'd like to read your AI sheath paper draft before we merge this to better understand what this is contributing and how it's changing the physics (beyond what you've put in the description).
3. The description mentions that this model doesn't work for ion reflection. As you know Gkeyll can produce transients with negative MPE potential. What does the surrogate do to electrons and ions in that case? Are the input files only giving this option to the electrons?
4. We should add the assert that you mentioned for alpha=pi/2. Also, since the validity of using this outside of the training domain is unknown, I think we should add an `fprintf(stdout,` with a warning like "sheath_surrogate: outside of region of convergence". In general we stand by what Gkeyll produces, meaning we say its results can be trusted (at least as far as we know). But this option is adding the ability to produce results whose validity is knowningly unknown.
5. Is the gkylcas PR up to date?

gyrokinetic/apps/gk_species.c
```diff
@@ -204,14 +205,24 @@ gk_species_apply_bc_dynamic(gkyl_gyrokinetic_app *app, const struct gk_species *
...
     if (species->bc_is_np[d]) {
 
       switch (species->lower_bc[d].type) {
         case GKYL_BC_GK_SPECIES_SHEATH:
-          gkyl_bc_sheath_gyrokinetic_advance(species->bc_sheath_lo, app->field->phi_smooth, 
-            app->field->phi_wall_lo, f, &app->local);
+          gkyl_bc_sheath_gyrokinetic_update_vcutsq(species->bc_sheath_lo, 
+            app->field->phi_smooth, app->field->phi_wall_lo, species->dens_sheath, species->temp_sheath, 
+            app->gk_geom->geo_int.bmag, app->gk_geom->geo_surf[cdim-1].bimpactangle, &app->local);
+          gkyl_bc_sheath_gyrokinetic_advance(species->bc_sheath_lo, f, &app->local);
```
manauref: I realized two more things:
1. You don't need to create separate dens and temp arrays. Just pass the array with the Maxwellian moments to the updater, and in the while loop where you fetch pointers (gkyl_array_cfetch) you can separately fetch the dens and temp components.
2. `gkyl_bc_sheath_gyrokinetic_update_vcutsq` can simply be called as the first thing inside `gkyl_bc_sheath_gyrokinetic_advance`. I don't see a the reason to keep it as a separate call in the app.

manauref (changes requested):

gyrokinetic/apps/gkyl_gyrokinetic.h
Antoinehoff: I will try for the `GKYL_BC_GK_SPECIES_SHEATH_SURROGATE` but I don't know if it makes a lot of sense to have `GKYL_BC_GK_SPECIES_SHEATH` and then `GKYL_BC_GK_SPECIES_SHEATH_SURROGATE` because this looks like we have the "true" sheath or a surrogate of the sheath. Maybe we should have `GKYL_BC_GK_SPECIES_SHEATH_CONDUCTING` and `GKYL_BC_GK_SPECIES_SHEATH_SURROGATE`.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Antoinehoff: "two operators"? I see only one, `struct gkyl_gk_maxwellian_moments *maxwell_mom;`, which I added because `struct gk_species_moment *moms;` is a diagnostic moment so I assumed that this can be deactivated depending on which diagnostics we are asking. I want to be sure that we always allocate what is necessary to compute the maxwellian moments when the sheath surrogate model is used.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Antoinehoff: ok I will remove it and see how it looks with `SHEATH_SURROGATE` option.

gyrokinetic/creg/rt_gk_ltx_1x2v_p1.c
Antoinehoff: Just sent a reminder to the other reviewers of PR #1073

gyrokinetic/creg/rt_gk_tcv_iwl_1x2v_p1.c
Antoinehoff: yeah I saw you were enforcing that. Personally, I prefer num_cell_x than Nx because it's more explicit but as you started to make everything Nx I will follow this standard, which can be discussed in another PR.

gyrokinetic/ker/bc_sheath_gyrokinetic/gkyl_bc_sheath_gyrokinetic_kernels.h
Antoinehoff: it is added to the maxima.

gyrokinetic/zero/bc_sheath_gyrokinetic.c
Antoinehoff: yeah should be just local. It was for diagnostics so I thought for a moment we needed the entire range but this is only for the cutting velocity curve diagnostic.

gyrokinetic/zero/gkyl_bc_sheath_gyrokinetic.h
Antoinehoff: ok

gyrokinetic/zero/gkyl_bc_sheath_gyrokinetic.h
Antoinehoff: this is not used anymore. I remove it.

gyrokinetic/apps/gk_species.c
Antoinehoff: will try something

gyrokinetic/apps/gk_species.c
Antoinehoff: I will merge the update vcutsq and the advance routine ok.

gyrokinetic/apps/gkyl_gyrokinetic.h
manauref: i do think
`GKYL_BC_GK_SPECIES_SHEATH_CONDUCTING`
`GKYL_BC_GK_SPECIES_SHEATH_SURROGATE`
is a good choice, as in the future we may implement other sheath models (e.g. logical). So feel free to rename `GKYL_BC_GK_SPECIES_SHEATH` -> `GKYL_BC_GK_SPECIES_SHEATH_CONDUCTING`

Antoinehoff: good :)

gyrokinetic/apps/gk_species.c
Antoinehoff: the vcut_sq is stored in a DG modal representation on a p=1 3D (x,y,mu) serendipity basis. It does not use the gk hybrid one so I think this works like that. I checked the resulting .gkyl output and it matches with what should the cutting velocity look like.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Antoinehoff: Ok I got it now

gyrokinetic/apps/gk_species.c
Antoinehoff: Done

gyrokinetic/unit/ctest_bc_sheath_gyrokinetic.c
```diff
@@ -723,217 +962,302 @@ test_bc_sheath_gyrokinetic_3x2v(const int *cells, enum gkyl_edge_loc edge,
...
+    .phi_mpe = 15.0,
+    .phi_wall = 0.0,
+    .impact_angle = 5.0 * GKYL_PI / 180.0, 
+    .x0 = 0.0,
+    .sigmax = 1.0,
+    .y0 = 0.0,
+    .sigmay = 1.0,
+    .z0 = 0.0,
+    .sigmaz = 1.0,
+    .use_surrogate = true,
```
manauref: you are hardcoding this, so the non_surrogate test is not tested?

make this a parameter to this function, and run each unit test with this set to true or false

manauref (changes requested):
please answer my comments/questions from my last review, and see small change below.

Aside from that, this is ready for approval, just: A) resolve conflicts, B) run CI

gyrokinetic/unit/ctest_bc_sheath_gyrokinetic.c
Antoinehoff: Ok but I don't think this it is necessary to run every combination. Testing the NN in one case is fine and we don't want to make the unit test too long.

PR #987, opened 2026-04-01
Add the impact angle of the magnetic field on the e_1 e_2 plane

Antoinehoff:
This PR implements the impact angle of the magnetic field vector on the $e_1$ $e_2$ plane. This new feature allows us, e.g., to inquire the angle between the magnetic field and the divertor plate at the magnetic presheath entrance.

#### Derivation of the formula

We introduce:
- $\mathbf{b}$: unit vector of the magnetic field
- $\mathbf{n}$: unit normal vector to the plate
- $\alpha$: angle between the magnetic field and the plate
- $\beta$: angle between the magnetic field and the normal to the plate

By definition,

$$\alpha + \beta = \frac{\pi}{2},$$

$$\mathbf{b} \cdot \mathbf{n} = \cos\beta,$$

which yields

$$\sin\alpha = \mathbf{b} \cdot \mathbf{n}.$$

Following [D'haeseleer et al. (1991)](https://doi.org/10.1007/978-3-642-75595-8) Fig. 2.2, we write $\mathbf{b} = \mathbf{e}_3 / |\mathbf{e}_3|$ and $\mathbf{n} = \mathbf{e}^3 / |\mathbf{e}^3|$, where $\mathbf{e}^3 = \nabla\psi \times \nabla\theta$ is the contravariant basis vector along the magnetic field direction, $\mathbf{e}_3$ is the covariant basis vector along the plate normal, $\psi$ is the poloidal flux, and $\theta$ is the poloidal angle. This gives

$$\sin\alpha = \frac{\mathbf{e}^3 \cdot \mathbf{e}_3}{|\mathbf{e}^3||\mathbf{e}_3|} = \frac{1}{|\mathbf{e}^3||\mathbf{e}_3|}.$$

Finally, using the metric definitions of the contravariant and covariant basis vectors,

$$\boxed{\sin\alpha = \frac{1}{\sqrt{g^{33} g_{33}}}}.$$


manauref (changes requested):
This is useful.
@Antoinehoff , given that adding new geo quantities is often prone to memory error, did you valgrind/compute-sanitizer check?

another thing: how about we write this out in _write_geo? it could be used in post-processing, e.g. to compute the correction factor to the heat flux due to oblique incidence.

Antoinehoff:
The branch is valgrind clean now (there were a missing `gkyl_array_release` indeed @manauref ). I've also added the impact angle as a new geometric output. Here how it looks like for the TCV 3x2v reg test:

```
pgkyl rt_gk_tcv_iwl_adapt_source_3x2v_p1-bimpactangle.gkyl interpolate -p 1 sel --z0 0 pl 
```
yields

image

and

```
pgkyl rt_gk_tcv_iwl_adapt_source_3x2v_p1-bimpactangle.gkyl interpolate -p 1 sel --z1 0 pl 
```

yields

image

manauref:
this branch should be basically ready, but need to resolve conflicts

Antoinehoff:
> this branch should be basically ready, but need to resolve conflicts

Done !

manauref (approved):

Maxwell-Rosen (approved):

PR #988, opened 2026-04-03
Add capability of loading the field and distribution function moments from files

Maxwell-Rosen:
I’m using this in kinetic electron simulations of magnetic mirrors, initialized from a Boltzmann electron simulation. For Boltzmann electrons, $n_i = n_e$, so the electrons should be initialized to $n_i$ in the kinetic electron simulation. Furthermore, we need to rescale the ions to account for the polarization density, so we need to read the field from that simulation as well. For symmetry, I added the options of loading $U_\parallel , T, T_\parallel , T_\perp$ from files as well.

This PR is nearly 100% github copilot, with strong steering from me.

Here is an example simulation on 2 GPUs using this loading

Density
image

Field
image


manauref (changes requested):
This is useful. I too have used it when it was implemented in another branch but it never made it to main.

I don't think a new import struct is necessary. Each of those quantities (upar, etc), can come from a `gkyl_gyrokinetic_ic_import` struct. This struct has enough flexibility.

I think you should add support for reading in MaxwellianMoments and BiMaxwellian moments, because that's what a lot of people have.

As it is with LLMs, I think this is probably more verbose and has more code duplication than it needs to be. e.g.
- `header_from_file` is either duplicated from another file or unnecessary. 
- `gk_field_polarization_potential_from_file_new` also seems suspicious. We used to load in polarization potential already, why is this needed here/again?

Regarding names:
- the `_from_file` quantities (e.g. upar_from_file) should instead be `_import`, as they point to an _ic_import struct.
- use `_from_file` instead of `is_imported`

More importantly:
Don't ever submit LLM-written code again without saying how you tested it, and without submitting (unit/regression) tests that test it. Continuing to do so can get you kicked out of this repo.

Maxwell-Rosen:
I'll address many points.

I removed the import struct. Good catch.

I added support for loading the moments directly. An additional flag could be used to load the Maxwellian moments, but only import the upar, for instance, or just T. So, loading an individual moment from the agglomerated file. This would require additional modifications. This is left to future work.

`header_from_file` in gk_field.c was not duplicated. I just moved it around. This was already there.

`gk_field_polarization_potential_from_file_new` is kind of the whole point of this PR. This quantity was not loaded from files. The only supported option was loading it from a user-defined function.

File names updated.

This code WAS tested. That's the plot above. The plot in the original PR was a showcase that the imported moments are correct. To expand the testing, I added a new regression test that imports various initial conditions from files. The results must be postprocessed with pgkyl, and the results are shown below. Any errors are due to our projection routines, not the loading mechanism. Several of these moments are initialized to zero, which is why there are errors. Temperatures are machine-precision zero error. Upar is just initialized to zero (m/s), so there are errors. This regression test is valgrind clean.


image

image

image

image

image

manauref:
use of formatter or LLMs introduced a bunch of changes that:
- i do not trust
- don't want to review
- are not needed for what you are trying to add

please undo and open again.
I think the formatter shouldn't be used if it's going to do this

gyrokinetic/apps/gk_species_projection.c
```diff
@@ -140,83 +187,147 @@ static void
...
+    proj->dens_from_file = true;
+    proj->upar_from_file = true;
+    proj->temppar_from_file = true;
+    proj->tempperp_from_file = true;
+  }
+
+  // prim_moms_host is a combined array of all the projected moments
+  int prim_moms_ncomp = proj->proj_id == GKYL_PROJ_MAXWELLIAN_PRIM ? 3 : 4;
+  proj->prim_moms_host = mkarr(false, prim_moms_ncomp*app->basis.num_basis, app->local_ext.volume);
+  proj->prim_moms = mkarr(app->use_gpu, prim_moms_ncomp*app->basis.num_basis, app->local_ext.volume);
```
manauref: remember one allocates the non-host array, then the _host array should get its size from the non-host array

gyrokinetic/apps/gk_species_projection.c
```diff
@@ -140,83 +187,147 @@ static void
...
+    proj->dens_from_file = true;
+    proj->upar_from_file = true;
+    proj->temp_from_file = true;
+  }
+  else if (bimaxwellian_moms_from_file) {
+    proj->dens_from_file = true;
+    proj->upar_from_file = true;
+    proj->temppar_from_file = true;
+    proj->tempperp_from_file = true;
+  }
```
manauref, line 245: in these, isn't it shorter to just write
proj->dens_from_file = maxwellian_moms_from_file || bimaxwellian_moms_from_file
?

gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -43,6 +43,15 @@ struct gkyl_gyrokinetic_projection {
...
       void *ctx_tempperp;
 
+      // Optionally read primitive moments from files.
+      const struct gkyl_gyrokinetic_ic_import *maxwellian_moms_import;
+      const struct gkyl_gyrokinetic_ic_import *bimaxwellian_moms_import;
+      const struct gkyl_gyrokinetic_ic_import *density_import;
+      const struct gkyl_gyrokinetic_ic_import *upar_import;
+      const struct gkyl_gyrokinetic_ic_import *temp_import;
+      const struct gkyl_gyrokinetic_ic_import *temppar_import;
+      const struct gkyl_gyrokinetic_ic_import *tempperp_import;
```
manauref: don't think these need to be const, but i'm not sure

Maxwell-Rosen:
I undid the formatter.

If you’re complaining that the formatter changed a bunch of lines, that’s because those lines weren’t formatted according to our coding standards. It's fixing all the inconsistencies and mistakes. There were several braces in the wrong places, astrisks in the wrong places, trailing spaces, lines of code that are too long, etc. which are reasonable to change. None of it affected functionality and improved readability. However, it did increase the diff, which is why I reverted it to expedite this review. A formatter is a good thing to have for our codebase as it ensures consistent styles across the codebase, improving readability.

gyrokinetic/apps/gk_species_projection.c
```diff
@@ -89,12 +121,27 @@ static void
...
-  gkyl_proj_on_basis_advance(proj->proj_temppar, tm, &app->local, proj->vtsqpar);
-  gkyl_proj_on_basis_advance(proj->proj_tempperp, tm, &app->local, proj->vtsqperp);
-  gkyl_array_scale(proj->vtsqpar, 1.0/s->info.mass);
-  gkyl_array_scale(proj->vtsqperp, 1.0/s->info.mass);
+  if (proj->bimaxwellian_moms_from_file) {
+    gkyl_array_copy(proj->prim_moms, proj->prim_moms_host);
+    gkyl_gk_maxwellian_proj_on_basis_advance(proj->proj_max,
+      &s->local, &app->local, proj->prim_moms, false, f);
+    return;
+  }
```
manauref, line 166: i don't think it should use this logic, of returning for this case.
It's best to farther down have a logic like

if X {
  project moments
}

project f

manauref (changes requested):
2 more things:
1. there's some repetition in several places that could be eliminated by using a function
for example
```
f (!proj->dens_from_file) {
    gkyl_proj_on_basis_advance(proj->proj_dens, tm, &app->local, proj->dens);
  }
  if (!proj->upar_from_file) {
    gkyl_proj_on_basis_advance(proj->proj_upar, tm, &app->local, proj->upar);
  }
  if (!proj->temppar_from_file) {
    gkyl_proj_on_basis_advance(proj->proj_temppar, tm, &app->local, proj->vtsqpar);
    gkyl_array_scale(proj->vtsqpar, 1.0/s->info.mass);
  }
  if (!proj->tempperp_from_file) {
    gkyl_proj_on_basis_advance(proj->proj_tempperp, tm, &app->local, proj->vtsqperp);
    gkyl_array_scale(proj->vtsqperp, 1.0/s->info.mass);
  }
```
could instead be 4 lines, each calling a little function that takes the boolean, the proj operator, time, range, array and a multiplier.

2. please runregression and ensure GK tests pass

gyrokinetic/apps/gk_species_projection.c
Maxwell-Rosen: This is how it was before, but I'll update it. This line was just moved up

gyrokinetic/apps/gk_species_projection.c
Maxwell-Rosen: Sure, but the temperatures still need this if statement. It's the same amount of code either way, but the current syntax is very clear, while I find the inline if statements difficult to read

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: Const is okay and good here to ensure that the structs are not changed anywhere else in the code. They shouldn't be changed

gyrokinetic/apps/gk_species_projection.c
Maxwell-Rosen: Doing the return at the beginning of the function reduces the number of nested if statements, reducing the layering and improving readability. We shouldn’t have one giant if statement over the whole function.

Maxwell-Rosen:
This branch does pass the runregression test. 

There is an issue with this regression test, which seems independent of this PR
,/build/gyrokinetic/creg/rt_gk_ltx_iwl_2x2v_p1 
calc_metric.c: Orthonormality violated : e_1 . e^1 = -1.000000e+00
rt_gk_ltx_iwl_2x2v_p1: zero/calc_metric.c:90: check_orthonormality: Assertion `!exit_at_check' failed.
Aborted

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: These should be passed as value, not as pointer

manauref: Actually i remember why this is a problem:
you can't hold this address. The input file needs to pass it by value, otherwise the thing the app is referencing (which is managed by the input file) may be deleted.

manauref (changes requested):

gyrokinetic/apps/gk_species_projection.c
```diff
@@ -65,14 +84,64 @@ gk_species_projection_calc_proj_func(gkyl_gyrokinetic_app *app, struct gk_specie
...
 {
-  gkyl_proj_on_basis_advance(proj->proj_dens, tm, &app->local, proj->dens); 
-  gkyl_proj_on_basis_advance(proj->proj_upar, tm, &app->local, proj->upar);
-  gkyl_proj_on_basis_advance(proj->proj_temp, tm, &app->local, proj->vtsq);
-  gkyl_array_scale(proj->vtsq, 1.0/s->info.mass);
+  if (proj->maxwellian_moms_from_file) {
+    gkyl_array_copy(proj->prim_moms, proj->prim_moms_host);
+    gkyl_gk_maxwellian_proj_on_basis_advance(proj->proj_max,
+      &s->local, &app->local, proj->prim_moms, false, f);
+    return;
```
Maxwell-Rosen, line 136: Change this to if {
{
else {
}

Maxwell-Rosen: This was a note from a meeting with Mana, but I disagree with the sentiment. Early returns improve code readability and maintainability. It's a common debate, but the community seems to agree that early returns are good.

https://stackoverflow.com/questions/4369822/early-returns-vs-nested-positive-if-statements

https://dev.to/nazirul_amin/code-refactoring-avoid-nested-if-statements-with-early-returns-52ml

manauref (approved):

JunoRavin (approved):
Functionality-wise, I think this is great and I approve. Just to document an internal thought I have: the logic chain on checking whether or not we need to project a moment is interesting. You have to store internally whether certain moments need to be projected due to differences in e.g., reading Maxwellian vs. bi-Maxwellian moments. This makes sense but introduces a somewhat non-standard and ultimately unsafe memory access pattern to the code, one in which a pointer to a pointer is passed to a function and then that function allocates the object. 

You do your due diligence with booleans and checking whether something exists but I'd like to put it to the group that we should have in the back of our mind that this is a feature extension we should consider refactoring at some point. This functionality is obviously new and unique in that you are reading in specific moments while projecting others (if I understand the use case where you want to read in things like ni, but there will be other functions which construct the electron temperature). How much can this method of constructing initial conditions be thought of hierarchally, in the same way folks like @manauref and others have completely restructure gk_species.c to carefully handle a hierarchy of allocations (a gk_species_collisionless struct for example, that has also laid the foundation for a more modular extension to things like electromagnetics that @Antoinehoff is doing)? 

I should also myself refresh my knowledge of Rust to see if there is anything we can borrow in the same way we've effectively borrowed the acquire/release formalism from Rust for our pointer discipline, as I just get nervous passing pointers of pointers to functions so those functions can allocate objects (our traditional way of approaching this is to define hierarchal structs that may do internal allocations, and things are released in a hierarachal fashion).

manauref:
I think I understand what you mean @JunoRavin . For clarity, can you cite the line number were you are seeing less than ideal pointer passing? It'll help refactor in the future.

I did say to @Maxwell-Rosen that I could envision some ways to re-structure the projection and import structs, but I didn't want to drag him into that refactor at this point as he has other valuables PRs to get to and is doing good work pushing on the physics front.

JunoRavin:
@manauref Yea, I am thinking of this line here: https://github.com/ammarhakim/gkeyll/blob/1deded500db8f6329dd390fd8a83076e328881a6/gyrokinetic/apps/gk_species_projection.c#L114

where we pass in the pointer to the pointer of the proj_on_basis struct, and then allocate the struct inside the function.

PR #989, opened 2026-04-04
Fix github actions mac build, valgrind sanitize, all unit tests passing

### LINKED ISSUE #116: Failing Linux Action & Other Github Action issues
The Linux build is failing as the OpenBLAS build fails to detect the CPU type of the Linux instance. This seems like a Github issue, but perhaps one could try and install OpenBLAS from ubuntu apt get. Not sure how to fix.

Also, seems that some unit tests fail (in particular the ideal MHD ones) but the Actions still "pass". Need to figure out why this is the case.

Maxwell-Rosen:
# Documentation Changes

# Purpose

The current Mac CI build fails intermittently and inconsistently. I've made some changes, modularizing the build and ensuring that we use 3 processes.

Macos-latest has 3 CPU cores
https://docs.github.com/en/actions/reference/runners/github-hosted-runners

Unit testing and Valgrind were checked to ensure all unit tests pass. Those that were failing and I could not fix simply are not included.

I fixed a bug in make valcheck, where the command was not being executed from the root directory and was executed in the submodules, so certain unit tests were failing because they didn't know where the files to read were.

# Affected Areas

`.github/workflows` and files located within for modifying the CI.

Failing unit tests are commented out so we can establish a baseline. I am not going to spend time doing the hard work of fixing the unit test or identifying why it is failing. That work is left to the individuals to whom these unit tests are relevant. There is an open issue about this (https://github.com/ammarhakim/gkeyll/issues/845)

Failing unit tests are:
- gyrokinetic/unit/ctest_gyrokinetic_cross_prim_moms_bgk.c `test_1x2v_p1`
- gyrokinetic/unit/ctest_mom_gyrokinetic.c `test_2x2v_p1` `test_2x2v_p1_cu`
- gyrokinetic/unit/ctest_rescale_ghost_jacf.c `test_1x1v_ho` `test_2x2v_ho` and device versions
- gyrokinetic/unit/ctest_dg_gyrokinetic_kern_tm.c -- all tests removed and fail.
- gyrokinetic/unit/ctest_dg_interpolate.c -- all tests related to `_gk_` fail.
- gyrokinetic/unit/ctest_dg_rad_gyrokinetic.c -- All tests under `_Li1` have erroneous print statements and warnings that they are not set up correctly.
- moments/unit/ctest_gr_spacetime.c `test_gr_schwarzschild` `test_gr_kerr`
-  moments/unit/ctest_wv_gr_mhd.c `test_gr_mhd_waves_schwarzschild`
- moments/unit/ctest_wv_gr_mhd_tetrad.c `test_gr_mhd_tetrad_waves_schwarzschild`
- moments/unit/ctest_wv_gr_ultra_rel_euler.c `test_gr_ultra_rel_euler_waves_schwarzschild`
- moments/unit/ctest_wv_gr_ultra_rel_euler_tetrad.c `test_gr_ultra_rel_euler_tetrad_waves_schwarzschild`
-  vlasov/unit/ctest_hyper3x_dg.c -- Has some print statements which are removed
- vlasov/unit/ctest_dg_em_vars.c -- All gpu tests are commented out, as well as test_2x and 3x tensor p2.
- core/unit/ctest_cudss.cu -- test_simple has some print statements removed
- gyrokinetic/unit/ctest_dg_gyrokinetic_kern_tm.c -- all tests removed and fail

# Additional Notes

So far, I have decided not to run regression tests to speed up CI.
To enhance CI, a few regression tests can be run to compare with the main; however, these should be a selective sample. Rather than building a main for each CI instance, it would be more efficient to have a cron job to initialize the runregression system after each push to main (or every day, but that seems excessive)

We can add Valgrind testing to the CPU build and/or memory sanitizer checks to the GPU build.

This work is progressing to using @JunoRavin 's super server for enhanced Gkeyll robustness and testing. Future work will include nightly runregression testing, powered through cron jobs.

Relevant issues:
https://github.com/ammarhakim/gkeyll/issues/913
https://github.com/ammarhakim/gkeyll/issues/784
fixes https://github.com/ammarhakim/gkeyll/issues/116 (I just found out that using the keywords of "fixes ###" adds this issue to the "development" tab and that issue will be closed when the PR is merged)
https://github.com/ammarhakim/gkeyll/pull/941

# Checklist
- [x] I have reviewed the documentation for accuracy.
- [x] All technical terms and code examples have been double-checked.
- [x] The update aligns with the overall style and tone of the documentation.
- [x] The updated documentation builds correctly (if applicable).


gyrokinetic/data/eqdsk/ltx_miller.geqdsk
```diff
@@ -1,4 +1,3 @@
-81  91
```
JunoRavin, line 1: Just curious, what causes these lines to be modified?

gyrokinetic/unit/ctest_deflated_fem_poisson.c
```diff
@@ -181,7 +181,7 @@ test_zdep_nd_nxnz(int nx, int ny){
   gkyl_fem_parproj_set_rhs(parproj, field_discont_dev, field_discont_dev);
   gkyl_fem_parproj_solve(parproj, field_dev);
 
-  struct gkyl_poisson_bc poisson_bc;
+  struct gkyl_poisson_bc poisson_bc = {0};
```
JunoRavin, line 184: If you are avoiding some issue with uninitialized values, our standard way of initializing structs since this is C and everything defaults to 0 is {}; (no 0 inside)

moments/unit/ctest_wv_gr_euler_tetrad.c
```diff
@@ -328,7 +328,7 @@ test_gr_euler_tetrad_basic_schwarzschild()
         q[67] = 0.0;
         q[68] = x; q[69] = y; q[70] = 0.0;
 
-        double prims[71];
+        double prims[71] = {0};
```
JunoRavin, line 331: Make sure if you are going to add these sorts of initialize to 0 lines that they are the right type ({0.0} in this case).

JunoRavin:
@JonathanGorard before we merge this PR, we should go through some of the GR unit tests to make sure no issue has crept in and it's not just a floating point precision issue on some machines.

gyrokinetic/unit/ctest_dg_gyrokinetic_kern_tm.c
```diff
@@ -181,9 +181,9 @@ test_gyrokinetic_3x2v_p1_cu()
 }
 
 TEST_LIST = {
-  { "test_gyrokinetic_3x2v_p1", test_gyrokinetic_3x2v_p1 },
+  // { "test_gyrokinetic_3x2v_p1", test_gyrokinetic_3x2v_p1 },
```
manauref: i think you should just delete this whole file
it's not being maintained. If someone wants to time the kernels, I think they are better off writing a new test with the latest API

gyrokinetic/zero/efit.c
```diff
@@ -19,6 +19,8 @@
 gkyl_efit* gkyl_efit_new(const struct gkyl_efit_inp *inp)
 {
   gkyl_efit *up = gkyl_malloc(sizeof(struct gkyl_efit));
+  // Zero-initialize the struct to avoid uninitialized value warnings
+  memset(up, 0, sizeof(struct gkyl_efit));
```
manauref: this is relatively harmless, but why is it needed? we don't do this anywhere else.
If necessary, I think it's preferable that instead of gkyl_malloc+memset that you use gkyl_calloc

gyrokinetic/zero/efit.c
```diff
@@ -364,21 +366,30 @@ gkyl_efit* gkyl_efit_new(const struct gkyl_efit_inp *inp)
   int num_max_xpts = 10;
   double Rxpt[num_max_xpts];
   double Zxpt[num_max_xpts];
+  // Initialize arrays to avoid uninitialized value warnings
```
manauref: the changes in this file seem dangerous. I suggest you put this in a separate PR.
It'd be nice to have a unit test for single and double null to make sure this sort of thing is safe, uh @akashukla

manauref (approved):
Approving, though I make a couple of simple requests in my comments that should be resolved.

Also: can @ammarhakim @JunoRavin take a look at the changes to the Makefile and the .github/workflows files please? it'll take you 3 min

gyrokinetic/unit/ctest_dg_gyrokinetic_kern_tm.c
Maxwell-Rosen: Deleted

gyrokinetic/zero/efit.c
Maxwell-Rosen: Changed

gyrokinetic/zero/efit.c
Maxwell-Rosen: Removed. This does seem dangerous, but the current implementation is not valgrind clean because num_xpoints can be zero, in which zero memory is allocated

manauref: open a separate PR just with this.
The biggest problem is the lack of a test. How do we know if something is valgrind clean or not without a test?

Maxwell-Rosen:
I merged main and verified that all unit tests pass and are valgrind clean. There was a merge error with the position map that was pushed to main which was uncaught because we do not have thorough CI.

manauref:
the GK moments unit test is fixed/updated in the SUNDIALS PR, so i think it's better if we let that one fail here for now.

JunoRavin (approved):

gyrokinetic/data/eqdsk/ltx_miller.geqdsk
manauref: those two numbrs are the old convention for our eqdsk file. We don't do that anymore.

PR #990, opened 2026-04-04
Add a low pass time filter to GK

### LINKED ISSUE #985: [DR] Low pass filter in gyrokinetics
# Design Review

## Overview

Description: This design review discusses adding a low-pass filter damping model to the gyrokinetic modules. The low pass filter evolves the species distribution function f_s and a secondary filtered distribution $\bar f_s$:

$$
\frac{d f_s}{d t} = -\nu_{\rm LP} \left(f_s-\bar f_s\right),
$$

$$
\frac{d \bar f_s}{d t} = \nu_{\rm LP} \left(f_s-\bar f_s\right).
$$

Here, $\nu_{\rm LP}$ is the low-pass frequency. The coupled system damps fast oscillations in $f_s$ by relaxing it toward $\bar f_s$, while $\bar f_s$ tracks the slow-time behavior.

Motivation: We studied this model for a low-pass filter with the toy models for the pseudo-orbit-averaging algorithm. We showed that in cases with a centered band gap around the correct mean, a low-pass filter is effective in damping oscillations arising from localized sources and sinks. The simulations I'm running right now have high-energy beams, which are farther away from the trapped-passing boundary. This leads to a regime of distributed source/localized loss. This introduces a band gap during the orbit-averaging phase, which then oscillates around the orbit contour during the full-dynamics phase. This pulse drives negativity and does not seem to dampen. A low-pass filter will enhance the convergence of the FDP in these simulations.

Impact: The change is internal to the gyrokinetic C implementation and primarily affects `gyrokinetic/apps/`.
- Added/updated damping internals in `gyrokinetic/apps/gk_species_damping.c` and `gyrokinetic/apps/gkyl_gyrokinetic_priv.h`.
- Integrated filtered-state stepping in `gyrokinetic/apps/gyrokinetic_update_ssp_rk3.c`.

## Proposed Design

Key Design Elements:
- Treat $\bar f_s$ as a second distribution state that is advanced with the same SSPRK3 stages used for $f_s$. This distribution function is stored inside the gk_species_damping modules.
- Compute filter RHS through damping interfaces:
  - $d\bar f_s/dt = \nu_{\rm LP}(f_s-\bar f_s)$
  - $df_s/dt$ receives the opposite contribution in damping.

At the same time as it making sense that this array of distribution function size is only used in the context of the low pass damping mode, it doesn't make a ton of sense for these deep modifications to SSPRK3. In a sense, we are adding a second distribution function to the simulation, so it would also make sense to add this as a second species. However, if we were to create an extra species, point them to each other for the damping, they would have different types of damping, and the input file would not be elegant. There is also a lot of extra stuff in the gk_species app that we do not need for this and would be completely unnecessary. However, if we proceed in this way, then we do not need to modify the time-advance, as both distribution functions are treated as a species. I can't decide what the right design is.

Interaction with Existing Code:
- New damping type is added
- Existing SSPRK3 flow is preserved; filtering is integrated into the three RK stages
- Current implementation is specific to SSPRK3. If a future SUNDIALS-based integrator is adopted, this coupling should be revisited and generalized.

Prototyping Efforts (if applicable):
- Prototyping was done directly in the working branch through incremental refactors.
- Main finding: evolving `fbar` with SSPRK3 stage consistency is necessary; forward-Euler updates of `fbar` are not consistent with the main time stepper.

Sample Code:

Option 1:
```c
    ...
    .damping = {
      .type = GKYL_GK_DAMPING_LOW_PASS_FILTER,
      .damping_const = 0.1,
    },
  }
```
Option 1 is far more elegant from an input file perspective, which is the most important thing we should design for. I don't think users should really have to remember to define a second species for this damping module.

Option 2:
```c
    ...
    .damping = {
      .type = GKYL_GK_DAMPING_LOW_PASS_FILTER,
      .refrence_species = "fbar",
      .damping_const = 0.1,
    },
  },
  struct gkyl_gyrokinetic_species fbar = {
    .name = "fbar",
    .vdim = ion.vdim,
    .lower = ion.lower,
    .upper = ion.upper,
    .cells = ion.cells,
    .damping = {
      .type = GKYL_GK_DAMPING_LOW_PASS_FILTER_BUFFER,
      .refrence_species = "ion",
      .damping_const = 0.1,
    },
  },
```
Option 2 looks much more elegant from an app layer, not needing additional modifications to SSPRK3, which is being refactored as we speak to integrate Sundials.

I really can't decide between these two designs. Someone help!

## Design Review Checklist

Approval Criteria:

- [] Is the design coherent and consistent with Gkeyll's existing architecture?
- [x] Does the design align with user-facing API standards?
- [x] Do functionalities overlap with existing features?
- [x] Are the algorithms robust and appropriate for the intended purpose?

## Additional Notes

Future Considerations:

Feedback Integration Plan:
- Collect reviewer feedback on numerical consistency, architecture fit, and test coverage.

## Links

Prototyping Branch: (https://github.com/ammarhakim/gkeyll/tree/low-pass-filter)

Additional Resources:
- Low-pass filter intuition: https://en.wikipedia.org/wiki/Low-pass_filter

Maxwell-Rosen:
# Feature

## Summary

The coding for this PR was almost entirely done with GitHub Copilot, with @Maxwell-Rosen providing direction. I certify the results and take responsibility for any mistakes that may have been made.

Purpose: A low-pass filter is added to the gk_species_damping modules to damp high-frequency oscillations in $f$.

$$
\frac{d f_s}{d t} = -\nu_{\rm LP} \left(f_s-\bar f_s\right),
$$

$$
\frac{d \bar f_s}{d t} = \nu_{\rm LP} \left(f_s-\bar f_s\right).
$$

Issue link: fixes #985 

## Implementation Details

Key changes: 
This PR features pretty big changes, extending from gk_species_damping.c, all the way to the SSPRK3 time stepper. This is necessary because the low-pass filter adds an entirely new equation, RHS, and a new variable to step in time. I did my best to preserve the current architecture while simply adding new arguments to existing functions and mirroring how the distribution function is updated.

The new variable `fbar` is constructed as P0 so that this feature doesn't take up too much memory. I am pretty sure that `fbar` and `f` having different bases will not cause discritization errors, but I wouldn’t be surprised if it did.

I had to refactor gk_species_damping quite a lot because there were tons of if statements in the advance functions and the write function.

There are severe merge conflicts with #934, so I ultimately deprecated the loss cone damping term. I'm not using this, and no one else is running mirror simulations. If it's important in the future, we can always bring it back. This was implemented as one of the first ideas regarding pseudo-orbit averaging, which did not work as we intended.

Although this function does look like a BGK operator, we should keep the BGK collisions app seperate, while this term is used to damp oscillations. The damping properties of the low pass filter are why I put this in damping, not the bgk app.

Automated testing: rt_gk_wham_1x2v has included this as a term in its input file

## Example Use

```
  struct gkyl_gyrokinetic_species elc = {
    ...
    .damping = {
      .type = GKYL_GK_DAMPING_LOW_PASS_FILTER,
      .rate_const = 1/1e-6, # Damps on a time scale of 1e-6 seconds
      .write_rate = true,
      .write_fbar = true,
    },
    ....
  },
```

## Community Standards

## Testing: 

## Additional Notes
gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1.c and gyrokinetic/creg/rt_gk_mirror_boltz_elc_damping_1x2v_p1.c are valgrind clean, even with the resets and restoring fbar during different cycles.

Here is a simulation that used the low-pass filter using R=50 for the mirror ratio. I do notice that the time evolution is much slower in the density; however, in the temperatures, it can be quite rapid. Although rapid oscillations are strongly damped, they still occur. This simulation uses `rate_const = 1/5e-6`.
image


Maxwell-Rosen:
This will merge after the sundials branch. It heavily interacts with #996

Issue #923, opened 2025-12-14
[DR] External Coupling to EIRENE


akashukla:
This DR is to describe an interface that we want to add to Gkeyll which will allow it to be easily coupled to external codes for neutrals. I have run various simulations coupled to SOLPS/EIRENE but  they have used modifications to the code that are hacky and not suitable for actually including in the Gkeyll code base.

The new version of the coupling allows EIRENE to do what EIRENE does best (evolution of neutrals, calculation of recycling, sputtering, and calculation of atomic and molecular reactions) while Gkeyll just evolves the plasma. EIRENE will calculate particle, energy, and momentum sources of plasma species and pass them to Gkeyll. Gkeyll will apply these sources with a BGK-like operator similar to what is done in this DR about heat sources https://github.com/ammarhakim/gkeyll/issues/806.

### Overall Coupling Scheme
The coupling uses IO and reads text files which are output from python scripts that interface with EIRENE data. The workflow is as follows: 
Gkeyll runs continuously 
1. Gkeyll outputs data and a flag that indicates EIRENE should be run. Gkeyll reads available EIRENE data.
2. Gkeyll data is processed by python and prepped for EIRENE
3. EIRENE runs and writes its data
4. Python preps the EIRENE data into text files to input into Gkeyll
Return to step 1

### Details of Source Data
Here is some more detail on how the sources from EIRENE go into Gkeyll. The three sources are added to the RHS of GK equation: 

```math
\partial_t f + \dots =  \mathcal{S}_{M2} + \mathcal{S}_{M1} + \mathcal{S}_{M0}
```

The sources take a BGK form where $\nu$ is 1 over the coupling time 
```math
\mathcal{S}_{M2} = \nu  \left(f_{M}(n, u_{\parallel}, v_{tS}^2) - f\right)
```   
```math
\mathcal{S}_{M1} = \nu  \left(f_{M}(n, u_{\parallel S}, v_{t}^2) - f\right)
```  
```math
\mathcal{S}_{M0} = \nu  \left(f_{M}(n_S, u_{\parallel}, v_{t}^2) - f\right)
```  

The subscript S indicates that this is a source temperature, parallel velocity, or density. EIRENE provides a \dot{E}, \dot{M0}, and \dot{M1} for the source so the source temperature, density, and momentum can be calculated by requiring:

```math
   \dot{E} = \frac{m}{2}\int d\mathbf v v^2 S_{M2}
```  
```math
[...]

PR #992, opened 2026-04-13
Eirene Coupling

akashukla:
###  Overview

In this PR we enable the coupling of Gkeyll to Eirene for neutral evolution. The model for including plasma sources in Gkeyll based on eirene output is described in DR #923 . Gkeyll evolves the plasma and Eirene evolves the neutrals (recycling, reactions, etc.). They exchange data through I/O every time Gkeyll outputs data.

This PR contains several parts. The first is the automatic installation of eirene and grid generation parts of SOLPS-ITER as part of the gkeyll mkdeps process. (This only occurs if one uses --build-solps=yes). The second part is the installation of a python library written by @jRoeltgen and me which does post-processing and interpolation routines to pass plasma information from gkeyll to eirene and plasma sources from eirene to gkeyll. The third is the code within Gkeyll itself that applies the sources.
 
### Automatic installation of Eirene + SOLPS-ITER Grid generation
Without modifying the eirene codebase itself, one has to use an existing interface for grid generation and coupling.  We are using the SOLPS-ITER (https://github.com/iterorganization/SOLPS-ITER ) interface. The script `install-deps/build-eirene-coupling.sh` automatically installs the necessary parts SOLPS_ITER into `gkylsoft/`. This is only tested on perlmutter, as the SOLPS-ITER installation process can be difficult. Use `--build-eirene-coupling=yes` in mkdeps to install this (only works on perlmutter).

### Automatic Installation of GK-Neutral_coupling
To actually pass information between Gkeyll and Eirene, post-processing and interpolation routines are required. This is taken care of by the  GK-Neutral_coupling repo (https://github.com/jRoeltgen/GK-Neutral_coupling/tree/main) which contains examples and instructions for how to run coupled simulations. The script `install-deps/build-eirene-coupling.sh` downloads this to gkylsoft.   Use `--build-eirene-coupling=yes` in mkdeps to install this.

 
### Code Within Gkeyll
With guidance from @manauref, the additions to the code have been written in such a way that they interfere minimally with the rest of the gyrokinetic app. Most of the new code is in `gyrokinetic/apps/eirene.c` and `gyrokinetic/apps/gk_species_source_bgk.c`. The second file, `gyrokinetic/apps/gk_species_source_bgk.c`, has now replaced gk_species_heating.c and retains the old functionality while also adding the ability to apply sources from an external model (EIRENE). I also had to modify `array_ops.c` with a new function `gkyl_array_ceil_range` to apply limits on the temperature of the sources come from eirene to improve the stability of simulations.

### Input Files

The input file interface is quite simple. You provide 

```
  struct gkyl_gyrokinetic_eirene  eirene = {
    .input_data_path = "./gkeyll_text_input/",          // Place where text files with eirene sources will appear
    .output_data_path = "./gkeyll_text_output/",    // Place where gkeyll will output the new data flag
    .num_coupling_species = 3,                                 // Number of species with eirene sources
    .coupling_species = {"elc", "ion", "molecule"},  // List of species that have eirene sources 
    .injection_time = {0.5e-3, 0.5e-3, 0.5e-4},          // Injection time sets the timescale or rate of the BGK operator
    .core_coll_factor = {75.0, 75.0, 7.5},                   // Factor by which injection_time is reduced in the core
    .damping_factor= {2.0e-1, 2.0e-1, 8.0e-1},       // This is the minimum fraction of the local 
                                                                                     // temperature that the Maxwellian source can have
                                                                                    // Important for stability
    .half_domain = true,
  };

  struct gkyl_gyrokinetic_multib app_inp = {
    .name = "hstep26",
    ...
    ...
    .eirene = eirene,

  };
```

### Extra Fix
This PR also fixes a typo in the field line tracing and then also adds an option to compress cells near the divertor as part of the nonuniform tokamak grids.
The typo was causing some regions where only one root is found (outer PF region for STEP for example) to be traced wrong. The divertor compression is just a useful option.

### Small Additional Feature
We also add an optional path to geometry FROMFILE so you can read your geometry from that path

### PDFS with instructions and documentation of the operator:
In this pdf the operator and numerical parameters are described. The document gives guidelines for how to set them.
[Eirene_Coupling_Short.pdf](https://github.com/user-attachments/files/27975578/Eirene_Coupling_Short.pdf)

In this pdf we have additional instructions, links to tutorials, etc. for SOLPS-ITER installation and run set up. It also includes some tips about grid generation for Eirene.

[SOLPS steps for Perlmutter.pdf](https://github.com/user-attachments/files/27975574/SOLPS.steps.for.Perlmutter.pdf)


gyrokinetic/apps/eirene.c
```diff
@@ -0,0 +1,212 @@
...
+#include <gkyl_elem_type.h>
+#include <gkyl_eqn_type.h>
+#include <gkyl_gyrokinetic_priv.h>
+#include <gkyl_array_rio.h>
+#include <gkyl_array_rio_priv.h>
+
+#include <assert.h>
+#include <time.h>
+
+static bool gyrokinetic_str_ends_in_b67(char *name){
```
Maxwell-Rosen: Maybe I'm ignorant, but I don't know what this b67 (or b1011) means. So if I'm a gkeyll person who doesn't know EIRENE, can we have a more descriptive names for what makes these two files different? We can put a comment that this means the files end in b67, but the function name should make the purpose and operation clear.

gyrokinetic/apps/eirene.c
```diff
@@ -0,0 +1,212 @@
...
+void
+gk_eirene_write(struct gkyl_gyrokinetic_app *app, double tm, int frame)
+{
+  struct gk_eirene *eirene = app->eirene;
+
+  // Read new EIRENE data.
+  for (int i=0; i<eirene->info.num_coupling_species; ++i) {
+    struct gk_species *gks = eirene->coupling_species[i];
+    struct gk_source_bgk *bgk_src = &eirene->bgk_src[i];
+    cstr fileNm = cstr_from_fmt("%s%s-%s_M0source.txt", eirene->info.input_data_path, app->name, gks->info.name);
```
Maxwell-Rosen, line 119: The convention us usually <species>_source_M0.gkyl, so maybe we should keep that for these formats

gyrokinetic/apps/eirene.c
```diff
@@ -0,0 +1,212 @@
...
+
+struct gk_eirene*
+gk_eirene_init(struct gkyl_gyrokinetic_app *app, struct gkyl_gk *gk)
+{
+
+  struct gk_eirene *eirene = gkyl_malloc(sizeof(struct gk_eirene));
+
+  eirene->info = gk->eirene;
+
+  for (int i=0; i<eirene->info.num_coupling_species; ++i)
```
Maxwell-Rosen: Please put braces around single line if statements and for loops

gyrokinetic/apps/gyrokinetic.c
```diff
@@ -1055,6 +1058,7 @@ gkyl_gyrokinetic_app_write_geometry(gkyl_gyrokinetic_app* app, struct gkyl_gk_ge
     gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->geo_corn.mc2nu_pos_deflated, arr_hocdim, "mc2nu_pos_deflated", mt);  
   }
 
+  gyrokinetic_app_geometry_copy_and_write(app, app->gk_geom->geo_int.mc2p       , arr_ho3, "mapc2pint", mt);
```
Maxwell-Rosen, line 1095: changing this to `mapc2p_int` might be clearrer and more organized

manauref:
Some initial comments:
1. "They exchange data through I/O every time Gkeyll outputs data": does Gkeyll then alter the user's `num_frames` or I/O triggers so that this exchange happens with some reasonable frequency? because what if the user set t_end =1e-3 and num_frames=2? we'll only have one exchange in a whole milisecond.
2. For the record I heard that EDGE2D couples to Eirene without involving SOLPS. They have their own grid generation tools that they use for Eirene. This would be a preferable route in the future.
3. In the input file description above, can you add comments to the lines with numeric parameters please? you could also reference the corresponding symbol/quantity in your DR/latex write up so one can make the connection, if applicable.
4. Neither the DR, nor the pdf in it, nor the description above explain what `core_coll_factor` and `damping_factor` are. Also, greater emphasis/documentation with regards to how `coupling_time` is chosen should appear above. It might even be good to explain what the `coupling_species` are.
5. Why is `.half_domain = true,` in the eirene input file table necessary? I think the code should figure this out under the hood.
6. The [GK-Neutral_coupling](https://github.com/jRoeltgen/GK-Neutral_coupling) repo has good README's. It'd be good if you add a "installation" section, that at the very least lists the dependencies (i.e. python modules, fortran?, etc).
7. We wrote a document with a list of steps for installing on Perlmutter. Although much of that is now in scripts in this PR, it might be good to put that (as PDF) around here somewhere. 
8. Also, it would be good to attach here a PDF with a high-level description of how to generate the EIRENE grid/files needed. It can be 1 or a half page, with links to the relevant tutorials in the SOLPS/EIRENE modules/repos.
9. It's a bit unfortunate not to have a regression test, though I understand why that's done. That said I think you should:
9a) add one anyway, and put pre-processor statements around the eirene table so that if someone runs it without eirene, then it just doesn't do that part but still runs.
9b) run the reg test on your Perlmutter build, on cpu and gpu, and ensure it is valgrind and compute-sanitizer clean.

install-deps/mkdeps.sh
```diff
@@ -167,6 +169,14 @@ do
...
       BUILD_ADAS="$value"
       ;;
+   --build-solps)
+      [ -n "$value" ] || die "Missing value in flag $key."
+      BUILD_SOLPS="$value"
+      ;;   
+   --build-eirene-coupling)
+      [ -n "$value" ] || die "Missing value in flag $key."
+      BUILD_EIRENE_COUPLING="$value"
+      ;;
```
manauref: why do we need two options, if we just care about the eirene coupling? can we just have a --build-eirene option that does it all?

note that I name it Eirene. It has to download SOLPS, but from our point of view that is just a dependency of EIRENE, so it's best not to mention SOLPS as the option. In the future we may change the actions of this command so it only works with EIRENE

install-deps/build-eirene.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
manauref, line 1: if these build scripts where partially generated by LLMs, please state so (and which) in a comment.

install-deps/build-eirene.sh
```diff
@@ -0,0 +1,19 @@
...
+# ==============================================================================
+# Gk-Neutral_coupling Automated Installer
+# ==============================================================================
+
+echo ">>> Starting GK-Neutral_coupling Installation..."
+cd ../../gkylsoft/
+
+if [ ! -d "GK-Neutral_coupling" ]; then
+    echo ">>> Downloading GK-Neutral_coupling..."
+    wget https://github.com/jRoeltgen/GK-Neutral_coupling/archive/refs/tags/gkeyll.tar.gz
```
manauref: is this downloading the released version?

core/zero/gkyl_eqn_type.h
```diff
@@ -75,10 +75,11 @@ enum gkyl_source_id {
...
 };
 
-// Identifiers for specific heating source object types.
-enum gkyl_heating_id {
-  GKYL_NO_HEATING = 0, // No source. This is default.
-  GKYL_HEATING_DEFAULT, // Default heating source.
+// Identifiers for specific bgk source object types.
+enum gkyl_source_bgk_id {
+  GKYL_NO_SOURCE_BGK = 0, // No source. This is default.
+  GKYL_SOURCE_BGK_DEFAULT, // Default heating source.
```
manauref: rename _HEATING instead of _DEFAULT?
there's no reason for heating to be the default. This can now be used to add particles, or momentum, or some combination of particles/momentum/energy

core/zero/gkyl_array_ops.h
```diff
@@ -325,6 +325,17 @@ void gkyl_array_flip_copy_to_buffer_fn(void *data, const struct gkyl_array *arr,
...
 
+/**
+ * Compute out = max(out,inp). Returns out.
+ *
+ * @param out Output array
+ * @param inp Input array
+ * @param range Range to take max over
+ * @return out array
+ */
+struct gkyl_array* gkyl_array_ceil_range(struct gkyl_array *out,
```
manauref: I think the name of this method and its description needs to be more specific. It seems that you taking the max of out and inp, but only based on the 0th coefficient. The description should state that, and perhaps the method should be called something like
`gkyl_array_max_by_cell_per_cell_avg`
(borrowing the part of the name name of a method @Maxwell-Rosen added in [Time dilation methods to suppress CFL](https://github.com/ammarhakim/gkeyll/pull/946/changes#top), but which I asked him to delete).

it'd be good if you add the non-range method too. Sometimes those perform better (because of GPU memory access patterns), even if one only cares about a specific range.

manauref (changes requested):

install-deps/build-eirene.sh
akashukla: yes

install-deps/build-eirene.sh
akashukla: done

core/zero/gkyl_eqn_type.h
akashukla: fixed and comments updated

core/zero/gkyl_array_ops.h
akashukla: Renamed. We decided not adding the no-range version is ok for now.

akashukla:
1. No we don't modify it. The user just has to set the configuration space output frequency to be reasonable. In the pdf I updated I have added guidelines for this. As discussed in the CEDA meeting, people don't really know how to do it correctly per se.
2. Agreed, we want to do this eventually.
3. I added descriptions in the input file example and I added guidelines about the numerical parameters in the PDF.  I  listed some physical processes and grid limitations to consider.
4. . I added descriptions in the input file example and I added guidelines about the numerical parameters in the PDF.  I  listed some physical processes and grid limitations to consider.
5.   Ok I made it part of gk_geom now.
6. Ok dependencies added to that README in a branch. Will make a new release when we are ready to merge.
7. Ok I added a pdf of that document. It has some additional tips and things for grid generation
8. The attached doc from 7. has links and the order of the steps in general and all that.
9. I added rt_gk_multib_step_eirene_2x2v_p1 and added the necessary data to gyrokinetic/data/eirene. compute sanitizer says :
 ========= LEAK SUMMARY: 0 bytes leaked in 0 allocations
 ========= ERROR SUMMARY: 0 errors
 I also ran the cpu version through valgrind and it's clean.

install-deps/mkdeps.sh
```diff
@@ -167,6 +168,10 @@ do
       [ -n "$value" ] || die "Missing value in flag $key."
       BUILD_ADAS="$value"
       ;;
+   --build-eirene-coupling)
```
Antoinehoff: I see that this flag is not the same as the one described in the PR text, is it the replacement of `--build-solps` ?

akashukla: yes now we just have --build-eirene-coupling that builds everything. I changed it in response to Mana's comments. Let me change the main PR comment.

manauref: think it should be `--build-eirene`

Antoinehoff: yes and also add this option to the `-h` message please

Antoinehoff: I propose to have this in the help message:
```
--build-eirene              [no] Should we build eirene? (includes installing SOLPS-ITER)
```
to make aware that this will install SOLPS-ITER

manauref (approved):
approving, but maybe wait to merge until a release in the GK-Neutral_coupling repo is created.

note for future: it's a bit concerning that MB input files are seg faulting if we don't malloc the input struct

install-deps/mkdeps.sh
akashukla: ok did this

gyrokinetic/apps/eirene.c
akashukla: done

gyrokinetic/apps/eirene.c
akashukla: This is not the same type of file (it is not a moment of a gkeyll distribution function or phase space source). The output files that are moments of the phase space source are named using the normal convention. It would be confusing with the same name and this is compatible with the coupling repo, so we will keep this name for now.

gyrokinetic/apps/eirene.c
akashukla: I added comments for these to explain.

gyrokinetic/apps/gyrokinetic.c
akashukla: mapc2pint is a fine name. It is different from the corner mapc2p.

akashukla:
> approving, but maybe wait to merge until a release in the GK-Neutral_coupling repo is created.
> 
> note for future: it's a bit concerning that MB input files are seg faulting if we don't malloc the input struct

Ok now the new release is out and the build scripts download that.

Antoinehoff:
I am building from scratch this PR and will run the regression test on Perlmutter (once the maintenance is over) to be sure that everything is fine. Ideally, I would like also to try a very simple test with Eirene running too. If I understood correctly, the regression test uses a predefined Eirene output, is there another file I could try to run that would setup and run Eirene too?

akashukla:
> I am building from scratch this PR and will run the regression test on Perlmutter (once the maintenance is over) to be sure that everything is fine. Ideally, I would like also to try a very simple test with Eirene running too. If I understood correctly, the regression test uses a predefined Eirene output, is there another file I could try to run that would setup and run Eirene too?

You're right this regression test only checks the gkeyll side. There are things you could do to actually run the eirene side, but you would have to do a lot. The other repo: GK-Neutral_coupling has example bash scripts and such for how to do such a test. (There are quite a few steps). If you want I could transfer the data (interpolation mappings generated using the other repo) to you from one of my simulations, and you can try to run eirene using your eirene install.

Antoinehoff (approved):
Great work thanks @akashukla and all the one who participated!!

PR #993, opened 2026-04-13
[aux PR] Get em back full upwind

Antoinehoff:
This is an auxiliary PR related to #888. It has a full upwind scheme instead of the partial upwinding scheme presented in #888. For now it is not working as expected, we use this PR to have a better view on the changes compared to main.


Antoinehoff:
The get_em_back branch is now pointing to the latest commit of this branch since it is now more advanced. The PR stays https://github.com/ammarhakim/gkeyll/pull/888

PR #995, opened 2026-04-16
Toroidally symmetry check and helical turbulence regression test

Antoinehoff:
We propose to add the sheared helical turbulence C file as a regression test, `rt_gk_helical_zvert_3x2v_p1.c`. It is based on Section 5.2, "Helical SOL configuration including magnetic shear", Thesis of Noah Mandell, 2021.

The test passes the geom right handed check (no warnings) and conserves particle exactly, the energy conservation is less good but okay:
Screenshot 2026-04-30 at 10 03 40 AM

If the resolution is increased to $N_x=16$, $N_y=32$, $N_z=12$, $N_{v\parallel}=10$, and $N_\mu=5$, one gets nice blobby turbulence
helical_sol_ne_frames_0_to_216
_This simulation ran for 4h on 3 Perlmutter's A100 GPUs. It performed 128400 time steps to arrive at 216 mus_

## Additional study: shearless regression test, right hand check and conservation test
We can also add the shearless geometry in `rt_gk_helical_zpar_3x2v_p1.c` where the `z` coordinate is aligned with the magnetic field. This geometry is taken from Mandell et al. 2020 and is known to have inconsistencies (one cannot hat a shearless magnetic field that intersect a horizontal plane with constant arc length). The original geometry is not passing the right hand check (see #933) and is not expected to show good conservation properties. One can modify the geometry to pass the right hand check with the following toroidal angle mapping
```
 phi(R) = (y/sin(theta) + z*cos(theta))/R;
```
However, this mapping does not show particle conservation, which demonstrate that the right hand check maybe a necessary but not sufficient constraint to have a well behaved geometry. 

Screenshot 2026-04-30 at 10 06 07 AM
395" />

One can fix it by adopting the same toroidal angle mapping as the sheared regression test, i.e.
```
phi(R) = y/Rc + (Bt * z*sin(theta))/(Bv * R);
```

Now the particle and energy conservation are fine
Screenshot 2026-04-30 at 10 04 18 AM

## Follow up
@akashukla found out that the geometry that was not conserving particle above was breaking toroidal symmetry. This PR now includes a check of the variation of the metric coefficients along `y`.


gyrokinetic/creg/rt_gk_helical_zpar_3x2v_p1.c
```diff
@@ -0,0 +1,634 @@
...
+
+  double t_end; // End time.
+  int num_frames; // Number of output frames.
+  double write_phase_freq; // Frequency of writing phase-space diagnostics (as a fraction of num_frames).
+  int int_diag_calc_num; // Number of integrated diagnostics computations (=INT_MAX for every step).
+  double dt_failure_tol; // Minimum allowable fraction of initial time-step.
+  int num_failures_max; // Maximum allowable number of consecutive small time-steps.
+};
+
+double random0to1()
```
manauref: random noise breaks regression testing. Better remove

manauref (changes requested):

gyrokinetic/creg/rt_gk_helical_zpar_3x2v_p1.c
Antoinehoff: ok

gyrokinetic/creg/rt_gk_helical_zvert_3x2v_p1.c
```diff
@@ -0,0 +1,653 @@
...
+{
+  double x = xn[0], y = xn[1], z = xn[2];
+  struct gk_app_ctx *app = ctx;
+  double Ls = app->Lz/4;
+  double xSource[3] = {x, y, 0};
+  double effectiveSource = sourceDensity(t, xSource, ctx);
+  double c_ss = sqrt(5/3*sourceTemperature(t, xSource, ctx)/app->mi);
+  double nPeak = 4*sqrt(5)/3/c_ss*Ls*effectiveSource/2;
+  pcg64_random_t rng = gkyl_pcg64_init(0);
+  double perturb = 1e-3*(gkyl_pcg64_rand_double(&rng) - 0.5)*2.0;
```
manauref, line 119: it's best to not use random number generators, otherwise regression testing fails. Just set pert to 1e-3.

manauref (changes requested):

gyrokinetic/creg/rt_gk_helical_zvert_3x2v_p1.c
JunoRavin: I pointed this out to @Antoinehoff because we use these deterministic random numbers elsewhere in the regression system and I thought it might be helpful if he wanted the perturbation to be random (or to add more randomness). gkyl_pcg64_init(0) initiates a random number generator from a deterministic seed that is encoded in util.c. We use this same random number generator and get regression tests which pass in moments and vlasov.

manauref (approved):

JunoRavin (approved):

PR #997, opened 2026-04-21
GK: loss cone updater refactored for arbitrary geometry

Maxwell-Rosen:
# Hamiltonian-based Classification of Loss and Trapped Orbits in Mirror Geometry

This PR was generated with the assistance of GitHub Copilot for the unit tests and the updater functions. Most of it is LLM-generated, but I've read it all and understand it.

Consider a one-dimensional coordinate $z\in[z_L,z_R]$ aligned with a magnetic field line, with magnetic-field magnitude $B(z)$ and electrostatic potential $\phi(z)$. For a species of mass $m_s$ and charge $q_s$, the guiding-center Hamiltonian in $(z,v_\parallel,\mu)$ coordinates is

$$
H(z,v_\parallel,\mu)=\frac{1}{2}m_s v_\parallel^2+\mu B(z)+q_s\phi(z),
$$

 where $v_\parallel$ is the parallel velocity and $\mu$ is the magnetic moment. In the absence of collisions, $H$ and $\mu$ are invariants of motion, and particle trajectories lie along contours of constant $H$ in $(z, v_\parallel)$.
Turning points satisfy $v_\parallel=0$, hence

 $$H=U(z,\mu),
\qquad
U(z,\mu)= \mu B(z)+q\phi(z),
$$

where $U$ is the Yushmanov potential.

Prior work simply identified the trapped/passing boundary based on the
loss cone criterion

$$\mu  = \frac{\frac{1}{2}m_s v_\parallel^2 + q_s\left(\phi(z) - \phi_m\right)}{B_m - B(z)}.$$

In that implementation, $B_m,\phi_m$ are determined at the peaks of $B$. However, the effective potential experienced by particles is $U$, which combines both $B$ and $\phi$. This can be essential for identifying the trapped region of phase space induced by angled neutral-beam injection of ions at $45^\circ$, where an off-axis density peak occurs. Furthermore, this implementation only uses a single loss criterion per region, when multiple can exist, for instance, in tandem mirrors. A Hamiltonian-based region-identification algorithm is more accurate and flexible because it directly addresses the Yushmanov potential. 

## Two-Wall Escape Criterion

One issue with a Hamiltonian-based identifier is that $H$ is non-unique. For instance, consider a symmetric simple mirror without an electric potential, just two peaks in B. The magnetic field has fourfold degeneracy at the half-maximum. The points outside the peaks in $B$ should be labeled passing, while those between peaks of $B$ should be trapped.

A trajectory at fixed $(H,\mu)$ can escape through the left wall at $z_L$ if it can traverse the interval $[z_L,z]$ without violating $H\ge U$. For a phase-space point $(z,v_\parallel,\mu)$, define the escape barrier to the left wall as

$$EB_L(z,\mu)=\max_{s\in[z_L,z]}U(s,\mu),
$$ 

and the escape barrier to the right wall as

$$EB_R(z,\mu)=\max_{s\in[z,z_R]}U(s,\mu).
$$ 

The minimum escape barrier required for escape through at least one wall is

$$EB(z,\mu)=\min\left(EB_L(z,\mu),EB_R(z,\mu)\right).
$$ 

Defining the signed loss function

$$\mathcal{F}(z,v_\parallel,\mu)=H(z,v_\parallel,\mu)-EB(z,\mu).
$$ 

$\mathcal{F}\ge 0$ means that there is no energy barrier between the wall and this point, indicating a passing orbit. If $\mathcal{F}<0$, then the escape barrier is at a higher energy than this point, so it is trapped. So the orbiting mask is

$$\chi_{\rm orbit}(z, v_\parallel,\mu) = \mathrm{bool}(\mathcal{F}(z,v_\parallel,\mu)<0)$$

## Inclusion of Sheath Wall Potentials

The field $\phi(z)$ does not directly represent the sheath drop at the wall, but instead it is assumed that $\phi=0$ in Gkeyll's sheath boundary conditions. Let $\phi_L^{\mathrm{bc}}$ and $\phi_R^{\mathrm{bc}}$ denote prescribed wall potentials at $z_L$ and $z_R$, respectively, and let $\phi(z_L)$ and $\phi(z_R)$ be the values implied by the loaded field data near the walls. These are left general, but taken as zero in the simulations. The effective wall potentials are taken as 

$$\begin{aligned}
U_L^{\mathrm{wall}}(\mu)
&=\max\left(\mu B(z_L)+q\phi(z_L),\mu B(z_L)+q\phi_L^{\mathrm{bc}}\right),
\\
U_R^{\mathrm{wall}}(\mu)
&=\max\left(\mu B(z_R)+q\phi(z_R),\mu B(z_R)+q\phi_R^{\mathrm{bc}}\right).
\end{aligned}$$ 

The corresponding augmented path barriers are

$$\begin{aligned}
\widetilde{EB}_L(z,\mu)&=\max\left(\mathcal{B}_L(z,\mu),U_L^{\mathrm{wall}}(\mu)\right),
\\
\widetilde{EB}_R(z,\mu)&=\max\left(\mathcal{B}_R(z,\mu),U_R^{\mathrm{wall}}(\mu)\right),
\end{aligned}$$

 with an augmented escape barrier

$$\widetilde{EB}(z,\mu)=\min\left(\widetilde{EB}_L(z,\mu),\widetilde{EB}_R(z,\mu)\right).
$$ 

Replacing $EB$ by $\widetilde{EB}$
yields a classifier that includes sheath reflection consistently.

## Integration into the Gkeyll Workflow

Within Gkeyll, one first interpolates the fields to nodal values, computes $EB$, calculates the loss barrier, and then, if any corners of the cell are not orbiting, the cell is not evolved.

## Community Standards

## Additional notes

Here is a comparison in a beam simulation with the current loss mask and the old one.
This is an R=32 mirror case with Boltzmann electrons, showing the difference in the fdot multiplier. Red means these cells are now masked with the new updater, while blue means cells that were masked, but are not with this version. In other words, this is new_mask - old_mask. We see the population of small mu particles which are trapped due to the off-axis beam injection, which is what we expect.

newplot

Here is another view, this time in pyvista
image

Here is a simulation run with the new loss cone mask. I'm re-running a beam R=32 case, which was used for the paper [Gyrokinetic equilibria of high temperature superconducting magnetic mirrors
](https://arxiv.org/abs/2604.11684)
This did decrease the time-step during the OAP to 3.4e-7 versus 2.6e-6 seconds. Perhaps a more coarse mesh in mu would help.
yushmanov-comparison


manauref:
Looking at the "Inclusion of Sheath Wall Potentials", why do you have to introduce augmented escape barriers (EBs), instead of defining the left and right EBs as

$$EB_L(z,\mu)=\max_{s\in[z_L,z]}U_L(s,\mu),
$$ 

$$EB_R(z,\mu)=\max_{s\in[z,z_R]}U_R(s,\mu),
$$ 

where

 $$U_L(z,\mu)= \mu B(z)+q(\phi(z)-\phi(z_L)),
$$

 $$U_R(z,\mu)= \mu B(z)+q(\phi(z)-\phi(z_R)),
$$

? If this is valid it seems to me like it simplifies the physical motivation/logic.

Maxwell-Rosen:
I'm not sure I understand. It seems like the equations you have in the comment are just a rephrasing of the ones in the PR description. If anything, I think my formulation of this is clearer because I explicitly use the term $\phi^{bc}$ for the boundary condition, so it's a bit unclear from your formulation where the boundary condition enters. 

I asked Codex about this, and it says the sheath potential modifications aren't currently implemented. I think it's misunderstanding, though, since $\phi^{bc}$ is zero for the sheath boundary conditions.

Maxwell-Rosen:
I meant to have the latest, most powerful models inspect this PR. It found that the wall potential wasn't included in this implementation (although it's zero for the sheath BCs), so I had it implement that to be more comprehensive and physical

gyrokinetic/apps/gk_species_damping.c
```diff
@@ -1,16 +1,24 @@
 #include <assert.h>
-#include <gkyl_gyrokinetic_priv.h>
```
manauref, line 2: why are you editing the gk_species_damping module? I thought you weren't using this nor needed to modify it?

gyrokinetic/zero/gkyl_position_map_priv.h
```diff
@@ -314,15 +314,24 @@ find_B_field_extrema(struct gkyl_position_map *gpm)
...
     xp[Z_IDX] = theta;
     gkyl_calc_bmag_global(0.0, xp, &bmag_vals[i], bmag_ctx);
     dbmag_vals[i] = calc_bmag_global_derivative(theta, gpm);
-    if (i==0) continue;
 
-    // Minima
-    if (dbmag_vals[i] > 0 && dbmag_vals[i-1] < 0){
+    // Near-zero derivative: B is locally flat here, record as a minimum.
+    // Use continue so this is mutually exclusive with the sign-change checks below.
+    if (fabs(dbmag_vals[i]) < 1e-10) {
```
manauref: should this comparison use the derivative normalized to some sensible value instead? perhaps something like
```
dbmag_vals[i] / ( B at this location / dz ) < 1e-10
```
typically comparing a physical value (dB/dz in this case) to a single number tends to be less robust.

Same for all the dbmag comparisons below.

gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1_2species.c
```diff
@@ -0,0 +1,1030 @@
+#include <math.h>
```
manauref: how is this reg test different than `rt_gk_mirror_boltz_elc_poa_1x2v_p1.c‎`? if it has kinetic elc, then name it so.

and are all these regression tests running in under 90 s?

gyrokinetic/unit/ctest_loss_cone_mask_gyrokinetic.c
```diff
@@ -1,220 +1,401 @@
 #include <acutest.h>
 
-#include <gkyl_util.h>
+#define _USE_MATH_DEFINES
```
manauref: don't see why this is needed

gyrokinetic/unit/ctest_loss_cone_mask_gyrokinetic.c
```diff
@@ -1,220 +1,401 @@
...
-#include <gkyl_const.h>
+#include <gkyl_position_map.h>
+#include <gkyl_range.h>
+#include <gkyl_rect_grid.h>
+#include <gkyl_util.h>
+#include <gkyl_velocity_map.h>
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
```
manauref: don't think this is needed
if necessary see gkyl_const.h

gyrokinetic/unit/mctest_loss_cone_mask_gyrokinetic.c
```diff
@@ -0,0 +1,230 @@
...
+#include <gkyl_rect_grid.h>
+#include <gkyl_velocity_map.h>
+
+#ifdef GKYL_HAVE_MPI
+#include <gkyl_mpi_comm.h>
+#endif
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
```
manauref: not needed, see comment in other unit test

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
```diff
@@ -10,70 +10,67 @@
...
 
-// Type of function expected for the ctp_pos_func input.
-typedef void (*loss_cone_mask_gyrokinetic_c2p_t)(const double *xcomp, double *xphys, void *ctx);
-
-// Available options:
-//   A) num_quad=1, qtype=GKYL_GAUSS_QUAD. Output: ncomp=1 array.
-//   B) num_quad>1, qtype=GKYL_GAUSS_QUAD or GKYL_GAUSS_LOBATTO_QUAD, cellwise_trap_loss=true. Output: ncomp=1 array.
-//   C) num_quad>1, qtype=GKYL_GAUSS_QUAD or GKYL_GAUSS_LOBATTO_QUAD, cellwise_trap_loss=false. Output: ncomp=phase_basis.ncomp array.
+/** Boundary behavior used by the orbit classifier. */
+enum gkyl_loss_cone_boundary_type {
```
manauref: rename `gkyl_gk_loss_cone_boundary_type` and options 'GKYL_GK_LOSS`

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
```diff
@@ -10,70 +10,67 @@
...
+  GKYL_LOSS_CONE_BC_OPEN = 0, // A particle that reaches the domain edge is lost.
+  GKYL_LOSS_CONE_BC_SHEATH,  // The material-wall potential may reflect the particle.
+  GKYL_LOSS_CONE_BC_CLOSED,  // No particle can be lost through this edge.
```
manauref: why does there need to be a distinction between being reflected inside the domain or the sheath?

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
manauref: also, it seems to me, seeing the description of the options below, that this is more of an particle trajectory type, rather than a boundary type. If you agree, then maybe this should be renamed

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
```diff
@@ -10,70 +10,67 @@
...
- *     f = 0
- * where mu_bound = (0.5*m*vpar^2+q*(phi-phi_m))/(B*(B_max/B-1))
- * is the trapped-passing boundary in vpar-mu space.
+ * based on the escape-barrier criterion. For sheath boundaries the material
+ * wall is a virtual endpoint outside the plasma domain, so
+ *   EB_L = max(max_{s in [z_L,z]} U(s,mu), mu*B(z_L)+q*phi_wall_L)
+ *   EB_R = max(max_{s in [z,z_R]} U(s,mu), mu*B(z_R)+q*phi_wall_R)
+ * with U = mu*B + q*phi. Open boundaries omit the virtual endpoint and
+ * closed boundaries cannot be escaped.
+ * A node is trapped if H < EB.
```
manauref: i would change the language from open/closed boundaries to open and closed trajectories. The current language is confusing and not something I've heard physicists use

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
```diff
@@ -10,70 +10,67 @@
...
+ * the app is responsible for assembling global data when needed.
  *
  * @param up Project on basis updater to run.
  * @param phase_rng Phase-space range.
  * @param conf_rng Configuration-space range.
- * @param phi Electrostatic potential.
- * @param phi_m Electrostatic potential at the mirror throat (on GPU if use_gpu=true).
- * @param mask_out Output masking function.
+ * @param bmag Magnetic field magnitude (cdim DG expansion).
+ * @param phi Electrostatic potential at the plasma/sheath entrance.
```
manauref: delete " at the plasma/sheath entrance". I think this is phi everywhere

gyrokinetic/zero/gkyl_position_map_priv.h
```diff
@@ -276,13 +276,13 @@ calc_bmag_global_derivative(double theta, void *ctx)
...
   xh[1] = gpm->constB_ctx->alpha;
-  xh[2] = theta - h;
+  xh[2] = theta + h;
   gkyl_calc_bmag_global(0.0, xh, fout, bmag_ctx);
   double Bmag_plus = fout[0];
-  xh[2] = theta - 2*h;
+  xh[2] = theta - h;
   gkyl_calc_bmag_global(0.0, xh, fout, bmag_ctx);
   double Bmag_minus = fout[0];
-  return (Bmag_plus - Bmag_minus) / (h);
```
manauref, line 285: what's the reasoning behind these changes?

manauref: and can you add a comment in the PR description about the changes to position map please? it's not described there

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -1,485 +1,166 @@
...
-  struct gkyl_range qrange;
-  gkyl_range_init_from_shape(&qrange, dim, qshape);
-  return qrange;
+  return use_gpu? gkyl_array_cu_dev_new(GKYL_DOUBLE, nc, size)
+    : gkyl_array_new(GKYL_DOUBLE, nc, size);
 }
 
-// Sets ordinates, weights and basis functions at ords.
-// Returns the total number of quadrature nodes
 static int
```
manauref: add description comment to this function. And specify which "nodes" are you referring to here (gauss-legendre, gauss-lobatto, basis nodes, positivity nodes, etc)

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic_priv.h
```diff
@@ -1,112 +1,228 @@
...
-#include <gkyl_rect_grid.h> 
 #include <gkyl_util.h>
 #include <assert.h>
 
 GKYL_CU_DH
 static inline void
-log_to_comp(int ndim, const double *eta,
-  const double * GKYL_RESTRICT dx, const double * GKYL_RESTRICT xc,
-  double* GKYL_RESTRICT xout)
+nodal_coords(int ndim, int node, double *x)
```
manauref: add descriptive comment for this function,  including which nodes this refers to (guass-lobatto, gauss-legendre, basis, positivity, etc)

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -1,485 +1,166 @@
...
-
 void
 gkyl_loss_cone_mask_gyrokinetic_advance(gkyl_loss_cone_mask_gyrokinetic *up,
   const struct gkyl_range *phase_range, const struct gkyl_range *conf_range,
-  const struct gkyl_array *phi, const double *phi_m, struct gkyl_array *mask_out)
+  const struct gkyl_array *bmag, const struct gkyl_array *phi,
+  const struct gkyl_array *phi_wall_lo, const struct gkyl_array *phi_wall_up,
+  struct gkyl_array *mask_out)
 {
+  assert(up->lower_boundary != GKYL_LOSS_CONE_BC_SHEATH || phi_wall_lo);
```
manauref: does this whole PR work if you change SHEATH to ABSORB in the input file?

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -1,485 +1,166 @@
...
-      phi_quad[n] = 0.0;
-      for (int k=0; k<num_basis_conf; ++k)
-        phi_quad[n] += phi_d[k]*b_ord[k];
-
-      if (Dbmag_quad[n] > 0.0)
-        qDphiDbmag_quad[n] = up->charge*(phi_quad[n]-phi_m[0])/Dbmag_quad[n];
-      else
-        qDphiDbmag_quad[n] = 0.0;
+  int num_phase_nodes = (int)pow(2.0, (double)pdim);
+  int num_vel_nodes = (int)pow(2.0, (double)vdim);
```
manauref: see my comment on nodes above, add comments here

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -1,485 +1,166 @@
...
+
+  // Outer loop over all phase space cells
+  struct gkyl_range_iter phase_iter;
+  gkyl_range_iter_init(&phase_iter, phase_range);
+  while (gkyl_range_iter_next(&phase_iter)) {
+    int conf_idx[GKYL_MAX_DIM] = { 0 };
```
manauref: MAX_CDIM I think

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -1,485 +1,166 @@
...
+    int vel_idx[GKYL_MAX_DIM] = { 0 };
```
manauref: MAX_VDIM I think

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
```diff
@@ -1,485 +1,166 @@
...
-          fq[0] = 0.0;
+    const struct gkyl_velocity_map *gvm = up->vel_map;
+    long linidx_vel = gkyl_range_idx(&gvm->local_ext_vel, vel_idx);
+    const double *vmap_d = gkyl_array_cfetch(gvm->vmap, linidx_vel);
+
+    // At each phase space cell, loop over all nodes and check if they are trapped or not. If any node is not trapped, the whole cell is not trapped, so we can break out of the loop early in that case.
+    bool cell_trapped = true;
+    for (int node = 0; node < num_phase_nodes && cell_trapped; ++node) {
+      int conf_node = node / num_vel_nodes;
+      int vel_node = node % num_vel_nodes;
```
manauref: note that this hardcodes a particular ordering to the nodes, which is not desirable. It's best to use ranges where that order is abstracted out and can more easily be changed

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic_priv.h
```diff
@@ -1,112 +1,228 @@
...
-                                                          // space nodal to modal gkyl arrays.
+  enum gkyl_loss_cone_boundary_type lower_boundary; // Lower parallel-boundary behavior.
+  enum gkyl_loss_cone_boundary_type upper_boundary; // Upper parallel-boundary behavior.
+
+  struct gkyl_array *basis_at_nodes_conf; // Basis functions at configuration-space nodes.
 };
 
+GKYL_CU_DH
+static inline int
+conf_node_z_endpoint_index(int cdim, int conf_node, int zdim)
```
manauref: add some prefix, like gk_lcm_, to all these static private functions so it's clear where they are and that they don't come from other headers

manauref: and add comments describing them, and their inputs when they are not absolutely obvious (cdim is an example of obvious)

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic_priv.h
```diff
@@ -1,112 +1,228 @@
...
- * @param up Project on basis updater to run.
- * @param phase_rng Phase-space range.
- * @param conf_rng Configuration-space range.
- * @param phi Electrostatic potential.
- * @param phi_m Electrostatic potential at the mirror throat (on GPU).
- * @param mask_out Output masking function.
- */
-void
-gkyl_loss_cone_mask_gyrokinetic_advance_cu(gkyl_loss_cone_mask_gyrokinetic *up,
+#ifdef __cplusplus
```
manauref: don't know why this is needed. We've never had this before

gyrokinetic/zero/loss_cone_mask_gyrokinetic_cu.cu
```diff
@@ -1,277 +1,120 @@
...
-    else
-      fq[linc2] = 0.0;
+    long linidx_phase = gkyl_range_idx(&phase_range, phase_idx);
+    double *mask_d = (double *)gkyl_array_fetch(mask_out, linidx_phase);
+    mask_d[0] = cell_trapped ? 1.0 : 0.0;
   }
 }
 
-void
+extern "C" void
```
manauref: i'd prefer that the use of extern C follows the pattern used before and in other .cu files

manauref (changes requested):
Detailed comments in the review. Overall this seems good, thanks for doing it Max. My only overall comments are:
1. It's clear that LLMs wrote this and several of its choices seem odd. I think the code would've been improved (and reviewer's time saved) with a more careful read/thought prior to PR submission.
2. Please explain how correctness is determined or understood. What is each unit test checking and how was the correct value obtained?
3. I think the equations/logic for escape barriers can be made clearer, as in my earlier comment, but we don't have to act on that now.

gyrokinetic/apps/gk_species_damping.c
Maxwell-Rosen: I'm editing the damping module because there is code in there that utilizes the loss cone updater. Nobody is using this feature, so I deprecated it.

gyrokinetic/unit/ctest_loss_cone_mask_gyrokinetic.c
Maxwell-Rosen: I agree. It might be doing this because there is a longstanding linting issue (or its just on my laptop) where it doesn't recognize M_PI.

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
Maxwell-Rosen: Even better, how about GKYL_GK_MIRROR_TRAPPED

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
Maxwell-Rosen: The boundaries differ if one edge is a Yushmanov potential peak versus a wall.

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
manauref: first thing to clarify is whether these are particle trajectories. If so I'd change the language throughout. Then maybe call this
```
gkyl_gk_trapped_passing_orbit_type {
  GK_GK_TRAP_PASS_ORBIT_PASSING,
  GK_GK_TRAP_PASS_ORBIT_TRAPPED_SHEATH,
  GK_GK_TRAP_PASS_ORBIT_TRAPPED_WALL,
}
```
at first I was hesitant to use "orbit" (and favored "trajectory") because orbit often means closed to some people, but open/unbounded/passing "orbits" are ubiquitous in many parts of physics so I think orbit is fine.

gyrokinetic/zero/gkyl_position_map_priv.h
Maxwell-Rosen: It's just checking if Bmag is increasing or decreasing here. The difference is taken elsewhere

gyrokinetic/creg/rt_gk_mirror_boltz_elc_poa_1x2v_p1_2species.c
Maxwell-Rosen: Deleted. I think this was doing ion-electron collisions, which is why it was a _2species

gyrokinetic/unit/ctest_loss_cone_mask_gyrokinetic.c
Maxwell-Rosen: Not needed. deleted

gyrokinetic/unit/mctest_loss_cone_mask_gyrokinetic.c
Maxwell-Rosen: removed

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
Maxwell-Rosen: renamed

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
Maxwell-Rosen: changed

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
Maxwell-Rosen: changed

gyrokinetic/zero/gkyl_position_map_priv.h
Maxwell-Rosen: I changed it here to do a central difference rather than a backward difference

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
Maxwell-Rosen: Done

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic_priv.h
Maxwell-Rosen: All functions in the private header now have a description

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
Maxwell-Rosen: Yes

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic.h
Maxwell-Rosen: Yes. It classifies particle trajectories, so this terminology fits. I'll update it to use this

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic_priv.h
Maxwell-Rosen: Done

gyrokinetic/zero/loss_cone_mask_gyrokinetic.c
Maxwell-Rosen: Updated

gyrokinetic/zero/gkyl_loss_cone_mask_gyrokinetic_priv.h
Maxwell-Rosen: Removed

gyrokinetic/zero/loss_cone_mask_gyrokinetic_cu.cu
Maxwell-Rosen: Changed

manauref:
ready for approval but: A) has conflicts, B) must pass CI

PR #998, opened 2026-04-22
Checks for IWL Plate Functions

akashukla:
Fix the numerical IWL LTX reg test: the plate functions had mistakes and the LCFS value was just a little bit off. Add some warnings/checks and a convention for the IWL plates to make it easier to construct them correctly. The iwl plate functions should return the corner of the limiter plate (see comments in the header file and reg test) when input parameter s=0.


Antoinehoff (approved):

manauref:
The plate function still has to be designed right, following some criteria we don't fully understand. But this PR prints a key ingredient, which is R_LCFS at the IMP. Taking in that into account makes it possible to produce reasonable grids now. See below:

image

where the plate function had to me modified so that it had this shape near the IMP in order for gridding to work:
image

manauref (approved):

PR #1001, opened 2026-04-28
Broken Mac Build badge from README

Antoinehoff:
The Mac Build badge is broken, should we remove it or replace it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-![Mac Build](https://github.com/ammarhakim/gkeyll/actions/workflows/osx-build.yml/badge.svg)
-
 # About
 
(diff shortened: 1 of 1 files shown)


Maxwell-Rosen:
I fixed the Mac build in #989, so please make sure this PR is compatable with that one. I don't think I fixed the badge, but I think I changed the name / location of the workflow.

manauref (approved):

Antoinehoff:
> I fixed the Mac build in #989, so please make sure this PR is compatable with that one. I don't think I fixed the badge, but I think I changed the name / location of the workflow.

I just removed a broken line, #989 will bring it back I hope.

Maxwell-Rosen (approved):

Issue #1003, opened 2026-05-05
[DR] Core-only GK simulations and deprecation of ES_IWL field_id


manauref:
Tokamak core-only simulations have a variety of uses, e.g.
- Testing various algorithms (e.g. TS BCs).
- Benchmarking against existing results (e.g. T. Gorler, et. al Phys. Plasmas 23, 072503 (2016)).
- Exploring core confinement in STs and other machines.

In this work we propose small modifications to enable core-only simulations with the same ingredients we are currently using for IWL simulations. We additionally propose to deprecate the ES_IWL field_id.

App changes for core-only
---------------------------
The changes are rather minimal. Simply:
1. Allow users to specify the BC_TWISTSHIFT type for BCs in the input file.
2. In the field app, detect this option, and choose methods that use a solver like the core part of the IWL solver.

At first there may be a bit of repetition between core and IWL, but we can clean that up after

Deprecation of ES_IWL
-----------------------
This flag (currently in the field table of the input file) is a bit cumbersome. The field ID should rather pertain the physics model (e.g. ES, EM, Boltzmann, adiabatic, etc). IWL is a geometry thing. So we can instead determine whether it is an IWL sim from the species parallel BCs, or from whether the geometry has an x_LFCS.

We'll introduce a boolean in the gk_field app which simply indicated what type of BC, if any, the field has to obey.

This work is prototyped in the `poisson_perp_bias-core_only` branch.

manauref:
completed in PRs #1019 and #1040

PR #1004, opened 2026-05-05
Core-only GK simulations and deprecation of ES_IWL field id

manauref:
This cleans up the prototyping branch poisson_perp_bias-core_only (see [DR in issue #1003](https://github.com/ammarhakim/gkeyll/issues/1003)).

This enables running core-only simulations specifying TWISTSHIFT as the parallel BC of the species. See reg test
rt_gk_tcv_core_3x2v_p1
which is based on just taking the SOL part of the TCV IWL test. Note however that I did this months ago, and the TCV IWL test may have changed by now. However I ran this simulation (with higher resolution) for several 100 microsec, into turbulence. I'll try to do it again and share plots.

@Antoinehoff noticed that this test uses Neumann BCs for phi at the inner radial boundary. I don't have a good reason for this, and perhaps the BC should be Dirichlet. Neumann is just something I tried months ago that behaved better back then, but we should re-examine.

Future work
------------

Piping the numerical shift into the twistshift updater so we can do core-only with numerical equilibria.


gyrokinetic/apps/gk_field_2x3x.c
```diff
@@ -81,6 +81,69 @@ gk_field_3x_write_twistshift(struct gkyl_gyrokinetic_app *app, struct gk_field *
...
+  // Smooth the DG array.
+  gkyl_fem_parproj_set_rhs(field->fem_parproj_rho_core, field->rho_c_global_dg, field->rho_c_global_dg);
+  gkyl_fem_parproj_solve(field->fem_parproj_rho_core, field->phi_fem);
+
+  // Copy global, continuous FEM array to a local array.
+  gkyl_array_copy_range_to_range(arr_fem, field->phi_fem, &app->local, &field->global_sub_range);
+}
+
+static void
+gk_field_fem_projection_par_phi_ts_2x(gkyl_gyrokinetic_app *app, struct gk_field *field,
```
Antoinehoff, line 103: I don't understand why both `gk_field_fem_projection_par_rho_ts_2x` and `gk_field_fem_projection_par_phi_ts_2x` exist, it seems that you can use only one generic `gk_field_fem_projection_par_ts_2x` of them and pass the field you want to project.

manauref: i thought about this. I wanted it too. The problem is that this function essentially gets called by the gyrokinetic app, not the field app, so the gyrokinetic app would have to know which updater to call, and that doesn't follow the layering paradigm in the code.
If you think of a way, just put it in a branch and let me know.

Antoinehoff: Yeah I see the issue. A good solution in terms of code duplication is to create a local function called `gk_field_fem_projection_par_ts_2x` that contains the algo that projects and ts. Then call it with either `gk_field_fem_projection_par_rho_ts_2x` or `gk_field_fem_projection_par_phi_ts_2x` to pass the field you want. This would avoid duplicating the algo. You could even make it a function pointer an switch for 2x vs 3x case maybe

manauref: yeah but it'd create a lot of functions

Antoinehoff: this is fine, at least the receipe is the same, it just creates one function more.

PR #1007, opened 2026-05-08
NCCL reorganization for NCCL 2.19+ including modern NCCL implementations with new group semantics and changes to what config.blocking seems to do

JunoRavin:
Many of our production simulations on certain clusters, such as Perlmutter, are utilizing NCCL 2.18. The most recent version of NCCL is 2.29 and certain clusters such as Delta (machine files added in this PR) only have more modern NCCL implementations such as NCCL 2.25. 

To my understanding: In NCCL 2.18 and earlier, the library provided implicit communicator-level serialization. Operations submitted to a communicator were internally queued and matched across ranks, even across separate `ncclGroupStart`/`ncclGroupEnd` boundaries. This meant that:

- Per-direction groups (separate group start/end for each edge direction) worked because NCCL would internally match send/recv pairs across group boundaries.
- Ranks that raced ahead to the next group call would have their operations queued behind pending operations from other ranks.
- Empty groups (ncclGroupStart + ncclGroupEnd with no operations) were effectively no-ops.

Now, this implicit cross-group serialization is gone in NCCL 2.19+ and further, the config.blocking flag we were setting was causing all ncclGroupEnd calls to return ncclInProgress (and even collective calls like allreduce and broadcast were returning ncclInProgress). If ncclInProgress is returned, we require `ncclCommGetAsyncError()` until `ncclSuccess` is returned and this call and wait is **expensive**. 

We have thus rewritten significant chunks of nccl_comm.c to do the following:

- We have consolidated groups in `array_per_sync`. All multi-rank periodic send/recv operations are batched into a single `ncclGroupStart`/`ncclGroupEnd` call. Self-periodic cases (same-rank copies) are handled before the group with local buffer copies only. With per-direction groups, a send on rank A's lower-x group and the matching recv on rank B's upper-x group could be split across different groups, and ranks deadlock waiting for matches that never arrive within the same group in the new NCCL 2.19+ group semantics. We have also introduced an empty-group guard; the consolidated group is itself wrapped in an if (has_nccl_ops) check so that interior ranks with no periodic neighbors skip ncclGroupStart/ncclGroupEnd entirely. This avoids a segfault that NCCL 2.25 produces when one rank posts an empty group on the same communicator that other ranks have actual operations on. 

- Because CUDA operations and NCCL operations share the same stream, we have moved CUDA operations outside of the ncclGroupStart and ncclGroupEnd calls. All buffer preparation (range computation, resize, `copy_to_buffer`) happens in Phase 1 before `ncclGroupStart`. Only `ncclRecv`/`ncclSend` calls appear between `ncclGroupStart` and `ncclGroupEnd`. Buffer readout (`copy_from_buffer`) happens after `cudaStreamSynchronize`. This change is due to a note in the documentation "Caution: When called inside a group, stream operations (like ncclAllReduce) can return without having enqueued the operation on the stream. Stream operations like cudaStreamSynchronize can therefore be called only after ncclGroupEnd returns." from https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/groups.html 

While it may be unnecessary to move all operations which use the stream outside of ncclGroupStart and ncclGroupEnd, it does make the structure of the communication transparent and avoids Nvidia hardening any other stream and group semantics on us in the future. 

- We have introduced a proper barrier on the MPI communicator inside the NCCL communicator. Because of these stricter group semantics and comments like "Although NCCL group allows different operations to be issued in one shot, users still need to guarantee the same issuing order of the operations among different GPUs no matter whether the operations are issued to the same or different communicators." in the current documentation (https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/groups.html), we wish to properly barrier in between the `array_per_sync` and `array_sync` (since interior ranks return early from `array_per_sync` and can race ahead to `array_sync` and submit new NCCL group operations on the communicator while edge ranks are still executing their `per_sync` group). 

- Switch config.blocking=1 for performance reasons and delete all calls to `ncclCommSetAsyncError()` to avoid these issues associated with just how expensive this call is. 

**Unit tests pass and Vlasov multi-GPU tests are running fine, but I am tagging in the GK team for testing on Perlmutter and other multi-GPU machines to make sure production GK simulations are unaffected by these changes (and perhaps you can revisit running on Perlmutter with NCCL 2.29).**


Antoinehoff:
I merge this into the sheath AI branch and here's a comparison between a restart before and after. This is done on Perlmutter, interactive node, 4 GPUs. The command is
```
srun -u -n 4 ./gkeyll -g -M -e 4 -r 100 -s 100
```
ran in `/pscratch/sd/a/ah1032/gkeyll_sheath_ai/tcv_pt_0.5mw/wk` and the modules are
```
Currently Loaded Modules:
  1) conda/Miniforge3-25.11.0-1       4) craype/2.7.35    (c)   7) craype-network-ofi  (cpe)     10) cudatoolkit/13.0      (g)    13) nccl/2.29.2-cu13 (g)
  2) python/3.13-26.1.0               5) cray-dsmml/0.3.1       8) cray-libsci/25.09.0           11) craype-accel-nvidia80 (cpe)
  3) gcc-native/14              (c)   6) libfabric/1.22.0       9) PrgEnv-gnu/8.6.0    (prgenv)  12) cray-mpich/9.0.1      (mpi)
```

**Before**
```
Restarting from frame 100 at time = 0.000200004
Starting main loop ...
Taking time-step at t = 200.004 mus ... dt = 0.00465438 mus
Taking time-step at t = 200.465 mus ... dt = 0.00465524 mus
 ... finished

Number of update calls 100
Number of forward-Euler calls 300
Number of write calls 4
Timing:
  - Time loop:                         2.4015e+01 sec.
```

**With the NCCL fix**
```
Restarting from frame 100 at time = 0.000200009
Starting main loop ...
Taking time-step at t = 200.009 mus ... dt = 0.00465358 mus
Taking time-step at t = 200.47 mus ... dt = 0.00465527 mus
 ... finished

Number of update calls 100
Number of forward-Euler calls 300
Number of write calls 4
Timing:
  - Time loop:                         2.3864e+01 sec.
```

Not clear if there is a performance gain. However we see that the dt are not exactly the same...

JunoRavin:
@Antoinehoff is there a reason the restart time is different in both cases? Presumably the data is the same from before and after the restart, but the restart time itself is different and that along could explain a small variation in the size of the time step.

Antoinehoff:
> @Antoinehoff is there a reason the restart time is different in both cases? Presumably the data is the same from before and after the restart, but the restart time itself is different and that along could explain a small variation in the size of the time step.

I don't understand why the restart does not occur at the same time 🤔

JunoRavin:
> > @Antoinehoff is there a reason the restart time is different in both cases? Presumably the data is the same from before and after the restart, but the restart time itself is different and that along could explain a small variation in the size of the time step.
> 
> I don't understand why the restart does not occur at the same time 🤔

Are these two different simulations, or are you restarting the same simulation two different times? We don't precisely write out data at time t = t_end (so if the turbulence has some small variation in the potential that leads to a small change to the time step, you could write out the final frame at 200.005 versus 200.009 or something like that).

Antoinehoff:
Both runs are restarts from the exact same frame. I would have assumed that the time displayed in 
```
Restarting from frame 100 at time = 0.000200009
```
was the time of frame which should not change upon a restart right?

JunoRavin:
> Both runs are restarts from the exact same frame. I would have assumed that the time displayed in
> 
> ```
> Restarting from frame 100 at time = 0.000200009
> ```
> 
> was the time of frame which should not change upon a restart right?

Okay, that seems like something with the actual restart functionality then or what's stored via I/O, because that time there is supposed to be read from the meta-data in the output file, and I don't know what the time there would be different restart to restart.

manauref:
I'll try a multiblock regression test.

It'd be good to check if the
```
mctest_multib_allgather
mctest_multib_sync
```
unit tests still work.

manauref (approved):
Approving, but @JunoRavin I think you should add a job script for Delta showing what's the best set way to run sims on that machine

manauref:
One additional comment. NERSC support said they do not recommend using the modules we currently use in our Perlmutter machine files because they don't yet have an mpich module compatible with cuda13. They recommend
cuda 12.9, nccl/2.24.3, cray-mpich/9.0.1
however, my testing with the new flags seems to suggest that the modules our machine files have, with the new env flags, work for us. So we'll stick with that. If problems arise we can downgrade to cuda 12.9

Antoinehoff:
Will test it now and come back shortly!

Antoinehoff:
@manauref did you try a multi node GPU run on Perlmutter? I tried a  TCV clopen 3x2v with 8 GPUs and it is stuck at `Creating app object` phase.

Maxwell-Rosen:
Please check multi-node configurations as well.

manauref:
@Antoinehoff make sure your multi-node jobscript looks like the one in machines/

Antoinehoff:
It worked! Thanks @manauref. Did you try to run partially full nodes? I tried a 6 GPUs run and it is blocked.

manauref:
I didn't try partially full nodes. What command did you try to do that with? it's likely that we need to get crafty with both the SLURM and srun options. We should put an example of both full and partially full nodes in the machines/ folder

Antoinehoff:
I just changed the following lines:
```
#SBATCH --tasks-per-node=3
#SBATCH --gpus 6
```
and the command to
```
srun -u -n 6 ./gkeyll -g -M -c 1 -d 1 -e 6
```

Antoinehoff (approved):
Thanks a lot for working on this @JunoRavin and @manauref !

PR #1009, opened 2026-05-08
BUG: there are different numbers of quadrature points and basis functions

### LINKED ISSUE #1008: Mismatch of number of quadrature nodes and number of basis functions
Right here there is a serious mistake.

https://github.com/ammarhakim/gkeyll/blob/b31c61f1f3b1a63086ef3b0548fc47a9ac6a0861/core/zero/nodal_ops.c#L361

It says the number of quadrature points is the same as the number of basis functions. This is simply not true! In branch `positivity-by-supressing-dfdt`, I have addressed this by adding a flag to the basis for the number of quadrature points. This is neccisary to loop over all quadrature points. In other parts of the code, the number of quadrature points is hardcoded irrespective of the basis type. This must be fixed

Maxwell-Rosen:
Luckily, this was not a widely used function, and its only use cases were coincidentally cases where num_basis==num_quad (p1), but this is not true in general. I went through all the `modal_to_quad` functions and counted their cases, setting the number of cases to the number of quadrature points for that basis set. Importantly, `modal_to_quad` takes a modal expansion and computes the values at quadrature points. It takes an index, which is the index of the quadrature point. To loop over the number of quadrature points, we must know how many there are.

Fixes #1008

### DIFF
--- a/core/zero/array_dg_reduce.c
+++ b/core/zero/array_dg_reduce.c
@@ -28,7 +28,7 @@ gkyl_array_dg_reducec(double *out, const struct gkyl_array *arr, int comp,
 
   double *arr_d = arr->data;
 
-  int num_nodes = pow(basis->poly_order+1,basis->ndim);
+  int num_nodes = basis->num_quad;
 
   switch (op) {
     case GKYL_MIN:
@@ -94,7 +94,7 @@ gkyl_array_dg_reducec_range(double *out, const struct gkyl_array *arr, int comp,
 
   double *arr_d = arr->data;
 
-  int num_nodes = pow(basis->poly_order+1,basis->ndim);
+  int num_nodes = basis->num_quad;
 
   struct gkyl_range_iter iter;
   gkyl_range_iter_init(&iter, range);
--- a/core/zero/array_dg_reduce_cu.cu
+++ b/core/zero/array_dg_reduce_cu.cu
@@ -60,7 +60,7 @@ dg_arrayMax_blockRedAtomic_cub(const struct gkyl_array* inp, double* out, int co
 
   long nCells = inp->size;
   size_t nComp = inp->ncomp;
-  int num_nodes = pow(basis->poly_order+1,basis->ndim);
+  int num_nodes = basis->num_quad;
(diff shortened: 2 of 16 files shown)


JunoRavin (approved):

ammarhakim (approved):
Ready to merge

manauref:
A question:
here
Screenshot 2026-05-10 at 8 56 07 AM
you are pointing `num_nodes` to `num_quad`. But which "nodes" is this referring to? basis nodes or quad nodes? Did you check all instances of num_nodes to make sure it is referring to number of quad nodes?

At some point, or progressively, we should make "nodes" more specific throughout the code, to differentiate between
- positivity nodes
- basis nodes
- gauss-legendre nodes
- gauss-lobatto nodes

JunoRavin:
> A question: here Screenshot 2026-05-10 at 8 56 07 AM you are pointing `num_nodes` to `num_quad`. But which "nodes" is this referring to? basis nodes or quad nodes? Did you check all instances of num_nodes to make sure it is referring to number of quad nodes?
> 
> At some point, or progressively, we should make "nodes" more specific throughout the code, to differentiate between
> 
> * positivity nodes
> * basis nodes
> * gauss-legendre nodes
> * gauss-lobatto nodes

All of the instances @Maxwell-Rosen changed were in the context of calls to modal_to_quad_nodal from my review, so I didn't see any ambiguity on which nodes are being referred to here.

Maxwell-Rosen:
I agree. We should even add something to evaluate modal expansions at corners, which are a different kind of node. In general, if we expand the scope of the evaluations that the basis object can do, it will lower the threshold for someone to write kernels (and I really don’t like writing kernels).

JunoRavin:
You're referring to the Gauss-Lobatto nodes, which are already contained in the basis object: https://github.com/ammarhakim/gkeyll/blob/main/core/ker/basis/basis_node_coords_ser.c

the only subtlety is that the Serendipity basis does not define the complete Gauss-Lobatto basis, which is why @manauref refers to the differences between basis nodes and Gauss-Lobatto nodes, but these are the same thing for p=1 and all tensor bases with p>1.

Issue #656, opened 2025-03-20
[DR] New interface for species diagnostics


manauref:
We propose to organize the species diagnostics into a cohesive table within the species tables of input files. Presently the GK input files allow specifying species diagnostics as follows (see e.g. rt_gk_sheath_2x2v_p1):
```
  struct gkyl_gyrokinetic_species elc = {
...
    .num_diag_moments = 5,
    .diag_moments = { "M0", "M1", "M2", "M2par", "M2perp" },
    .num_integrated_diag_moments = 1,
    .integrated_diag_moments = { "HamiltonianMoments" },
    .time_rate_diagnostics = true,

    .boundary_flux_diagnostics = {
      .num_integrated_diag_moments = 1,
      .integrated_diag_moments = { "HamiltonianMoments" },
//      .time_integrated = true,
    },
...
  },
```
As we can see the loose options (not within a logically constrained table) are accumulating. We propose to organize them as follows:

```
  struct gkyl_gyrokinetic_species elc = {
...
    .diagnostics = {
      .num_diag_moments = 5,
      .diag_moments = { "M0", "M1", "M2", "M2par", "M2perp" },
      .num_integrated_diag_moments = 1,
      .integrated_diag_moments = { "HamiltonianMoments" },
      .time_rate_diagnostics = true,

      .boundary_flux = {
        .num_integrated_diag_moments = 1,
        .integrated_diag_moments = { "HamiltonianMoments" },
//        .time_integrated = true,
      },
    },
...
  },
```
simply by putting the existent options into a `.diagnostics` table, which could be a struct such as the existent
```
struct gkyl_phase_diagnostics_inp {
  int num_diag_moments; // Number of diagnostic moments.
  char diag_moments[24][24]; // List of diagnostic moments.
  int num_integrated_diag_moments; // Number of integrated diagnostic moments.
  char integrated_diag_moments[24][24]; // List of integrated diagnostic moments.
  bool time_integrated; // Whether to use time integrated diags.
};
```
but enriched with another `gkyl_phase_diagnostics_inp boundary_flux` to pass the boundary flux diagnostics arguments.

**The main benefit** of this coalescence is that modifications to diagnostic API will be constrained to modifications of the `gkyl_phase_diagnostics_inp` struct, and no modifications to the species input struct may be needed. Possible modifications we could conceivably specify include:
a) Specifying the LCFS so boundary flux diagnostics are only computed in the SOL:
```
    .boundary_flux = {
      ...
[...]

PR #1013, opened 2026-05-13
Add cross-collisions diagnostics to the GK LBO app

Maxwell-Rosen:
# Feature

This PR is nearly 100% Claude, but it's not AI slop. I verify that this implementation is how I would do it, also.

## Summary

Purpose: My mirror simulations are exposing an issue with the implementation of cross-species collisions. The moments can go negative if $m_i \gg m_e$, $T_e \ll T_i$, $u_{\parallel,i} \ll u_{\parallel,e}$. This is just in the mathematics of how these cross-moments are computed. By outputting detailed cross-species collision diagnostics, we can effectively diagnose such issues.

## Implementation Details

Key changes: gk_species_lbo.c is expanded to include a new diagnostic for the cross-species collisions. Since these diagnostics are on intermediate quantities, they must be re-computed, so I refactored the update function as well.

I decided to reuse some memory from the prim_moms_ho and nu_sum_ho to save some memory and avoid additional allocations.

I could see someone arguing that this should just be under the `write_diagnostics` flag and not a separate flag. It's good to keep this modular.

Automated testing: Existing tests

## Example Use

```
    .collisions =  {
      .collision_id = GKYL_LBO_COLLISIONS,
      .den_ref = ctx.n0,
      .temp_ref = ctx.Ti0,
      .num_cross_collisions = 1,
      .collide_with = { "elc" },
      .write_diagnostics = true,
    },
```

## Community Standards

## Testing: 

- [ ] I ran a few regression tests to ensure no apparent errors.


gyrokinetic/apps/gkyl_gyrokinetic.h
```diff
@@ -106,6 +106,7 @@ struct gkyl_gyrokinetic_collisionless {
 struct gkyl_gyrokinetic_collisions {
   enum gkyl_collision_id collision_id; // type of collisions (see gkyl_eqn_type.h)
   bool write_diagnostics; // Whether to output diagnostics.
+  bool write_cross_diagnostics; // Whether to output per-cross-species diagnostics (prim_moms_sr, nu_sr).
```
manauref: I don't think we should ad more `write_diagnostics`-type flags. Cross diags should just be part of `write_diagnostics`.

In the future we'll just replace `write_diagnostics` with a table that offers more flexibility, see issue #656

manauref (changes requested):

gyrokinetic/apps/gkyl_gyrokinetic.h
Maxwell-Rosen: I merged the two flags together. I implemented it as two seperate flags because it exposes different levels of diagnostics, but I agree it should just be one flag. I was unsure about the implementation when I first did it

Maxwell-Rosen:
Continued in #1076

Issue #1015, opened 2026-05-18
Something broken in the CPU GK app in main


Antoinehoff:
A core only GK 2x2v simulation setup is showing **strongly disagreeing results between CPU and GPU** in [main]((e40fdaf0220b046c4101431e5f18bffc0f4799fa) ). Following the issue #1012, running the same test in an old main version (early February [5506bb](https://github.com/ammarhakim/gkeyll/commit/5506bbe9d39a48cc81da9b132be9b8b400ccb213)) does not display the discrepancy and show the same result between GPU and CPU (which is the result of `main` GPU).

Input file of the test: [rt_gk_cbc_2x2v_p1.c](https://github.com/user-attachments/files/27969775/rt_gk_cbc_2x2v_p1.c)

`main` CPU 

<img width="477" height="328" alt="Image" src="https://github.com/user-attachments/assets/5f292f78-f29a-4c3d-a160-7a9283b8dfd0" />

`main` GPU

<img width="479" height="328" alt="Image" src="https://github.com/user-attachments/assets/cc4e6d9d-e44d-4d7f-9920-ef376e9a92e7" />

`5506bb` CPU

<img width="479" height="328" alt="Image" src="https://github.com/user-attachments/assets/d02e15e3-ca1d-4ae2-bc87-7ccd39f02a51" />

`5506bb` GPU

<img width="479" height="328" alt="Image" src="https://github.com/user-attachments/assets/d714097e-41fa-4882-a6c8-2ba9f683d90c" />

The CPU `main` result is **wrong** because we expect to have GAMs appearing in this non canonical Maxwellian initialization of a global CBC.

Antoinehoff:
Using `git bisect` we find:
```
338333682368e9edc316c287e4fb4bdf58255106 is the first bad commit
commit 338333682368e9edc316c287e4fb4bdf58255106 (HEAD)
Author: manauref <mfrancisquez@gmail.com>
Date:   Tue Mar 17 13:57:19 2026 -0700

[...]

PR #1016, opened 2026-05-18
Superlu non equil cpu fix

Antoinehoff:
This fixes the incoherent results between GPU and CPU presented in issue #1015. 

**CPU result = GPU result with this branch**

We deactivate the `Equil` option of the `dgssvx` SuperLU solver to avoid any further modification of the matrix in the Poisson solver. This bug must have sneaked in because it does not show up in conservation issues and we tested the branch on GPU most of the time. 

### Enhance unit test robustness
Also the unit tests do not loop over the solver so this issue could not be detected.
To address this we loop twice over each `set_rhs`+`solve` pair in the `ctest_fem_poisson_perp.c` unit test now. **This new version of the unit test fails 17/19 tests in `main` and passes successfully on this branch.**

### Performance comparison
We check if changing the SuperLU option affects performance (CPU code only) by comparing the time it takes to perform 5 steps of the TCV 3x2v regression test with increased resolution (18x8x8x8x8).

We see that the performance varies but the physics too, the number of calls is not the same since this branch provides a more dynamic electrostatic field that asks to reevaluate the time step. In proportion it does not look like the field solve has an increased cost.

`main`
```
ah1032@nid006593:/pscratch/sd/a/ah1032/gkeyll_main/rt_gk_tcv_iwl_adapt_src_3x2v> srun -n 4 -u ./rt_gk_tcv_iwl_adapt_source_3x2v_p1 -M -e 4 -s5
        Step 2 at time 2.0517872e-08.  Time-step  1.025911e-08.  Completed 10%
        Step 3 at time 3.0777800e-08.  Time-step  1.025993e-08.  Completed 20%
        Step 4 at time 4.1038838e-08.  Time-step  1.026104e-08.  Completed 30%
        Step 5 at time 5.1301211e-08.  Time-step  1.026237e-08.  Completed 40%

Number of update calls 5
Number of forward-Euler calls 15
Number of write calls 4
Timing:
  - Time loop:                         7.3666e+00 sec.
    * Forward Euler:                   5.8688e+00 sec. / 79.67 %.
...
    * Field solves:                    6.2203e-02 sec. / 0.84 %.
      ^ Phi eqn RHS:                   1.9048e-02 sec. / 30.62 %.
      ^ Phi eqn solve:                 4.3130e-02 sec. / 69.34 %.
```

this branch
```
ah1032@nid006593:/pscratch/sd/a/ah1032/gkeyll_main/rt_gk_tcv_iwl_adapt_src_3x2v> srun -n 4 -u ./rt_gk_tcv_iwl_adapt_source_3x2v_p1 -M -e 4 -s5
        Step 2 at time 2.0536559e-08.  Time-step  1.027779e-08.  Completed 10%
        Step 3 at time 3.0828468e-08.  Time-step  1.029191e-08.  Completed 20%
        Step 4 at time 4.0032197e-08.  Time-step  9.203729e-09.  Completed 30%
        Step 5 at time 4.7943316e-08.  Time-step  7.911118e-09.  Completed 40%

Number of update calls 5
Number of forward-Euler calls 19
Number of RK stage-2 failures 2
  Max rel dt diff for RK stage-2 failures 0.307969
  Min rel dt diff for RK stage-2 failures 0.0990775
Number of write calls 4
Timing:
  - Time loop:                         9.6846e+00 sec.
    * Forward Euler:                   7.6183e+00 sec. / 78.66 %.
 ...
    * Field solves:                    7.9741e-02 sec. / 0.82 %.
      ^ Phi eqn RHS:                   2.9936e-02 sec. / 37.54 %.
      ^ Phi eqn solve:                 4.9782e-02 sec. / 62.43 %.
```

## Consequences of the bug fix
It seems that the conservation properties that we tested with the poisson perp solve in #928 are not valid anymore with that fix. Many of the tests were run with a non-physical phi that was mainly stabilizing exponentially. Here is an example of before and after particle conservation test with the helical 3x2v test on 1 CPU:

before the fix 

Screenshot 2026-05-18 at 6 42 14 PM

after the fix 

Screenshot 2026-05-18 at 6 41 51 PM


Antoinehoff:
I miss clicked, I did not want Copilot to review -_- go away shoo!

manauref (changes requested):
1. I think this reg test doesn't belong here.
2. Can you add a test in ctest_fem_poisson_perp, that simply runs one of the existing solves in a loop, and checks the results every time?
3. I think you should double check the performance hit, if any, of this change. These seemingly benign changes can sometimes worsen or improve performance by large factors. Maybe just running rt_gk_tct_iwl_3x2v with larger res is sufficient.

Antoinehoff:
@manauref 
1. yes I saw it
3. In the CBC 2x2v test I used, the performance did not change a iota. I don't think anyone wants to run a 3x2v TCV simulation on CPU but I can try... The thing is, this may improve the performance since it is skipping a preconditioning step. It may however break the solver if the original problem has a nasty condition number. 
2. working on it...

manauref (approved):

JunoRavin (approved):

PR #1019, opened 2026-05-19
Core-only GK simulations and deprecation of ES_IWL field id

manauref:
This cleans up the prototyping branch poisson_perp_bias-core_only (see [DR in issue #1003](https://github.com/ammarhakim/gkeyll/issues/1003)).

This enables running core-only simulations specifying TWISTSHIFT as the parallel BC of the species. See reg tests
rt_gk_tcv_core_3x2v_p1
which is based on just taking the SOL part of the TCV IWL test. Note however that I did this months ago, and the TCV IWL test may have changed by now. However I ran this simulation (with higher resolution) for several 100 microsec, into turbulence. I'll try to do it again and share plots.  @Antoinehoff noticed that this test uses Neumann BCs for phi at the inner radial boundary. I don't have a good reason for this, and perhaps the BC should be Dirichlet. Neumann is just something I tried months ago that behaved better back then, but we should re-examine.

There's also the
rt_gk_cbc_3x2v_p1
test.

Future work
------------

Piping the numerical shift into the twistshift updater so we can do core-only with numerical equilibria.
This is done in the `mb_sol_3x-prototyping` branch


Antoinehoff:
So the TCV 3x2v regression test is 2 GPUs compute sanitizer clean and CPU valgrind clean on the current state of this branch.

Antoinehoff (approved):

tnbernard (approved):

Issue #1024, opened 2026-05-21
[DR] Descriptions in metadata


Maxwell-Rosen:
I've been frustrated by this for a long time. Our file outputs do not have good descriptions, and we should have better documentation about the files we save. I often get confused about what is written in a file (for instance, is it the distribution function f or JBf in GK, or is BiMaxwellianMoments vt^2 or temperature. I know these by now, but new users will be very confused.

We should aim to give the units of the files, spelled out in words, or just given in SI units. See the example below (I think since J [m^3], B [T], and f [1/m^3], that our written out ion_0.gkyl is therefore units [T] in SI).

Here is an example of how we can add descriptions to written files using the metadata.

```C
static void
gk_species_write_dynamic(gkyl_gyrokinetic_app* app, struct gk_species *gks, double tm, int frame)
{
  struct timespec wst = gkyl_wall_clock();
  gks->write_cfl_func(app, gks, tm, frame);

  // Metadata from app and geo object.
  gkyl_msgpack_map_elem_set_double(app->io_meta_basic_len, app->io_meta_basic, "time", tm);
  gkyl_msgpack_map_elem_set_uint(app->io_meta_basic_len, app->io_meta_basic, "frame", frame);
  struct gkyl_msgpack_map_elem io_meta_f[] = {
    { .key = "Description", .elem_type = GKYL_MP_STRING,
      .cval = "J*B*f: distribution function (f) scaled by phase-space Jacobian (J) and magnetic field (B). Units are particles per distance cubed (SI: T)" }
  };
  int io_meta_f_len = sizeof(io_meta_f)/sizeof(io_meta_f[0]);
  int io_meta_len[] = {app->io_meta_basic_len, gks->io_meta_len, app->gk_geom->io_meta_len, io_meta_f_len};
  const struct gkyl_msgpack_map_elem* io_meta[] = {app->io_meta_basic, gks->io_meta, app->gk_geom->io_meta, io_meta_f};
  struct gkyl_msgpack_data *mt = gkyl_msgpack_create_union(sizeof(io_meta_len)/sizeof(int), io_meta_len, io_meta);

  const char *fmt = "%s-%s_%d.gkyl";
  int sz = gkyl_calc_strlen(fmt, app->name, gks->info.name, frame);
  char fileNm[sz+1]; // ensures no buffer overflow
  snprintf(fileNm, sizeof fileNm, fmt, app->name, gks->info.name, frame);
  
  // copy data from device to host before writing it out
[...]

PR #1025, opened 2026-05-21
Descriptions in metadata

Maxwell-Rosen:
See design review #1024 

I added descriptions to some, not all, of the diagnostics in gyrokinetics. I encourage others to add descriptions to the parts of the code they are working on for diagnostics, so others can easily understand the outputs.

I added descriptions to the distribution function, its moments, geometry, and the sources. These descriptions are mostly LLM-generated (Claude Opus 4.7), but I'm pretty sure they are correct. I've gone through them for accuracy. If you notice anything that should be amended, please feel free to push directly to this branch.

The purpose of this PR is to set an example for how to add descriptions in fields. I'm not aiming to add descriptions to all fields, as this exceeds my understanding (which is exactly why descriptions are needed!).

One important decision I made is to make the descriptions GK specific here. This is to avoid confusion in the definitions of velocity and geometry across modules. Particularly, I added the descriptions to the moments (M0, M1, etc) into GK specifically, although these are core/ updaters that it's refrencing. Although other modules could benifit from this, it's important to keep this GK specific.

Claude code was used to generate most of the code in this PR, using human review to ensure accuracy.

Example output from postgkeyll
```
pgkyl rt_gk_mirror_kinetic_elc_1x2v_p1-ion_source_M0_0.gkyl info
(default#0)
├─ Time: 0.000000e+00
├─ Frame: 0
├─ Number of components: 2
├─ Number of dimensions: 1
├─ Grid: (uniform)
│  └─ Dim 0: Num. cells: 32; Lower: -2.515312e+00; Upper: 2.515312e+00
├─ Maximum: 8.538827e+22 at (16,) component 0
├─ Minimum: -1.012117e+22 at (17,) component 1
├─ DG info:
│  ├─ Polynomial Order: 1
├─ Created with Gkeyll:
│  ├─ Changeset: e6c85a23574a
│  └─ Build Date: 2026-05-21 11:34:18 EDT
├─ Geometry info:
├─ Description: M0 = int S d^3v: source density flux
```


manauref:
This is fine, but I think it needs two modifications:
1. I think the description should be short, and include only words. Math may not render correctly in some places, and some math may be too long/complicated to include. So you'll end up inconsistently having some quantities with math, and some without math. So best keep the math out, and the descriptions short.
2. If you want to point to something that has the math, we can use my gkeyll_gk_ documents that explain some of these things and put them in the code, citing the file names.
3. The description shouldn't have units. Our documentation (2 above) should say what the units are. Also, sometimes we do unitless simulations. The units are decided by the input file. I think the correct way to do this is for every outputted file to give a de-normalization scale factor. Postprocessing can then provide that factor, and multiply by it if the user wants. The thing about this I haven't figured out is how to pass these denorm facs to the various quantities written by the solver.
4. Add descriptions to every file written by the app, not just the few you added it to.

JunoRavin:
I think as an improvement in functionality to the code, this is good. I’m in favor of using MP_STRING to add text descriptions and I’m not opposed to a LLM doing the bit of tedious plumbing to add this metadata where appropriate.

However, I strongly agree with Mana on the math rendering front. I’ve been meaning to experiment with a skill to make the LLMs not do this because I think trying to render the math with ASCII is a recipe for intense confusion and their default is to always do this. And I also strongly agree that we should not have units because across the group we do so many different types of simulations with normalized vs unnormalized units that I don’t think the units additions help.

Fundamentally you were motivated to do this work because of clarification on Jacobians vs no Jacobians in outputs. Defining a minimal set of information to include is the right call. Plus you’re likely looking ahead to how this information is usable by a LLM for post processing. Adding clarity about Jacobian factors while not adding information that could lead to context poisoning (if the LLM gets confused about units because you ran a normalized unit simulation) is likely to be a net benefit.

Maxwell-Rosen:
> This is fine, but I think it needs two modifications:
> 
> 1. The description should be short and include only words. Math may not render correctly in some places, and some math may be too long/complicated to include. You’ll end up with some quantities using math and others not. Please keep the math out and the descriptions short.
> 2. If you want to point to something with the math, we can use my gkeyll_gk_ document, which explains some of these things and puts them in the code, citing the file names.

The math here is Unicode, which will render in the terminal on all machines. The math makes it clear what these terms are. The math is clear.

> 3. The description shouldn’t have units. Our documentation (2 above) should specify the units. Also, we sometimes run unitless simulations. The input file decides the units. The correct way to do this is for every output file to give a de-normalization scale factor. Postprocessing can then provide that factor, and multiply by it if the user wants. I haven’t figured out how to pass these denorm facs to the various quantities the solver writes.

The units are debatable. Let's give an example in SI units; if you’re using something else (cgs, normalized), you can understand the units relative to the SI coordinates (if you’re in normalized coordinates, nothing has units anyway, so it’s not even a fair criticism). We can provide SI, CGS, and normalized (blank) units for comparison, but units are important for understanding the outputs. Plots without units don’t make sense.

> 4. Add descriptions to every file the app writes, not just the few you added it to.

This is just an example of how to add descriptions to the most important files. I commented on this in the PR itself. Please add descriptions to your own outputs. I don’t understand everything; that is why descriptions are needed in the first place!

Maxwell-Rosen:
I can remove the units and change the math characters. I think it looks much nicer with the math, but it's definitely more universal to use the characters on a normal keyboard.

akashukla:
I like the math and the units. I just think one should be very careful that they are actually correct for every single case. Particularly the ones that AI came up with. I see my comment on the DR (it says units of JBf is Tesla)

manauref:
I do like this more and more. But I think it should be used as follows:
1. Descriptions should be short and include only/mostly English.
2. No math in metadata. We should create an overleaf where we define every diagnostic using math, and we link to it in our docs. I'll create the overleaf and ensure it remains up to date going forward.
3. It should be done to every file written out.

This will also play nicely with some future additions to postgkyl which expect certain names for GK diagnostics ( ;) @Antoinehoff )

manauref:
I'm absorbing this into a slightly larger DR/PR, and closing this.

PR #1026, opened 2026-05-22
Remove IWL enums (from input file)

manauref:
Remove the IWL species and field enums from input files and from the app. Clopen is a property of the geometry, and it's been cumbersome to have it given to the app through the species BC. Here we fix that.

Now, for clopen IWL calculations, one should:
- Specify the SOL BC in the species BC table.
- Pass the LCFS location to the geometry table (we already did that).
- For 3x: pass the twistshift shift via the geometry table.

The app will know to apply periodicity (2x) or twistshift (3x) in the core.

Similar treatment should be followed in input files with numerical geometry. This new approach also made it a bit easier to write out the shift in both cases.

### DIFF
--- a/gyrokinetic/apps/gk_species_projection.c
+++ b/gyrokinetic/apps/gk_species_projection.c
@@ -359,8 +359,8 @@ init_maxwellian_gaussian(struct gkyl_gyrokinetic_app *app, struct gk_species *s,
   struct gkyl_gyrokinetic_bc *bc_up = gk_fetch_bc_with_dir_edge(s->info.bcs, 2*app->cdim, app->cdim-1, GKYL_UPPER_EDGE);
   if (bc_lo != 0 && bc_up != 0) {
     // Apply periodicity condition if both edges are IWL.
-    fg_ctx.is_dir_periodic[app->cdim-1] = (bc_lo->type == GKYL_BC_GK_SPECIES_IWL && bc_up->type == GKYL_BC_GK_SPECIES_IWL)
-      || (bc_lo->type == GKYL_BC_GK_SPECIES_TWISTSHIFT && bc_up->type == GKYL_BC_GK_SPECIES_TWISTSHIFT);
+    fg_ctx.is_dir_periodic[app->cdim-1] = app->gk_geom->has_LCFS ||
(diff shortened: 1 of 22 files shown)


gyrokinetic/apps/gk_species_projection.c
```diff
@@ -359,8 +359,8 @@ init_maxwellian_gaussian(struct gkyl_gyrokinetic_app *app, struct gk_species *s,
   struct gkyl_gyrokinetic_bc *bc_up = gk_fetch_bc_with_dir_edge(s->info.bcs, 2*app->cdim, app->cdim-1, GKYL_UPPER_EDGE);
   if (bc_lo != 0 && bc_up != 0) {
     // Apply periodicity condition if both edges are IWL.
-    fg_ctx.is_dir_periodic[app->cdim-1] = (bc_lo->type == GKYL_BC_GK_SPECIES_IWL && bc_up->type == GKYL_BC_GK_SPECIES_IWL)
-      || (bc_lo->type == GKYL_BC_GK_SPECIES_TWISTSHIFT && bc_up->type == GKYL_BC_GK_SPECIES_TWISTSHIFT);
+    fg_ctx.is_dir_periodic[app->cdim-1] = app->gk_geom->has_LCFS ||
+      (bc_lo->type == GKYL_BC_GK_SPECIES_TWISTSHIFT && bc_up->type == GKYL_BC_GK_SPECIES_TWISTSHIFT);
```
Antoinehoff, line 363: Should we remove this line and all `GKYL_BC_GK_SPECIES_TWISTSHIFT` enum as well? One could imagine that TSBC is just a special periodicity and handle it fully through the geom table instead of keeping it in the Species.
In the geom table we could have
```C
.num_periodic_dim = 1,
.periodic_dim = {1},
.num_twist_shift_dim = 1,
.twist_dim = {1},
```
What do you think?

manauref: I thought about this, but I'm still unsure about the best way to handle this, so I opted to leave it as is.
- On the one hand it's true that TS is a special type of periodicity, and that treating it like we treat periodicity seems reasonable (then the app automatically applies TS to whatever quantities it has to).
- However, that would remove TS from the species BC table, which seemed odd to me (especially for core-only).

Dunno, maybe i'm getting convinced that we should do it (add the `twist_dim` option).

Issue #1022, opened 2026-05-20
Static source BGK operator


Antoinehoff:
For analytical studies like cyclone base case (CBC), it can be useful to have BGK source operator that are simply applying 
```math
S_{BGK} = -\nu_{eq} (x,y,z) \left[ f - f_{eq}(x,y,z,v_\parallel, \mu) \right]
```
where $\nu_eq$ and $f_{eq}$ are static relaxation frequency and equilibrium distribution function, respectively, provided by the user.

 The implementation of this source will leverage the work done with PR #992 on unification of heating source and external Eirene ionization source in `gk_species_source_bgk.c` and will mostly add routines and options to be able to write, e.g.,
```C
    .source_bgk = {
      .source_bgk_id = GKYL_SOURCE_BGK_STATIC,
      .rate_profile = bgk_source_rate_profile,
      .rate_profile_ctx = &ctx,
      .feq_shape = eval_canon_maxwellian_e,
      .feq_shape_ctx = &ctx,
    }
```
in the species table.

PR #1028, opened 2026-05-22
[PR] Static BGK source operator

Antoinehoff:
**Note:** it is unclear how we want to proceed with the merge of this because this branch also contains the core only modifications. I propose we merge first #1019 and then this one. 

This PR is related to DR #1022 

We implement here the static BGK source operator to relax the distribution function towards a user provided equilibrium at a user provided rate, i.e.
```math
S_{BGK} = -\nu_{eq} (x,y,z) \left[ f - f_{eq}(x,y,z,v_\parallel, \mu) \right]
```
where $\nu_eq$ and $f_{eq}$ are static relaxation frequency and equilibrium distribution function, respectively, provided by the user.

Below, we present some results obtained with the 2x2v Cyclone Base Case (CBC) regression test.

This test, without any source, shows the following evolution of the plasma profiles:

### Initial condition freeze 🥶 
We check first that it is possible to maintain the initial conditions. We set $\nu_eq = 10^6$ Hz and $f_{eq} = f_0$, $f_0$ being the initial condition for both electrons and ions.

_evolution of the profiles along x_

image

_evolution of the profiles along z_

image

We see that the system keeps the profiles constant thanks to the static BGK source, aside of a steady state potential and flow that develops in a time scale faster than the relaxation rate (<1$\mu$s). This is a consequences of a non canonical Maxwellian IC. At t=0, the potential and the flows are 0 and the density are equal everywhere. However, this system is not a steady state solution of the GK equation. Even with a static BGK source trying to maintain this state, the plasma finds an equilibrium with a non zero potential, flow and charge density and displays a geodesic acoustic mode (GAM).

We also see that we can control the damping of the oscillations by modifying the relaxation rate:
_left to right:_ $\nu_eq = 10^7, 5\times 10^6, 10^6$ Hz
Screenshot 2026-05-22 at 3 04 01 PM

### Buffer regions
We can now use the static BGK source to create buffer regions where the plasma should not vary. We use a setup is inspired from the setup found in V. Grandgirard et al. / Computer Physics Communications 207 (2016) 35–68.
The relaxation rate is now given a spatial dependency, following:
```math
\nu_{eq}= \nu_{eq,0}\, H(x)
```
where
```math
H(x) = 1 + \frac{1}{2} \left[ \tanh \left( \frac{x - x_{\max} + B_l L_x}{B_s L_x} \right) - \tanh \left( \frac{x - x_{\min} - B_l L_x}{B_s L_x} \right) \right]
```

Without the buffer spurious oscillations due to the unphysical Dirichlet BC applied to $\phi$ at the outer radial wall occur

image

With the buffer activated, we see that the fast scale oscillations are quickly damped, putting in evidence clean GAMs (with the correct frequency)

image

Screenshot 2026-05-22 at 3 16 52 PM

Screenshot 2026-05-22 at 3 17 31 PM

Screenshot 2026-05-22 at 3 18 27 PM

## Diagnostics
The static BGK source diagnostics consist in 
- The equilibrium function `[sim_prefix]-[ion/elc]_source_bgk_feq_0.gkyl`
- The relaxation rate `[sim_prefix]-[ion/elc]_source_bgk_rate_0.gkyl`
- The integrated primary moments of the total source term `[sim_prefix]-[ion/elc]_source_bgk_integrated_M0M1M2.gkyl`
Here is the evolution of these moments in a CBC simulation with buffer regions:

Screenshot 2026-05-23 at 11 42 12 AM

## Additional modifiations

This PR also refactor the `gk_species_source_bgk.c` in various ways:

- We now name the equilibrium function `feq` and not `fmax` as it can be used for any distribution function.
- We reduce the code duplication in initialization and release phase.
- We adapt the diagnostics to allow a diagnostic during the RHS computation for the static BGK source. This is required to get integrated moments of $\nu(f - f_{eq})$ without creating a new full phase space array.


dingyunl:
I don't know the full background of this work but this piece looks very solid to me. 

Something might be unrelated to the pull request: I find the damping of the oscillations very interesting. Do we know why the oscillations are damped with decreasing relaxation rate? Maybe we can talk separately.

Antoinehoff:
> I don't know the full background of this work but this piece looks very solid to me.
> 
> Something might be unrelated to the pull request: I find the damping of the oscillations very interesting. Do we know why the oscillations are damped with decreasing relaxation rate? Maybe we can talk separately.

Thanks for your feedback @dingyunl ! My belief is that the oscillations we observe are related to the Dirichlet BC that reflects wave unphysically (you can observe them almost undamped if you do not place buffer regions). In the limit of an infinite relaxation rate, the buffer regions effectively become Dirichlet BC and only reflect the wave. It is unclear to me what is the correct relaxation rate as I am not familiar with such buffer regions, we'll see!

The broader context is to reach the local GK regime and compare Gkeyll with local codes. This could be a very good physics-based test of the implementation and future development like the EM and FLR effects.

Antoinehoff:
I will open a new PR with a branch that is made from `main`, this branch here is too messy...

PR #1029, opened 2026-05-26
[PR] Implementation of a BGK static source

Antoinehoff:
This PR is related to DR #1022 

We implement here the static BGK source operator to relax the distribution function towards a user provided equilibrium at a user provided rate, i.e.
```math
S_{BGK} = -\nu_{eq} (x,y,z) \left[ f - f_{eq}(x,y,z,v_\parallel, \mu) \right]
```
where $\nu_eq$ and $f_{eq}$ are static relaxation frequency and equilibrium distribution function, respectively, provided by the user.

Below, we present some results obtained with the 2x2v Cyclone Base Case (CBC) regression test. These require also #1019 to be stable.

### Initial condition freeze 🥶 
We check first that it is possible to maintain the initial conditions. We set $\nu_eq = 10^6$ Hz and $f_{eq} = f_0$, $f_0$ being the initial condition for both electrons and ions.

_evolution of the profiles along x_

image

_evolution of the profiles along z_

image

We see that the system keeps the profiles constant thanks to the static BGK source, aside of a steady state potential and flow that develops in a time scale faster than the relaxation rate (<1$\mu$s). This is a consequences of a non canonical Maxwellian IC. At t=0, the potential and the flows are 0 and the density are equal everywhere. However, this system is not a steady state solution of the GK equation. Even with a static BGK source trying to maintain this state, the plasma finds an equilibrium with a non zero potential, flow and charge density and displays a geodesic acoustic mode (GAM).

We also see that we can control the damping of the oscillations by modifying the relaxation rate:
_left to right:_ $\nu_eq = 10^7, 5\times 10^6, 10^6$ Hz
Screenshot 2026-05-22 at 3 04 01 PM

### Buffer regions and Cyclone Base Case like simulation
We can now use the static BGK source to create buffer regions where the plasma should not vary. We use a setup inspired from V. Grandgirard et al. / Computer Physics Communications 207 (2016) 35–68.
The relaxation rate is now given a spatial dependency, following:
```math
\nu_{eq}= \nu_{eq,0}\, H(x)
```
where
```math
H(x) = 1 + \frac{1}{2} \left[ \tanh \left( \frac{x - x_{\max} + B_l L_x}{B_s L_x} \right) - \tanh \left( \frac{x - x_{\min} - B_l L_x}{B_s L_x} \right) \right]
```

Without the buffer spurious oscillations due to the unphysical Dirichlet BC applied to $\phi$ at the outer radial wall occur

image

With the buffer activated, we see that the fast scale oscillations are quickly damped, putting in evidence clean GAMs (with the correct frequency)

image

This allows us to run a Cyclone Base Case simulation
image

## Diagnostics
The static BGK source diagnostics consist in 
- The equilibrium function `[sim_prefix]-[ion/elc]_source_bgk_feq_0.gkyl`
- The relaxation rate `[sim_prefix]-[ion/elc]_source_bgk_rate_0.gkyl`
- The integrated primary moments of the total source term `[sim_prefix]-[ion/elc]_source_bgk_integrated_M0M1M2.gkyl`
Here is the evolution of these moments in a CBC simulation with buffer regions:

Screenshot 2026-05-23 at 11 42 12 AM

## Additional modifiations

This PR also refactor the `gk_species_source_bgk.c` in various ways:

- We now name the equilibrium function `feq` and not `fmax` as it can be used for any distribution function.
- We reduce the code duplication in initialization and release phase.
- We adapt the diagnostics to allow a diagnostic during the RHS computation for the static BGK source. This is required to get integrated moments of $\nu(f - f_{eq})$ without creating a new full phase space array.

## Limitation: lack of implicit relaxation and future reorg.
Because the static BGK relaxation is implemented via the source operator, it does not currently support implicit relaxation. One might want an extremely high relaxation rate to create a mask region in a volume of the domain. However, this is for now not doable since migrating the feature in the collision operator would prevent us to use any other physical collision model.


core/zero/gkyl_eqn_type.h
```diff
@@ -2,36 +2,36 @@
 
 // Identifiers for various equation systems
 enum gkyl_eqn_type {
-  GKYL_EQN_EULER,     // Euler equations
```
manauref, line 5: i think you should undo changes to this file. They are not related

manauref (changes requested):
@Antoinehoff i'm sorry to have to do this (because I suggested what you did), and we can discuss offline, but I think this should be done with the BGK collision operator instead. That is because:
1. The BGK source was meant to source things, given M0, M1, M2 rates. It's not really meant to do the job of relaxation, that's what BGK collisions are for.
2. You may not want a strong relaxation rate, but someone might. In that case implicit relaxation will be necessary, and you get that for free by doing this with BGK collisions.
3. The ability to specify f_eq in BGK collisions is something that already exists/has been used in Vlasov, so we'd be staying consistent with that.
We can wait for a second opinion from @JunoRavin if you want.

core/zero/gkyl_eqn_type.h
JunoRavin: We should not need to specify numbering in enums now that we're not having to convert between C and Lua in the Lua ffi interface. I don't think this is necessary (and it being 1-indexed is especially weird in my opinion and I don't think is correct downstream).

Antoinehoff: done

Antoinehoff:
> @Antoinehoff i'm sorry to have to do this (because I suggested what you did), and we can discuss offline, but I think this should be done with the BGK collision operator instead. That is because:
> 
> 1. The BGK source was meant to source things, given M0, M1, M2 rates. It's not really meant to do the job of relaxation, that's what BGK collisions are for.
> 2. You may not want a strong relaxation rate, but someone might. In that case implicit relaxation will be necessary, and you get that for free by doing this with BGK collisions.
> 3. The ability to specify f_eq in BGK collisions is something that already exists/has been used in Vlasov, so we'd be staying consistent with that.
>    We can wait for a second opinion from @JunoRavin if you want.

1. Collisions are supposed to relax to a Maxwelian distribution but the BGK source here is meant to relax to an arbitrary function so I find using "collision" for that misleading.
2. Ok I could see that we may want to have a fixed function region (as a mask) somewhere inside the domain for example, and this is not doable with BCs.
3. I wasn't aware of that. The thing is, can we declare two collision operators? Because one may want to add also a collision operator to such simulations and allowing to setup the buffer with `bgk_source` lets the collision operator free.

Otherwise we could maybe create a separate `.buffer` structure but it would be reducing the range of application.

manauref:
3. This is a showstopper to putting this in bGK collisions. So we'll leave it in BGK source for now. When we have time we'll move it over (I think there's a `detach` branch where I upgraded collisions to allow multiple collisions ops (to have an LBO-BGK hybrid for detachment)

Antoinehoff:
> 3. This is a showstopper to putting this in bGK collisions. So we'll leave it in BGK source for now. When we have time we'll move it over (I think there's a `detach` branch where I upgraded collisions to allow multiple collisions ops (to have an LBO-BGK hybrid for detachment)

Ok I will write a limitation section to try to rehearse what was discussed.

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1252,7 +1259,8 @@ struct gk_neut_species {
 struct gk_field {
   struct gkyl_gyrokinetic_field info; // Data for field.
 
-  enum gkyl_gkfield_id gkfield_id;
+  enum gkyl_gkfield_id gkfield_id; // Field solver ID.
+  enum gkyl_gyrokinetic_bc_type bc_par_phi; // Parallel BC to enforce on phi.
```
manauref, line 1299: shouldn't be in this PR

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1388,6 +1396,8 @@ struct gkyl_gyrokinetic_app {
   struct gkyl_range global_upper_skin[GKYL_MAX_DIM];
   struct gkyl_range global_upper_ghost[GKYL_MAX_DIM];
 
+  // Range extended in parallel direction (for TS BCs).
+  struct gkyl_range global_par_ext, local_par_ext;
```
manauref, line 1436: i don't think this should be in this PR

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
```diff
@@ -1033,6 +1039,7 @@ struct gk_species {
   struct gkyl_range global_lower_ghost[GKYL_MAX_DIM];
   struct gkyl_range global_upper_skin[GKYL_MAX_DIM];
   struct gkyl_range global_upper_ghost[GKYL_MAX_DIM];
+  struct gkyl_range local_par_ext; // Range extended in parallel direction for TS BC.
```
manauref, line 1076: don't think this should be in this PR

manauref (changes requested):
I think eventually this operation should go into the BGK collision operator, but cause putting it there now would prevent us from adding collisions, we'll put it in the BGK source.

Since this is in a temporary placeholder (module), I think you should insert it without changing the rest of the BGK source modulte. I see you changed a bunch of things, like names of variables, that are used for other BGK source types, and that shouldn't happen. One reason I'm particularly protective is that the Gkeyll-EIRENE coupling is fragile, and I don't want to jeopardize it.

Do you think you can insert this new option without changing anything the existing options use?

gyrokinetic/apps/gkyl_gyrokinetic_priv.h
Antoinehoff: this is because I anticipated the merge of the core only branch which should occur very soon I hope...

Antoinehoff:
This is the comparison of ion integrated moments of the sheath heating regression test (`rt_gk_sheath_heat_source_2x2v_p1.c` between `main` and this PR:

#### `main`
image

#### PR
image

A very small discrepancy in c1 which is the parallel flow which should be zero in both case so I think the only difference is due to machine precision.

Antoinehoff:
So we observed with @manauref that the `rt_gk_multib_step_eirene_2x2v_p1.c` regression test is non deterministic in `main` and the initial `dt` is not always the same and so the `runregression run check` tool is sometimes failing and sometimes not within `main` itself. It is then hard to check if this branch does not break the eirene interface, though we see that this PR does not make this issue worse.

manauref (approved):

JunoRavin (approved):

PR #1030, opened 2026-05-26
[PR] Fix nodal range initialization to use global dimensions in geometry GK write geom

Antoinehoff:
This bug introduced in PR #992 was silent in 2x2v but makes 3x2v multi GPU run fails because of a `if (rank ==0)` statement followed by the usage of `local` ranges to write a global array.

This was introduced by a new code snippet to write nodal coordinates. I did not check if the fix is modifying the output.

### DIFF
--- a/gyrokinetic/apps/gyrokinetic.c
+++ b/gyrokinetic/apps/gyrokinetic.c
@@ -1181,19 +1181,26 @@ gkyl_gyrokinetic_app_write_geometry(gkyl_gyrokinetic_app* app, struct gkyl_gk_ge
 
   if (rank == 0) {
     // Create Interior Nodal Range and Grid and Write Nodal Coordinates
-    struct gkyl_array* mc2pint_nodal = mkarr(false, 3, app->gk_geom->nrange_int.volume);
+    // nrange_int must be sized from app->global (not the per-rank local nrange_int)
+    int num_quad_points_int = app->poly_order + 1;
+    int num_nodes_int[GKYL_MAX_CDIM];
+    for (int d=0; d<app->grid.ndim; ++d)
+      num_nodes_int[d] = gkyl_range_shape(&app->global, d) * num_quad_points_int;
+    struct gkyl_range nrange_int_global;
+    gkyl_range_init_from_shape(&nrange_int_global, app->grid.ndim, num_nodes_int);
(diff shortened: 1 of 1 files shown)


gyrokinetic/apps/gyrokinetic.c
```diff
@@ -1181,19 +1181,26 @@ gkyl_gyrokinetic_app_write_geometry(gkyl_gyrokinetic_app* app, struct gkyl_gk_ge
 
   if (rank == 0) {
     // Create Interior Nodal Range and Grid and Write Nodal Coordinates
-    struct gkyl_array* mc2pint_nodal = mkarr(false, 3, app->gk_geom->nrange_int.volume);
+    // nrange_int must be sized from app->global (not the per-rank local nrange_int)
+    int num_quad_points_int = app->poly_order + 1;
```
Maxwell-Rosen, line 1185: This can be found in the basis rather than hard coding it

Antoinehoff: what is hard coded?

Antoinehoff: next time @Maxwell-Rosen I really want you to propose a solution instead of just raising a problem

JunoRavin: @Maxwell-Rosen your addition to the basis object is not appropriate here because you added num_quad for the basis (which could be 3x), but this number of nodes has to be constructed dimension by dimension so that we can utilize gkyl_range_init_from_shape (and thus you have the total number of nodes in x in the [0] entry of num_nodes_int, the total number of nodes in y in the [1] entry of num_nodes_int, and so on).

JunoRavin (approved):
The structure of this range initialization follows https://github.com/ammarhakim/gkeyll/blob/main/gyrokinetic/zero/gkyl_gk_geometry_priv.h#L47 but for the global range so that the IO works. The fact that this is needed suggests we have some significant cleanup to undertake on the geometry organization (especially with respect to writing out geometry information), but considering simulations cannot be run at all until this change is merged, it should be merged with the team making a note for a larger discussion on geometry.

akashukla (approved):
Good catch, this was a mistake.

PR #1032, opened 2026-05-28
MB Diffusion Fix

akashukla:
For anomalous diffusion, the diffusion coefficient in computational coordinates is nu = g^{xx} J D. This differs by a factor of dz at radial MB boundaries (The same factor which appears in the surface elements at the boundary). The diffusion coefficient in the ghost cell needs to be rescaled by this ratio to ensure particle conservation. With this fix, my production simulations finally settle to the right particle and energy output in steady state.


manauref (approved):

Antoinehoff (approved):
good catch!

PR #1034, opened 2026-05-28
Gk ai sheath surf bimpactangle

Antoinehoff:
Change the representation of impact angle in the surrogate call (see PR #1033 )


Antoinehoff (commented):
I compared both representations using TCV 2x2v regression test:
`volume representation`
```
alpha = 4.28602
alpha = 4.0931
```
`surface representation`
```
alpha = 4.38445
alpha = 4.178780
```

PR #1040, opened 2026-06-02
Intermediate 3x GK MB infra

manauref:
This provides partial work to do 3x multiblock sims with limiters or divertors, and some work for single-block sims with numerical equilibrium for limited shots.

IMPORTANT: this PR deprecates the use of IWL BC options in the species BC tables (continuing the work in Issue #1003 ). Now for limited clopen simulations:
- Use the species BC table to specify the BC at the limiter plate.
- Give the geometry the LCFS.
If the LCFS is present, the app will automatically do what it needs to do (e.g. apply twist-shift in the core) under the hood.

Regarding 3x MB diverted shots:
1. An initial TCV-X21 input file is added. We also add a 3x SOL AUG input file meant to reproduce a single block sim. 
2. There are still some major improvements needed for 3x diverted shots to work and become routine. Some of these include:
- Need to deal with production of high-k modes at twist-shift boundaries.
- Need to improve the perpendicular poisson solver so that we don't need radially-connected blocks to have the same x resolution.
- Need to improve parallelism to give us more flexibility on how we map MPI processes to cuts/blocks.
- Need to test (resolution) convergence.

Finally:
There are no results from this PR. I just want to merge infra before:
- more work goes into 3x MB and the branch becomes too huge
- we put this aside for a little while and forget exactly what happened here
- other work on TS BCs (do deal with high-k) happens in other branches
So the emphasis of the PR review should be to ensure current capabilities don't break.


manauref:
I ran a production TCV IWL 3x2v case for several hundred microseconds and the result look essentially the same as in main, e.g. here's the ion density at the outboard midplane (y=0) over time:

image

and the corresponding regression test gave exactly zero difference with main (the production case doesn't give 0 difference because it was run on GPUs and those introduce tiny nondeterminism).

Antoinehoff (approved):

gyrokinetic/apps/gk_field_2x3x.c
```diff
@@ -506,34 +441,36 @@ gk_field_fem_release_2x3x(const gkyl_gyrokinetic_app *app, struct gk_field *f)
 
   gkyl_array_integrate_release(f->calc_em_energy);
 
-  // Release TS updaters.
-  if (f->bc_par_phi == GKYL_BC_GK_FIELD_TWISTSHIFT) {
+  if (app->gk_geom->has_LCFS) {
+    // Release IWL updaters.
```
tnbernard: Minor comment: "IWL" is not very general. What if the geometry is diverted? Should this be "TS" instead?

gyrokinetic/apps/gk_neut_species_fluid.c
```diff
@@ -228,7 +228,7 @@ gk_neut_species_fluid_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app,
   ns->local = app->local;
 
   // Keep a copy of num_periodic_dir and periodic_dirs in species so we can
-  // modify it in GK_IWL BCs without modifying the app's.
+  // modify it in IWL BCs without modifying the app's.
```
tnbernard: Again I thought we are avoiding "IWL"

gyrokinetic/apps/gk_neut_species_kinetic.c
```diff
@@ -675,7 +675,7 @@ gk_neut_species_kinetic_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *ap
     s->local, s->local_ext, s->local_vel, s->local_ext_vel, app->use_gpu);
 
   // Keep a copy of num_periodic_dir and periodic_dirs in species so we can
-  // modify it in GK_IWL BCs without modifying the app's.
+  // modify it in IWL BCs without modifying the app's.
```
tnbernard: Same comment about "IWL"

gyrokinetic/apps/gk_species.c
```diff
@@ -1448,7 +1434,7 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
   gkyl_velocity_map_write(gks->vel_map, gks->comm, app->name, gks->info.name);
   
   // Keep a copy of num_periodic_dir and periodic_dirs in species so we can
-  // modify it in GK_IWL BCs without modifying the app's.
+  // modify it in IWL without modifying the app's.
```
tnbernard: Same "IWL" comment

tnbernard (approved):
Have some minor cosmetic comments but otherwise looks ok.

gyrokinetic/apps/gk_field_2x3x.c
manauref: true, i'll change

gyrokinetic/apps/gk_neut_species_fluid.c
manauref: ja this isn't accurate. Will change.

gyrokinetic/apps/gk_species.c
manauref: also not accurate. will change

PR #1041, opened 2026-06-04
[PR] extension of the dg bin op unit test for subspace accumulation

Antoinehoff:
This PR extends the unit test `ctest_dg_bin_ops.c` to verify that we can perform an operation of the type

```math
h(x,y,z) = g(x,y,z) + \alpha f(x),
```

using `gkyl_dg_mul_conf_phase_op_accumulate_range`, which was previously untested. We do that by initalizing an array of ones in the full space $a(x,y,z)=1$ and use it as a multiplication factor in `gkyl_dg_mul_conf_phase_op_accumulate_range`. It is worth noting that the latter routine was not tested in any unit test previously.

### Context

This operation is useful to perform substraction of the flux surface average i.e.
```math
\phi \leftarrow \phi - \langle \phi \rangle_{yz}.
```

### Limitations

The sub-dimensional array **must share the first n dimensions** with the full dimensional one. In other word, one cannot do something like,
```math
h(x,y,z) = g(x,y,z) + f(y),
```
but this is outside of the current scope of interest and would require reworking the `gkyl_dg_mul_conf_phase_op_accumulate_range` routine.

As a last comment, one could consider renaming `gkyl_dg_mul_conf_phase_op_accumulate_range` to something like `gkyl_dg_mul_subspace_op_accumulate_range` to put in evidence it's more general capabilities.

### DIFF
--- a/core/unit/ctest_dg_bin_ops.c
+++ b/core/unit/ctest_dg_bin_ops.c
@@ -88,13 +88,13 @@ test_1d(int poly_order, bool use_gpu)
   gkyl_cart_modal_serendip(&basis, ndim, poly_order);
 
   // projection updater for dist-function
-  gkyl_proj_on_basis *projDistf = gkyl_proj_on_basis_new(&grid, &basis, poly_order+1, 1, f_1d, NULL);
-  gkyl_proj_on_basis *projDistg = gkyl_proj_on_basis_new(&grid, &basis, poly_order+1, 1, g_1d, NULL);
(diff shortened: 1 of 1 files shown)


Antoinehoff:
@JunoRavin, @manauref, the new test fails for $p=2$, any idea why?

Antoinehoff:
> @JunoRavin, @manauref, the new test fails for p = 2 , any idea why?

Actually it passes if I set the acceptable error level to be 1e-11 instead of 1e-12

JunoRavin (approved):

ammarhakim (approved):
Ready to merge

PR #1051, opened 2026-06-10
The runregression system was silently failing to do anything if a user requested to run a particular test (or tests) and those names did not match anything in our test suite.

JunoRavin:
Fix by adding new tables for what entires *can* be run and what entries were found. Also add a check to partial string parsing in case a user provides the name of a test which is close, such as

ERROR: no regression test named 'rt_gk_sheath_3x2v' found in layer(s): moments, vlasov, gyrokinetic, pkpm. Did you mean one of these?
   gyrokinetic/creg/rt_gk_sheath_3x2v_p1
   gyrokinetic/creg/rt_gk_sheath_3x2v_p1_cons
   gyrokinetic/luareg/rt_gk_sheath_3x2v_p1
Use 'gkeyll runregression list' to see all available tests.

Also fix an error where if someone provides the name of a test they want to run, but that test is currently in one of the ignore lists (because we're not currently accepting the results for whatever reason), then instead of just again, silently doing nothing, the system returns that the user has requested to run a test which is being ignored.

### DIFF
--- a/gkeyll/lua/Tool/runregression.lua
+++ b/gkeyll/lua/Tool/runregression.lua
@@ -647,6 +647,14 @@ end
 local function list_tests(activeLayers, args)
    local luaTests, cTests = {}, {}
 
+   -- Track which --run-only entries matched at least one existing test file
+   -- (in any scanned layer, Lua or C) so we can report unknown test names
+   -- instead of silently running nothing.
+   local runOnlyEntries, runOnlyFound = {}, {}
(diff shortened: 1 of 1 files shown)


Antoinehoff (approved):

Antoinehoff:
Thank you for making this idiot proof, I need it :)

jjz73 (approved):

manauref:
works as advertised!

PR #1054, opened 2026-06-12
Fix symlink creation in runregression

JunoRavin:
If you look in individual runs for your runregression system, you will see a directory for that app layer (so all gyrokinetic/creg-runs have a gyrokinetic/ folder in them). This is a symlink to your source app-layer directory so that any C regression test can resolve relative paths to what we keep in that layers data/ directory (such as eqdsk files).

This symlink was only checking if it existed and then running, which could produce a silent failure if you switched source repositories and changed something about the underlying data (if some eqdsk file gets updated). So always create the symlink safely by rm -f and ln -s. Note the lack of -rf because we don't want to accidentally delete a directory and delete data. We let the rm command fail in that case.

### DIFF
--- a/gkeyll/lua/Tool/runregression.lua
+++ b/gkeyll/lua/Tool/runregression.lua
@@ -1001,13 +1001,19 @@ local function prepareCRun(test, timeoutSecs, mode, skipCompile)
    end
 
    -- Symlink the layer source dir so tests can find data files by relative path.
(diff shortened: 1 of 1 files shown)


manauref (approved):
good to fix silent errors. Thank you!

ammarhakim (approved):
Ready to merge.

PR #1055, opened 2026-06-13
Standardize the Vlasov velocity map implementation, always creating the mapping and when using Serendipity basis, falling back to C^0 linear map

JunoRavin:
The work in kitp-hamil-vlasov implemented the higher order mapping function described in DR: https://github.com/ammarhakim/gkeyll/issues/444 and the following paper: https://arxiv.org/abs/2602.17487

However, that higher order mapping, in which the mapping is represented with a C^1 cubic and thus the velocity space Jacobian is a continuous quadratic function, is only valid for tensor bases with p>=2 in velocity space (due the form of the velocity space Jacobian division operation at Gauss-Legendre quadrature points). 

To enable users to still utilize nonuniform mappings even when running with the Serendipity basis, we now run in two modes: if running with a tensor basis (since there is currently no tensor p=1 hybrid implementation), everything proceeds as before with the creation of the C^1 cubic mapping and the continuous quadratic velocity space Jacobian, but if running with the Serendipity basis, project the mapping onto a C^0 linear mapping with piecewise constant (and thus discontinuous) velocity space Jacobian. 

Thus the velocity-space Jacobian can be created by default in all cases (with the mapping falling back to the identity map when no mapping is detected and nothing changing about the underlying operations). 

Few more quality of life improvements:
1. Consolidate all nonuniform velocity space arrays and operations into the gkyl_vlasov_velocity_map struct (with corresponding functions for writing, dividing out the Jacobian, and rescaling by the Jacobian). 
2. New unit tests for both moment computation and special relativistic auxiliary variable (things like the sr_pressure) computation. These are designed to compare against analytic and numerically expected results for a multivariate Gaussian with full M2ij in the moment case, and a drifting relativistic Maxwellian in the sr_vars case. 

Unit tests are passing but I haven't yet done a sweep of the kernels to double-check the new Maxima isn't doing anything non-standard on the simplification side


ammarhakim (approved):
This looks ready to merge. We should aim to merge the KITP branch into main as soon as possible.

JunoRavin:
I also added new collisions tests for checking kitp-hamil-vlasov's collisions implementation (https://github.com/ammarhakim/gkeyll/commit/5ccede2998e81d23560f73ba93d3899f4a01d9bc). I will keep working through this after this is merged, but we should be converging towards merging of kitp-hamil-vlasov.

johnson452 (approved):
I've tested that the dg_gr_maxwell and gr_traids fold in correctly and the physics results remain unchanged. Further, I've verified the CPU and GPU (Perlmutter) builds, ran select regression tests and sanitizer on these tests. This looks ready to merge.

PR #1059, opened 2026-06-15
Reorganize field struct in Vlasov-Maxwell app for modularity between Vlasov-Maxwell and Vlasov-Poisson

JunoRavin:
This PR goes with this DR: https://github.com/ammarhakim/gkeyll/issues/1058

Previously, the switching between different field types such as Vlasov-Poisson and Vlasov-Maxwell was handled at the app level, with function pointers defined in `gkyl_vlasov_app` like so:

```
// Vlasov object: used as opaque pointer in user code
struct gkyl_vlasov_app {
...
  bool has_field; // has field
  struct vm_field *field; // pointer to field object
  // Pointer to function that calculates external electric and magnetic fields.
  void (*field_ext_em_calc)(gkyl_vlasov_app *app, struct vm_field *field, double tm);
  // Pointer to function that calculates external applied currents.
  void (*field_app_current_calc)(gkyl_vlasov_app *app, struct vm_field *field, double tm);
  // Pointer to function that calculates external potentials (phi, A).
  void (*field_ext_pot_calc)(gkyl_vlasov_app *app, struct vm_field *field, double tm);
  // Function used to compute the integrated field energy (e.g., E^2 + B^2 or |grad(phi)|^2). 
  void (*field_energy_calc)(gkyl_vlasov_app *app, double tm, const struct vm_field *field);
  // Function which writes out electromagnetic/electrostatic fields. 
  void (*field_write)(gkyl_vlasov_app *app, double tm, int frame);
  // Function which writes out integrated field energy. 
  void (*field_energy_write)(gkyl_vlasov_app *app);
...
}
```

This organization has a number of disadvantages: firstly it mixes layers. It forces vlasov.c to both wrap the lower level private API calls to different Vlasov field functionality and the public API calls for user-defined components such as calculating the field energy. Further, it makes the deployment of Vlasov-Poisson vs. Vlasov-Maxwell nonuniform (some places you have app->func, some places you do an if statement because that function is not defined for that field type). Finally, and perhaps crucially, it's intrinsically different from the way the species wrapping is handled at this point for no clear reason, introducing potential confusion on how to extend the field objects, where to put certain methods, and which calls go where for different field types. 

I have refactored the field creation and use throughout the app program flow with a new API defined in vlasov_field.c 

vlasov_field.c and its associated private header gkyl_vlasov_field_priv.h are **not** new objects. They only wrap the existing vm_field.c and vp_field.c infrastructure in a way that there is a single set of function pointers each field object must define so that all of the vlasov.c (and associated vlasov_forward_euler.c and vlasov_update_ssp_rk3.c) functions are defined appropriately and there is no mixing of a public and private API. This refactor means that now, those function pointers are part of the `vm_field` object:

```
struct vm_field {
  struct gkyl_vlasov_field info; // data for field
  enum gkyl_field_id field_id; // Type of field.

  // Type-specific methods set by vm_field_new()/vp_field_new(). Dispatched
  // through the vlasov_field_* wrappers so callers stay field-type-agnostic.
  double (*update_func)(gkyl_vlasov_app *app, double tcurr,
    const struct gkyl_array *fin[], const struct gkyl_array *emin, struct gkyl_array *emout);
  void (*combine_func)(gkyl_vlasov_app *app, struct vm_field *field, struct gkyl_array *out,
    double c1, const struct gkyl_array *arr1, double c2, const struct gkyl_array *arr2);
  void (*copy_range_func)(gkyl_vlasov_app *app, struct vm_field *field,
    struct gkyl_array *out, const struct gkyl_array *inp);
  void (*apply_ic_func)(gkyl_vlasov_app *app, struct vm_field *field,
    const struct gkyl_array *fin[], double t0);
  void (*apply_bc_func)(gkyl_vlasov_app *app, const struct vm_field *field, struct gkyl_array *em);
  void (*limiter_func)(gkyl_vlasov_app *app, struct vm_field *field, struct gkyl_array *em);
  void (*complete_update_func)(gkyl_vlasov_app *app, double dt, const struct gkyl_array *fin[],
    const struct gkyl_array *fluidin[], const struct gkyl_array *emin, struct gkyl_array *emout);
  void (*calc_ext_em_func)(gkyl_vlasov_app *app, struct vm_field *field, double tm);
  void (*calc_app_current_func)(gkyl_vlasov_app *app, struct vm_field *field, double tm);
  void (*calc_ext_pot_func)(gkyl_vlasov_app *app, struct vm_field *field, double tm);
  void (*calc_energy_func)(gkyl_vlasov_app *app, double tm, const struct vm_field *field);
  void (*write_func)(gkyl_vlasov_app *app, double tm, int frame);
  void (*write_energy_func)(gkyl_vlasov_app *app);
  // Read the field's restart data for the given frame. Vlasov-Maxwell reads the
  // EM field from its restart file; Vlasov-Poisson (re-solved from the restarted
  // distribution elsewhere) and the null field are no-ops.
  struct gkyl_app_restart_status (*read_func)(gkyl_vlasov_app *app, struct vm_field *field, int frame);
  void (*release_func)(const gkyl_vlasov_app *app, struct vm_field *field);
...
}
```

and vlasov_field.c's API calls wraps these functions so that the calls in the top-level app and time stepper dispatch to the correct field calls.

Along the way additional quality-of-life improvements emerged from this refactor:
1. All the vm_field and vp_field private API now naturally sits in gkyl_vlasov_field_priv.h, making the overall gkyl_vlasov_priv.h more modular and giving developers a slightly easier time for where they can go to add new field capabilities, with a noted demarcation between the API that vlasov_field provides that wraps the private functions to vm_field and vp_field
2. NULL_FIELD is now a specific field object that is created, eliminating numerous app->has_field calls throughout the code and thus helping to prevent bugs/crashes from cropping up where someone adds something and forgets to guard it with app->has_field. Now so long as a new developer sees the list of functions expected to be defined for field, and correctly makes the function a no-op, there will be no error propagation into the main function loop. 
3. Further modularization revealed itself in the form of e.g., having a read_func for each field type, thus clearly separating field types such as Maxwell's that require us to read a restart frame and field types such as Poisson that obtain the potential solely from the distribution function that is read-in
4. **Important program flow piece: this refactor also revealed a straightforward way for the update method to fold into the vlasov_forward_euler call regardless of field type. We compute the "field update" before the species update. If we are Maxwell's nothing is changed and everything is still bit-identical. If we are Poisson, we have then obtained the potential at the correct time for the time stepping of the distribution function.**


johnson452 (approved):
Looks good, and passes make Vlasov-check on GPU. I ran several regression tests for GR / flat space Maxwell on GPU. Further ran sanitizer which was clean. Ready to merge pending a small commit by Jimmy on restarts.

PR #1060, opened 2026-06-15
Kinetic radiation fit parameters Python sub-repository

jaimix4:
## Purpose / Summary
This PR integrates a automated radiation fitting sub-repository (`kin-rad-py`) for calculating plasma radiation parameters in the fashion of a kinetic operator according to the methodology developed in [J. Roeltgen et al. NF 65 2025](https://iopscience.iop.org/article/10.1088/1741-4326/adff28). This implementation was based in the [original repository](https://github.com/jRoeltgen/radiation_operator) of this work. 

Key changes include:
* **Pipeline Integration:** Added the `kin-rad-py/` subdirectory in Gkeyll containing the standalone Python data fetching, batch fitting, and instructions for future development.
* **Updated Database:** Rebuilt `radiation_fit_parameters.txt` using non-linear optimization algorithms (SLSQP), improving curve accuracy across a wider range of density intervals. See the following example:

H_0_13 0_20260526_125622

The current database with which the new `radiation_fit_parameters.txt` is `prod_rs_server_1` inside `kin-rad-py/fits_data/`. To check all the comparison plots for the different elements and density [see plot files in the repository where this work was prototyped](https://github.com/jaimix4/kin-rad-gkeyll/tree/main/fits_data/fit-db_prod_rs_server_1/plots).

## Testing & Verification
* Verified that the updated database formats are compatible with the core C-reader parsing modules in `gyrokinetic/zero/radiation_read.c`. For further information in implementation check the README.md in `kin-rad-py`.
* Compared results of legacy database and the new produced test in regression test: `rt_gk_rad_1x2v_p1`, `rt_gk_rad_nonuniformv_1x2v_p1` and `rt_gk_rad_low_Te_1x2v_p1` for H plasma. The results for integrated moments are presented here:
image
image
image

## Future work
* In case a new database wants to be created, or add new elements (e.g. W) to repository already allows this functionality, please check README.md. 
* If fits needs to be fine-tune, the architecture allows for the implementation of the functionality of fine-tuning a fit and then replace it in a existing database. Several parameters can be fine tune in fit, as for example the minimum $T_e$ taken into account during the fit. Please check `fit_batch.py` as most scripts have useful documentation. 

## Core Files Added/Modified
* `gyrokinetic/data/adas/radiation_fit_parameters.txt` (Updated text database)
* `gyrokinetic/data/adas/kin-rad-py/` (New tracked python modules and sub-.gitignore)

Please any other test needed, let me know.


jRoeltgen:
The energy loss in the first 2 regression tests changed in a systematic way. Is the amount it changed the amount that is expected based on the different fits? It looks like it changes in the correct direction, but I don't remember the temperature in these tests.

jaimix4:
> The energy loss in the first 2 regression tests changed in a systematic way. Is the amount it changed the amount that is expected based on the different fits? It looks like it changes in the correct direction, but I don't remember the temperature in these tests.

For rt_gk_rad_1x2v_p1 and rt_gk_rad_nonuniformv_1x2v_p1, the normalized temperature is log10(30eV) ~ 1.5 and the radiation cutoff temperature is log10(12eV) ~ 1.07. From the $L_z$ vs $T_e$ plots for H, at log10($T_e$) = 1.5 my fit has higher emissivity, hence why systematically for these two test the 3rd moment shows higher loss of energy for the tests with my fit.

For  rt_gk_rad_low_Te_1x2v_p1, the normalized temperature is 0.72 eV, and it still has the 12eV cutoff, hence why the show the same energy loss since the emissivity is dictated by the established amount below the cutoff for both.

jRoeltgen:
> > The energy loss in the first 2 regression tests changed in a systematic way. Is the amount it changed the amount that is expected based on the different fits? It looks like it changes in the correct direction, but I don't remember the temperature in these tests.
> 
> For rt_gk_rad_1x2v_p1 and rt_gk_rad_nonuniformv_1x2v_p1, the normalized temperature is log10(30eV) ~ 1.5 and the radiation cutoff temperature is log10(12eV) ~ 1.07. From the L z vs T e plots for H, at log10( T e ) = 1.5 my fit has higher emissivity, hence why systematically for these two test the 3rd moment shows higher loss of energy for the tests with my fit.
> 
> For rt_gk_rad_low_Te_1x2v_p1, the normalized temperature is 0.72 eV, and it still has the 12eV cutoff, hence why the show the same energy loss since the emissivity is dictated by the established amount below the cutoff for both.

If you haven't already, it might be good to quantitatively check that difference. But it looks good.

jRoeltgen (approved):

PR #1061, opened 2026-06-16
There is a pathology in how we handle restarts that can cause the final frame of a simulation to not be written out because of how we anchor I/O vs. restarts

JunoRavin:
Our trigger logic is that the trigger fires at the first step that overshoots the time at which the trigger occurs. When our simulations start at t=0, since all frames are integer divisors of t_end, this logic has no issue. If your time step is 0.003, you are writing data every 1.0 time units, the I/O will occur at time t=1.002, 2.001, ... and so on

But consider what happens at restart: now the t_curr at restart is whatever the *actual* time was when you wrote our data, which could be 1.002 or 2.001 (not 1.0 or 2.0).

There is now a systematic drift in the simulation time vs. the trigger time. Now the simulation thinks the final I/O trigger time is t_end + epsilon, where epsilon is whatever offset you started at. This offset can cause the simulation to end before the final output time is written.

Instead, across all four apps, I propose to re-anchor the trigger *as if* the time we wrote out was the integer time, tcurr = frame_curr*(t_end/num_frames) instead of whatever the time is in the metadata. All this is doing is re-seeding the trigger to the nominal temporal grid point the data was written at, thus allowing us to avoid issues where the final output frame is not written because we undershoot the time we need to get to, to write out the final frame of data.

### DIFF
--- a/gyrokinetic/apps/gyrokinetic_lw.c
+++ b/gyrokinetic/apps/gyrokinetic_lw.c
@@ -2423,7 +2423,7 @@ gk_app_run(lua_State *L)
   int field_energy_calcs = app_lw->field_energy_calcs;
   int integrated_mom_calcs = app_lw->integrated_mom_calcs;
   // Triggers for IO and logging.
-  struct gkyl_tm_trigger io_trig = { .dt = t_end / num_frames, .tcurr = t_curr, .curr = frame_curr };
+  struct gkyl_tm_trigger io_trig = { .dt = t_end / num_frames, .tcurr = frame_curr * (t_end / num_frames), .curr = frame_curr };
   struct gkyl_tm_trigger fe_trig = { .dt = t_end / field_energy_calcs, .tcurr = t_curr, .curr = frame_curr };
   struct gkyl_tm_trigger im_trig = { .dt = t_end / integrated_mom_calcs, .tcurr = t_curr, .curr = frame_curr };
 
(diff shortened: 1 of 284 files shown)


manauref:
I haven't seen this happen in GK simulations, though now that I see your code change I remember I have some older input files where I did this, likely for a reason like yours. I have two questions:
1. Is there a reg test that reproduces this behavior?
2. Aren't you causing the restart to have a fake time? Like in your example, the restart is taking place at the physical time of 1.002, but you are artificially lowering the time to 1.0. That seems wrong to me, to alter the simulation time for the purposes of making I/O and postprocessing easier.

JunoRavin:
1. The Vlasov-Poisson regression tests broadly have this issue (like vp_buneman). 
2. You're only altering the *trigger*. You are not changing tcurr. You are telling the trigger at what integer multiples of t_end you're writing out data to avoid the trigger thinking the final trigger is at t_end + epsilon instead of t_end. What will happen after you restart is that you will rewrite that frame at the current time (because it's past the trigger) but that's okay. It's perhaps a testament to the confusion of the names that we call that input to the linear trigger t_curr when it need not be the actual current time.

manauref:
Interestingly, restarts for some GK reg test were indeed already behaving strangely in main already. I hadn't realized this. For example
`rt_gk_sheath_2x2v_p1`
does generate the _1 final frame. But if I run it, increase t_end and num_frames by 2x, and run it again with `-r 1` it says
```
Restarting from frame 0 at time = 0
```
What? I don't understand how/why this is possible.

But such behavior does not happen in this new branch.

manauref (approved):

JunoRavin:
With @manauref's approval, I wanted to tag in @Antoinehoff and @tnbernard as two folks I know most often dealing with our restart functionality for review and testing if either of you have the bandwidth to look at this.

Antoinehoff:
I don't think I've ever paid attention to that. Also, because I generally don't output phase space frame with a frequency of 1 (way too expensive for storage), it's uncommon that the last trigger is also a phase space frame. So I don't expect to restart precisely anyway.

Antoinehoff (approved):

PR #1064, opened 2026-06-19
Change dg_bin_op signature, fix/improve meta & rename eirene reg test files

manauref:
We make 4 changes:
1. Change the signature of the dg_bin_op functions so that the basis is passed by address instead of by value. This makes it easier to call dg_bin_op functions from postgkyl.
2. Change the way GK apps use metadata so that the species has a _basic, a _conf and a _phase metadata. These are for integrated diagnostics, conf-space diagnostics and phase-space diagnostics respectively, and each species submodule should be derived from these. This enables storing species-specific metadata that we can use in python for more complex post-processing (e.g. storing species mass in the `_source_M2_#.gkyl` file so that we can compute the integrated power in Watts).
3. Rename the gyrokinetic/data/eirene/ files so that the geo files have the new convention, this makes the rt_gk_multib_step_asdex_2x2v_p1 test work again.
4. Add a local (to a cell) differentiation updater, see PR #1068 

Regarding 1: I tried unit regression tests and they still work.

Regarding 2: here's what pgkyl info on a file looks like now, see the extra meta near the bottom

```
pgkyl rt_gk_tcv_iwl_adapt_source_2x2v_p1-elc_HamiltonianMoments_1.gkyl info
(default#0)
├─ Time: 1.001240e-07
├─ Frame: 1
├─ Number of components: 12
├─ Number of dimensions: 2
├─ Grid: (uniform)
│  ├─ Dim 0: Num. cells: 9; Lower: 0.000000e+00; Upper: 1.200000e-01
│  └─ Dim 1: Num. cells: 8; Lower: -3.141593e+00; Upper: 3.141593e+00
├─ Maximum: 9.829110e+19 at (0, 0) component 0
├─ Minimum: -1.410603e+19 at (2, 0) component 1
├─ DG info:
│  ├─ Polynomial Order: 1
├─ Created with Gkeyll:
│  ├─ Changeset: 761f7bdea58f+
│  └─ Build Date: 2026-06-18 19:43:09 MST
├─ Geometry info:
├─ Description: Number density, momentum density, and Hamiltonian energy density.
├─ Species properties:
│  ├─ Mass: 9.109382e-31
│  ├─ Charge: -1.602176e-19
│  ├─ Velocity dimensions: 2
```


Antoinehoff (approved):

Maxwell-Rosen:
Surely not all the regression tests are passing. There are several that are failing and are not addressed by this (particularly in moments). Did you check the eirine regression test though? That one is segfaulting on main because the data isn't labeled correctly.

manauref:
> Surely not all the regression tests are passing. There are several that are failing and are not addressed by this (particularly in moments). Did you check the eirine regression test though? That one is segfaulting on main because the data isn't labeled correctly.

eirene test runs. What tests do you think are failing?

Maxwell-Rosen:
Just the Eirene one was in the forefront of my mind. I found it unusual that so many Eirene files were changed, but the text of the PR did not mention that test. Thank you for fixing it.

JunoRavin (approved):

Issue #1065, opened 2026-06-19
[DR] dg_differentiate updater


manauref:
Summary
----------

I propose to build a simple (at first) updater in core to do local differentiation of a DG field, simply what Maxima would do with
```
calcInnerProdList( vars,1,basis,diff(f_e,diff_var,diff_order) );
```
and put it in core. The purpose is to call it from postgkyl like we can now call `dg_bin_mul_op` and `dg_bin_inv_op`, but to instead do a simple DG differentiation (without interpolation), e.g. to compute the gradient of phi.

Interface
---------

The core/dg_differentiate.c file will have a function similar to `dg_bin_mul_op` called `dg_differentiate_local` like
```
void gkyl_dg_differentiate_local(const struct gkyl_basis *basis,
  int dir, int order,
  int c_oop, struct gkyl_array *out,
  int c_iop, const struct gkyl_array *inp);
```
and this will load and use the corresponding Maxima-generated kernel.

In the future we'll extend this module, by adding other functions, that call hyper_dg and allow multi-cell (e.g. recovery) differentiation.

we'll prototype this in the `dg_binop_sig_change-differentiate` branch.

ammarhakim:
Don't we already have this? I see in the basis function struct a function pointer:

```c
  double (*eval_grad_expand)(int dir, const double *z, const double *f);
```

manauref:
> Don't we already have this? I see in the basis function struct a function pointer:
> 
>   double (*eval_grad_expand)(int dir, const double *z, const double *f);

But this evaluates it at a point. What if I don't want to evaluate it? I want to return DG coefficients.

manauref:
Addressed in PR #1068

PR #1068, opened 2026-06-23
DG differentiate updater

manauref:
This adds a simple dg_differentiate_op_local operation, structurally identical to dg_bin_mul_op, which computes the local derivative along a given direction and of a given order and projects it back to the original basis. See DR in issue #1065 .

The API is
```
/**
 * Compute df = d^order/dx_dir^order f, where f and df are DG fields
 * expanded in the given basis, using a local derivative (simply
 * differentiate the DG expansion in every cell).
 *
 * c_oop and c_iop select which component of a vector DG field to use.
 *
 * @param basis Basis functions used in the expansion.
 * @param dir   Direction in which to differentiate (0-based).
 * @param order Order of the derivative (1,2).
 * @param dx    Cell length in direction of the derivative.
 * @param c_oop Component of the output field to write into.
 * @param out   Output DG field.
 * @param c_iop Component of the input field to read from.
 * @param inp   Input DG field.
 */
void gkyl_dg_differentiate_op_local(const struct gkyl_basis *basis, int dir, int diff_order,
  double dx, int c_oop, struct gkyl_array *out, int c_iop, const struct gkyl_array *inp);
```

In the future we can add other functions/methods in this object, like wrapping hyper_dg and using multi-cell stencils (e.g. recovery). For the moment the main reason for this is to compute local derivatives in postgkyl.

Added a unit test, which passes on CPU and GPU. This PR depends on [gkylcas PR #101](https://github.com/ammarhakim/gkylcas/pull/101).


Antoinehoff:
I see that the unit test is in 1x only, is it a choice or you're planning to add more dimensionalities? it may be important to check what happens when you differentiate a 2D function along one variable only for example.

manauref:
I can add a 2x test easily

Antoinehoff (approved):
I approve because we want to move on and it is just a simple addition to the code base but I still think it's weird to have this implemented in Gkeyll while it is not meant to be used at all in a simulation. I think ideally we should have a real differentiation operator that could be used by postprocessing as well. Let's say this is a first step.

PR #1070, opened 2026-06-27
Improve native EFIT parser to handle arbitrary GEQDSK files

jjz73:
This PR seeks to accomplish two goals.

The first minor one is to fix several existing regression tests so that they initialize neutral species before populating species arrays, even if no neutrals are used. Prior to any changes, there was a memory access error that caused the tests to potentially fail to launch if the `num_neut_species` variable were not previously set to zero.

This issue above was brought to light when testing 6 regression tests against the changes made to accomplish the second major goal: improve geometry generation algorithms in the `gyrokinetic` module in order to handle arbitrary GEQDSK files fed into tokamak simulations.

The motivation came from a planned geometry scan of elongation and triangularity on the NSTX-U commissioning run. On the current version of `main`, out of the 16 desired shot-time pairs for the scan, only 11 equilibria yielded full grids, 2 yielded partial grids, and 3 failed to launch entirely due to unidentified X-points using only the quadratic basis for the search.

Now, on this branch `improve-xpt-finder`, all 16 grids generate completely, running on a modified algorithm (see `efit.c`, `efit_utils.c`, `calc_metric.c`, and `gkyl_efit_utils_priv.h`). Here, the code now uses the cubic X-point calculation (generally more consistently accurate than the quadratic counterpart) before applying the quadratic algorithm for the rest of the grid generation, in the name of speed. These grids still take on the order of 1-2 minutes locally on a MacBook CPU, no significantly slower than prior to these changes. Moreover, these changes successfully yield 16 generated grids on Perlmutter's GPUs as well.

dashboard_main_4x4

dashboard_ixf_4x4

The following 6 regression tests were used to confirm that significant changes to the metrics, field energy, and final potential remained the same across different devices (i.e., LTX, ASDEX, STEP, WHAM) and domains (i.e., half-domain NSTX-U, full-domain ASDEX):

```
gyrokinetic/creg/rt_gk_ltx_iwl_2x2v_p1.c
gyrokinetic/creg/rt_gk_bgk_asdex_2x2v_p1.c
gyrokinetic/creg/rt_gk_multib_asdex_2x2v_p1.c
gyrokinetic/creg/rt_gk_multib_step_2x2v_p1.c
gyrokinetic/creg/rt_gk_multib_step_nonuniform_2x2v_p1.c
gyrokinetic/creg/rt_gk_wham_2x2v_p1.c
```

Close quantitative agreement and consistent qualitative patterns were observed across all of these file outputs. The more significant quantitative differences were localized to downstream calculations of high-order moments (e.g., M2, M3, M3perp), which one might hypothesize is due to small deviations in geometry setup becoming exacerbated with each higher order of a calculated moment.

field_energy
phi_final
g_ij

For the files that produced a `runregression` failure due to a slight value mismatch or a "rel diff = 1", qualitative agreement was nevertheless present (see below for an example).

rt_gk_bgk_asdex_2x2v_p1-elc_M3perp_0


gyrokinetic/zero/efit_utils.c
```diff
@@ -1,7 +1,29 @@
 #include <gkyl_efit_priv.h>
 #include <float.h>
+#include <math.h>
 #include <string.h>
 
+static bool newton_raphson_seed(struct gkyl_efit *up, const double *coeffs,
+  const double *xinit, double *xsol, bool cubics);
+
+static inline double
+efit_sq(double x)
```
Antoinehoff, line 10: Isn't it possible to just use `pow(x,2)` everywhere instead of defining this special function?

jjz73: If `pow(x,2)` preserves double inputs as double outputs then yes. I mainly wanted to keep variable types as doubles everywhere because they then go into the `efit_dist2` function, and I wanted to keep the precision level constant.

ammarhakim (approved):
I think this is ready to merge.

gyrokinetic/zero/efit_utils.c
Antoinehoff: I think it does, `pow` has its `fpow` equivalent that handles `float` so I am pretty sure that `pow` is meant to work with `double`.

jjz73: Okay cool, I will make a new commit to simplify that then.

Antoinehoff:
I want to merge this soon but I guess it should wait for some testing on GPU when Perlmutter is back next week?

jjz73:
Yes, I understand. The high-resolution grids required additional fixes that I am currently still testing against STEP, ASDEX, and TCV, across DN, SN, uniform, and nonuniform grid cases. I will open a new PR once these are complete.

Issue #1069, opened 2026-06-23
[DR] Updater to evaluate a DG field at a coord and projects onto lower-dim basis


manauref:
# Summary

We propose to build an `core/` updater, `dg_eval_at_coord_proj`, that evaluates a DG field at a coordinate and projects onto the corresponding lower dimensional basis. This updater would take an donor DG field (its gkyl_array and basis) and arbitrary number of coordinates to evaluate at, and also takes a target DG field (its gkyl_array and basis) which is populated.

The idea is to support the following down projections:
- 1D down to:
  * 0D (scalar).
- 2D down to:
  * 1D along x.
  * 1D along y.
  * 0D (scalar).
- 3D down to:
  * 2D along y-z.
  * 2D along x-z.
  * 2D along x-y.
  * 1D along z.
  * 1D along y.
  * 1D along x.
  * 0D (scalar).

Note that most of this is already done by the gyrokinetic/zero/translate_dim updater, but that updater:
- Only evaluates at logical -1,0,1, so here we extend to allow passing an arbitrary grid coordinate.
- Also does up-projection from lower to higher dim, while this new updater only does down projection.
So we'll base `dg_eval_at_coord_proj` on the down projection part of `translate_dim`.

The current purpose of this updater is to call it from postgkyl to do true DG evaluation (presently postgkyl can only select the nearest interpolation/plotting point).

# API

The updater's `_new` method will look like
```
struct gkyl_dg_eval_at_coord_proj*
gkyl_dg_eval_at_coord_proj_new(const struct gkyl_rect_grid *grid_do, const struct gkyl_basis *basis_do,
  const struct gkyl_basis *basis_tar, int num_eval_dirs, const int *eval_dirs, bool use_gpu);
```

and the `_advance` method like
```
void gkyl_dg_eval_at_coord_proj_advance(struct gkyl_dg_eval_at_coord_proj *up, const double *eval_coords,
  const struct gkyl_range *rng_do, const struct gkyl_range *rng_tar,
  const struct gkyl_array *fdo, struct gkyl_array *ftar);
```

# Future extensions

Sometimes one also wants a DG field evaluated at a coordinate but projected onto the higher-dimensional basis (because it makes later arithmetic operations easier). In principle we could later allow for `basis_do` and `basis_tar` to be equal in order to support this.

We will prototype this in the `dg_evproj` branch.

manauref:
addressed in PR #1071

PR #1071, opened 2026-06-29
Updater to evaluate and project DG fields

manauref:
This goes with [PR 102 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/102).

1. Following the DR in issue #1069, add a new updater that evaluates a DG field at a coordinate and projects it onto the lower dimensional basis. There's a lot of kernels because of all the possible combinations of coordinates to evaluate at (and thus sub-space to project onto).

The main purpose of this updater is to give postgkyl a way to truly evaluate a field at an exact coordinate, and return a DG field (its coefficients) so that subsequent pgkyl commands work as expected (e.g. interp).

2. In the process of doing this we also added a new basis, GKHYBRID_VEL, which is the velocity space part of GKHYBRID. This is to make it easier to manipulate v-space only fields when working with gkhybrid; on this occasion, for the case in which all conf-space coords are evaluated.

3. Lastly, in order for pgkyl to create opaque gkyl_range, gkyl_basis and gkyl_rect_grid objects, we added new functions to allocate these to heap memory and initialize them. For the same reason, since postgkyl sees opaque objects, we added methods to retrieve the members of these structs.


gyrokinetic/zero/bc_twistshift.c
```diff
@@ -222,7 +222,8 @@ ts_check_shifted_test_point(struct gkyl_bc_twistshift *up, const double *test_pt
     false) }; // Shifted test point.
   int shift_dir_idx_test_pt[1];
 //  gkyl_rect_grid_coord_idx(&up->shift_grid, shifted_test_pt, shift_dir_idx_test_pt); 
-  gkyl_rect_grid_find_cell(&up->shift_grid, shifted_test_pt, pick_lower, (int[]) {-1}, shift_dir_idx_test_pt);
+  bool pick_lower_arr[] = {pick_lower};
```
Antoinehoff, line 225: What is this? can you explain a bit more why this changed?

manauref: if choose a coordinate that's at a cell boundary, this forces the find_cell method to select the lower cell of the two abutting this boundary. I meant to make this an option to the dg-evproj command but forgot.

Antoinehoff: it was more about why do we have to make it an array now? I guess you want to allow to pick lower in some directions but not others and this was not possible before?

manauref: yes, that's right, so we can pick the lower cell in some directions, but the upper in others (say if you evaluate the DG field at the corner between 4 cells in 2D). Previously we either picked the lower in all directions.

Antoinehoff (approved):

Antoinehoff (changes requested):
`gkyl_dg_eval_at_coord_proj_advance_cu` does not compile...

manauref:
> `gkyl_dg_eval_at_coord_proj_advance_cu` does not compile...

fixed

Antoinehoff (approved):

tnbernard (approved):

PR #1076, opened 2026-07-04
Floor cross species thermal velocities at zero

Maxwell-Rosen:
There is a fundamental flaw with the equations we use to compute the cross-species thermal velocities. These equations are defined in:
M. Francisquez, J. Juno, A. Hakim, G. W. Hammett, and D. R. Ernst, Improved multispecies Dougherty collisions, J. Plasma Phys. 88, 905880303 (2022).

The key equations are 2.9 and 2.10
image

When the mass ratio is large enough and Ti >> Te, this can lead to a negative thermal velocity squared.

To patch this, we insert the appropriate lines of code after the cross species thermal velocity is calculated

```C
// Floor the cell-average of vtSq_{sr} to zero, setting the rest of the coefficients to zero as well
gkyl_array_set_offset_range(lbo->cross_vtsq, 1.0, cross_prim_moms, 1*app->basis.num_basis, &app->local);
gkyl_array_max_by_cell_per_cell_avg_range(lbo->cross_vtsq, lbo->cross_vtsq_floor, &app->local);
gkyl_array_set_offset_range(cross_prim_moms, 1.0, lbo->cross_vtsq, 1*app->basis.num_basis, **&app->local);
```

This is an appropriate patch for robustness; however, a future endeavor will need to reformulate cross-species collisions so as not to produce this error. Rosenbluth collisions will be added in the future, so it's not worth redoing the Dougherty collision-operator formulation for the mirrors.

This PR requires a cross-species collisions diagnostic, so this is combined with #1013.

Before: 
image

After:
image


manauref:
i would think that flooring it to the minimum representable temperature (rather than 0) would be better.

Also, I think this needs some serious testing, perhaps with:
- a dedicated [collisional relaxation test](https://github.com/gkeyllorg/gkeyll/pull/905) (adapted to be in the appropriate regime)
- a production TCV IWL run (which are relatively cheap and @Antoinehoff has good baselines for).

Maxwell-Rosen:
Although I understand the physical reasons for setting the floor to the minimum representable temperature, I do not think this will make a significant numerical difference. I'm skeptical that this change would alter the results meaningfully, but we can do it.

manauref:
Closing for now. I think this may be a good idea, but needs to be explored a little further. To re-open I suggest:
1. Switch floor value from 0 to minimum representable temperature.
2. Do a couple of other tests, specifically: the LBO relaxation test (see PR #905 ) and a turbulent TCV test (@Antoinehoff can provide input file, preferably one that has some negative T even if transiently)

PR #1082, opened 2026-07-08
New C^0 position map for non-uniform position space simulations in Vlasov

JunoRavin:
Add the ability to specify a C^0 position map in the same fashion as the C^0 Serendipity velocity map (though this position map is utilized by all basis types). This work follows an identical structure to this recent PR https://github.com/ammarhakim/gkeyll/pull/1055 but now with a new gkyl_vlasov_position_map struct that holds the C^0 position map.


vlasov/apps/vm_species.c
```diff
@@ -654,6 +658,8 @@ vm_species_write_cell_avg_enabled(gkyl_vlasov_app* app, struct vm_species *vms,
   // the velocity-space Jacobian at specific quadrature points. 
   gkyl_vlasov_velocity_map_divide_jacobvel(vms->vel_map, &app->basis, &vms->basis,
     &vms->local, vms->f, vms->f_no_J);
+  gkyl_vlasov_position_map_divide_jacobpos(vms->pos_map, &vms->basis,
+    &vms->local, vms->f_no_J, vms->f_no_J);
```
johnson452, line 662: Do we potentially want this to be a divide Jacob total and it takes two Jacobians, Jc, and Jv?

PR #1087, opened 2026-07-19
Add loading a source from a .gkyl file

Maxwell-Rosen:
Adds `GKYL_SOURCE_FROMFILE`, a new `gkyl_source_id` that lets a gyrokinetic species (or kinetic neutral species) source term be loaded directly from a `.gkyl` file instead of being built from a projection, mirroring the existing `init_from_file` / `GKYL_IC_IMPORT_F` mechanism used for initial conditions.

## Verification

- A new regression test (`rt_gk_sheath_1x2v_p1_source_import.c`) runs a sheath sim once with `GKYL_PROJ_SOURCE` to write out `-elc_source` files, then a second run reloads those same files via `GKYL_SOURCE_FROMFILE`; the reloaded source arrays are confirmed bit-for-bit identical to the originals.
- Valgrind reports zero errors and zero leaks.

## Bug fix caught along the way

Valgrind caught a real crash in `gk_species_source_release` / `gk_neut_species_source_release`, which unconditionally released `proj_source[k]` even when it was never initialized for a from-file source; both release paths are now guarded on `source_id != GKYL_SOURCE_FROMFILE`.

## Implementation

Claude Sonnet 5 one-shot this using the prompt:
"Similar to how a distribution function can be loaded from a file
        .init_from_file = {
          .type = GKYL_IC_IMPORT_F,
          .file_name = "gk_wham_1x2v_p1-elc_0.gkyl",
          // .jacobtot_inv_file_name = "gk_wham_1x2v_p1-jacobtot_inv.gkyl",
        },,
  Enable sources to be loaded from files. This is specified inside as a GKYL_SOURCE_FROMFILE in the source_id of the source table"


manauref:
ready for approval. Please: a) resolve conflicts, b) run CI

Issue #1083, opened 2026-07-10
Spatial discrete low pass filter updater


Antoinehoff:
We propose to implement a low pass filter updater to gkeyll.

## Motivation
We identified aliasing rising from the twist-and-shift boundary condition and we want to be able to filter the waves that cannot be resolved on the grid. Note that this aliasing appears in the x-direction, which is not periodic.

## Method
Since we cannot expect periodicity in the direction, we propose here to implement a discrete sinc filter combined with a Blackman window. The discrete impulse response for a sample index $k \in [-M, M]$ (where $M$ is the half-width of the filter) is:
```math
    h_{\text{sinc}}[k] = \text{sinc}(2 f_c k) = \frac{\sin(2 \pi f_c k)}{2 \pi f_c k}
```

To prevent spectral leakage (the Gibbs phenomenon) caused by truncating the sinc function, we apply a Blackman window function $w[k]$ over the same domain $k \in [-M, M]$,
```math
    w[k] = 0.42 + 0.5 \cos\left(\frac{\pi k}{M}\right) + 0.08 \cos\left(\frac{2 \pi k}{M}\right),
```
where $0.42$, $0.5$, and $0.08$ are standard coefficients.
The practical filter coefficients are obtained by multiplying the ideal sinc response by the window function element-wise:
```math
    h[k] = h_{\text{sinc}}[k] \cdot w[k]
```

To guarantee that the filtering process preserves the overall mean amplitude of the wave, we normalize the discrete weights so they sum exactly to $1$,
```math
    \hat{h}[k] = \frac{h[k]}{\sum_{j=-M}^{M} h[j]}.
```

The filtered field writes
```math
    \hat{\phi}(x) = \sum_{k=-M}^{M} \hat{h}[k] \phi(x - k\Delta x),
```

## Implementation
We plan to create a new updater in the same fashion as `dg_interpolate` in `core/zero`. This will make it accessible to the entire code base since it is pretty generic. The updater will have the usual `_new`, `_advance`, and `_release` methods. Here is a sketch of it's interface:
```C
/**
 * @param dir Direction along which to filter.
 * @param half_width Stencil half-width M in cells (stencil spans 2M+1 cells).
 * @param cutoff_wavelength Cutoff wavelength (physical units) of the filter.
 * @param basis DG basis of the filtered field.
 * @param grid Grid the filtered field is defined on.
 * @param range Range to filter in (a sub-range of the range the fields are
[...]

PR #1088, opened 2026-07-27
Add a filter and upsampler to gk twist-shift bc

Antoinehoff:
This PR is related to DR #1083 and gkylcas PR https://github.com/gkeyllorg/gkylcas/pull/106

### Modal filter application
For simplicity we apply the filter on each DG modal coefficient separately. This holds **as long as the filter width is a multiple of the cell width**. Since the basis elements follow the property
```math
\psi_i^{(p)}(x)=\psi^{(p)}(x-x_j),
```
i.e. they are all the same except for a shift that center them in the middle of their respective cell $i$, the introduction of a shift[+] $k\Delta x$ yields[/+][-] which consist in a multiple of the cell width $k\Delta x$ yields[/-]
```math
\psi_i^{(p)}(x - k \Delta x) = \psi^{(p)}(x-x_i - k\Delta x) = \psi^{(p)}(x-x_{k+i}) = \psi_{i+k}^{(p)}(x).
```
Consequently we can write
```math
\phi(x - k\Delta x) = \sum_{i=1}^N \sum_{p=0}^{P} \phi_i^{(p)} \psi_{i+k}^{(p)}(x),
```
where we do not care about the $i+k>N$ limit for now (this can be resolved by a reflection at the boundary which preserves M0).
Now we see that the coefficient of order $q$ of the filtered DG rerpresentation in a cell $j$ is
```math
\hat{\phi}_{j}^{(q)} = \int dx \hat{\phi}(x) \psi_j^{(q)}(x) = \sum_{k=-M}^{M} \hat{h}[k] \sum_{i=1}^N \sum_{p=0}^{P} \phi_i^{(p)} \int dx \psi_{i+k}^{(p)}\psi_j^{(q)} = \sum_{k=-M}^{M} \hat{h}[k] \phi_{j-k}^{(q)},
```
where the integration limits are not necessary since the basis elements have a finite support.

#### Caveat
The filter is done on the DG modal coefficients directly and **cannot prevent oscillations smaller than the cell width.** This is not a problem for our main application, twist-shift anti-aliasing, because we filter a field that is previously upscaled to a resolution expected to be large enough so that the twist-shift does not introduce grid scale mode.

### Implementation and interface
We now separate the twist-shift machinery between the core DG updater that applies the shift, named `gkyl_twistshift_dg`, and the "orchestrator" of the full twist-shift boundary condition method, i.e. shift + filter, named `gkyl_bc_twistshift`. The source files are split accordingly, `bc_twistshift.c` vs. `twistshift_dg.c` and their `.cu` equivalent. The `gkyl_bc_twistshift` updater contains a `gkyl_twistshift_dg` updater, a filter, and an upscaler. This structure is efficient to define a twist-shift updater that is meant to be applied on a higher resolution. 

The parametrization of the filter applied in the twist-shift BC  is based on:
- `ts_filter_cutoff_wavelength`: the minimal wavelength where the filter should start filtering, in the ideal limit. Default value: $2 (x_{max}-x_{min})/N_x$, the mesh Nyquist frequency.
- `ts_filter_half_width`: the number of neigboring cells of the original resolution that the filter stencil parcours. 1 means that the stencil has a total length of 3 (1 + 2*1). Default value: 1.
- `ts_upsample_factor`: the refinement factor of the field that will be twist-shifted. This factor will affect the stencil size too. Default value: 4.
- `type`: gives the option to feactivate the filter with the enum `GKYL_CLOSED_FLUX_TSBC_NOFILTER`. Otherwise it is `GKYL_CLOSED_FLUX_TSBC` by default.
They are placed in the new `gkyl_gyrokinetic_geometry.closed_flux_bcs` sub-structure, where we regroup also the shift functions:
```C
  // Geometry
  .geometry = {
    .closed_flux_bcs = {
      .parallel_lower_bc_shift_func = bc_shift_func_lo,
      .parallel_upper_bc_shift_func = bc_shift_func_up,
      .parallel_lower_bc_shift_ctx = &ctx,
      .parallel_upper_bc_shift_ctx = &ctx,
      .type = GKYL_CLOSED_FLUX_TSBC,
      .ts_filter_cutoff_wavelength = 2.0*ctx.Lx/ctx.Nx,
      .ts_filter_half_width = 1,
      .ts_upsample_factor = 4,
    },
  },
```
Notes:
- `ts_upsample_factor` and `ts_filter_half_width` interact together. If the upscale resolution is 3, a filter half width of 2 will creates a filter stencil length of 13 (1 + 2*3*2) on the upscaled data.
- the filter stencil is based in terms of number of DG cells which means that it does not affect the field if it's set to filter the mesh Nyquist frequency, i.e. ~0.5 times the DG Nyquist frequency.

### Unit tests
The twistshift unit testing is now split between `ctest_twistshift_dg.c` which is equivalent to the former unique twistshift unit test. It checks that the shift operation on a given field is equal to hard coded data that were produced when the updater was developped.
A second and new test is now in `ctest_bc_twistshift.c`. It verifies that the total TSBC pipeline (upsample -> twistshift -> filter -> downsample) is functionning well by checking some basic features. Here are all the tests ran by this unit test in order:
- Identity between applying `twistshift_dg` updater and a filterless `bc_twistshift` updater.
- No alteration of a 3x field except the ghost values in the parallel direction.
- Identity between applying the `bc_twistshift` updater without a filter or with a mesh Nyquist frequency one as long as there is no aliasing rising from the shift operation.
- Like previous test but in the shearless case and mutliple upscaling factors.
- Identity between passing a DG array and passing an analytical shift function to the `bc_twistshift` updater, and that for multiple upscaling parameters.
- Variation of the cell average along x, $\sum_{i}^{N_x} |f_{i+1}^{(0)}-f_{i}^{(0)}|$, is reduced when we filter in a strong shear setup.
- Conservation of the integral, $\sum_{i}^{N_{ghost}} f_{i}^{(0)}$, for various upsample, half width and cutoff parameters and high shear.
- Zero shift is identity.
- Zero shear is cell translation by applying a shift that is a multiple of the $y$ grid, i.e. $S=n \Delta y$, and verifying that the cell $i$ content before the shift is equal to the cell $i+n$ content after the shift. This for different upsampling factors.

### Single mode example
We show here the example of a single two period $k_y$ mode as a "donor", the data that need to be twisted. Below, you see from left to right: i) the donor on the original resolution, ii) the shifted version of the donor on a 4x resolution (the truth), iii) the shifted version using the original Gkeyll TSBC, iv) the new Gkeyll TSBC that includes supersampling and filter.

image

In the original TSBC we see a reversed kx>0 wave appearing in the 2.4<x<3.0 region.  This is the aliasing we want to get rid of. A finite perturbation remains at the border of the domain. This is part of the limit of a discrete filter capability. At the boundary, the filter cannot be applied entirely as it is applied to neighbors cells. Here we reflect the data to apply the filter but this causes a reduction of the damping capability.

The Fourier spectral analysis shows that the new TSBC is indeed removing that branch and mimics better the spectrum of the truth. The higher modes, rising from the shift, that cannot be resolved are simply dissipated similarly to some local GK codes (GS2, GENE, stella, Gyacomo).

image

### Two mode example
We consider here a higher resolution and a two-mode input. The first mode is sheared outside of the resolution that can be resolved by the original grid. We see that the TSBC without anti-aliasing filter is presenting a checkerboard pattern while the anti-aliased version presents a much smoother wave, though boundary effects remain.

image
image

### GK passive advection regression test
We compare how the the shift alias or not by using `rt_gk_passive_3x2v_p1` setting `t_end=50` and 40
We see that the `main` version shifts without losses until the radial mode number decreases which is unphysical aliasing. 
gk_passive_elc_main

On the other hand, the filtered version has a similar start but dissipate the radial mode when it gets too high, removing totally aliasing.
gk_passive_elc_filt

We also verified that `main` result and the no filter result agree.

relative particle conservation for main (left) and filter (right)

Screenshot 2026-09-14 at 5 08 58 PM


manauref:
1. It'd be good to add integration limits to the math above for clarity.
2. If there are changes to the user-facing API, the DR or PR should describe those.
3. There are a few tests that would be good to do and publish, this comes to mind:
  A. I think some tests of the filter along (w/o TS) are in order. At the very least, apply it on a fully periodic domain and show what happens to the moments (and integrated moments) after the filter.
  B. Perform a filter+TS test on a periodic-x domain (so use a periodic shift) to create a baseline of what happens without boundary effects.
  C. Perform a filter+TS test for a shift that shouldn't produce any aliasing (on a given grid). I'm curious to see if it does anything at all, and how much it affects the non-aliasing modes.
  D. Run the 3x passive test form the GK paper, and plot the integrated moments like we did there (I think you did this already).
4. You reflect the data at the x-boundary. This seems reasonable (though it may come back to bite us if we try to parallelize along x). What if we don't apply the 
5. What exactly is `ctest_bc_twistshift.c` testing, and how where the accepted results generated?
6. I think you should change all the reg tests with TSBC to use this
7. I think these input file options
```
    .ts_filter_cutoff_wavelength = 2.0*ctx.Lx/ctx.Nx,
    .ts_filter_half_width = 1,
    .ts_upsample_factor = 2,
```
should have default values that are used if the user doesn't specify them (check if `ts_upsample_factor=0`).

Antoinehoff:
> 1. It'd be good to add integration limits to the math above for clarity.

Since the basis have a finite support I don't see this necessary. This is explicitely said in the PR test now.

> 2. If there are changes to the user-facing API, the DR or PR should describe those.

Done.

> 3. There are a few tests that would be good to do and publish, this comes to mind:
>    A. I think some tests of the filter along (w/o TS) are in order. At the very least, apply it on a fully periodic domain and show what happens to the moments (and integrated moments) after the filter.

The unit test `ctest_bc_twistshift.c` shows actually that we conserve the integral through the filter process. I will check if this holds true in a shearless simulation good call. From my understanding, the loss of particles comes from the resulting flux of the filtered field on one side and unfiltered field on the other one when shear is present. We may be able to test these.

>    B. Perform a filter+TS test on a periodic-x domain (so use a periodic shift) to create a baseline of what happens without boundary effects.

I'm not sure of this, you mean a regression test like a CBC with periodicity along x? I remember having a bad time setting x-periodicity in the EM work. The unit test shows that you conserve integral regardless if x is periodic or not.

>    C. Perform a filter+TS test for a shift that shouldn't produce any aliasing (on a given grid). I'm curious to see if it does anything at all, and how much it affects the non-aliasing modes.

The unit test `ctest_bc_twistshift.c` verifies this.

>    D. Run the 3x passive test form the GK paper, and plot the integrated moments like we did there (I think you did this already).

Done, in the description.

> 4. You reflect the data at the x-boundary. This seems reasonable (though it may come back to bite us if we try to parallelize along x). What if we don't apply the [reflection?]

We would loose the integral conservation. I don't think we want to create a new source of particle loss. I suggest to not treat this issue as long as it's not showing up.

> 5. What exactly is `ctest_bc_twistshift.c` testing, and how where the accepted results generated?

The PR description is updated.

> 6. I think you should change all the reg tests with TSBC to use this

So I should provide explicitely the default setup?

> 7. I think these input file options
> 
> ```
>     .ts_filter_cutoff_wavelength = 2.0*ctx.Lx/ctx.Nx,
>     .ts_filter_half_width = 1,
>     .ts_upsample_factor = 2,
> ```
> 
> should have default values that are used if the user doesn't specify them (check if `ts_upsample_factor=0`).

Done

PR #1089, opened 2026-07-28
Unit test fixes

Antoinehoff:
We fix a few failing unit tests on CPU and GPU. We also clean unused code in the GK app.

### `ctest_array_ops`
- `cu_array_copy_range_to_range_diff_range_dim` was failing on GPU because the deflated range is not providing the inverted indexer which is needed by the GPU code of this test.

### `ctest_dg_array_mask`
- The global_max variable (on host) is passed by reference in the reduction that uses the device array. We now do the reduction with the host array to get the global_max.

### `ctest_tensor_field_ops`
- The copy of a pointer address from host to device was not working because the data that was pointed was living on the host. We just fix that by copying the host pointer to device data with `->on_dev`.

### `ctest_dg_em_vars`
We import the fix present in kitp-hamil-vlasov branch.
- The components of the `gkyl_dg_mu_op_range` in the GPU part of the test were not set correctly (0 everywhere instead of 0,1, and 2 like in the CPU code).
- The tensor 2x and 3x p=2 tests were commented our for CPU but not for GPU.

### `ctest_fem_parproj`
- The `p=2` Dirichlet BC case is protected with an assert. We comment this case now for CPU and GPU.

### `ctest_dg_gyrokinetic_kern_tm`
- The value `0` was used as a collisionless enum but this corresponds to `GKYL_GK_COLLISIONLESS_NONE` which returns a `NULL` updater and fails the test. We now pass `GKYL_GK_COLLISIONLESS_ES`.

### `ctest_gyrokinetic_cross_prim_moms_bgk ` 
- The BGK cross-primitive-moments kernel was rewritten in commit bcc40b7c3 (Oct 14 2025, "Modify the BGK app to have the same structure as the LBO app…") and so the result changes slightly. We update the expected M2 moment value from 2.83391995e9 to 2.86262992e9.

### `ctest_mom_gyrokinetic`
- Only the M2 check line 887 fails typically `M2[0] computed = 172.5525 vs expected 172.3291`. This seems to be a consequence of the change of `bmag` representation. We update the reference values stored in  `m2Correct`  line 855.

### `ctest_rescale_ghost_jacf`
- This updater is not used anymore. We just remove it and the associated kernels at `gyrokinetic/ker/inflate_surf`.

### `ctest_gk_geometry_tok`
- Missing null position map for ASDEX tests (3 and 4). 
- For the elliptical and straight cylinder tests, some surface node had a small negative value in the argument of the square root in the dphidhteta (~-1e-7) so **we propose here to clamp the argument of the square root** though this could make some bug silent. 
- Finally the cylinder test was out of date in terms of the new geometry convention.
 
### `ctest_bc_twistshift`
- This will be fixed in the #1084 PR.

### `ctest_dg_interpolate` (Stellar `nvcc` fail)
- This test failed on stellar with `nvcc` because of a wrong memory access when computing the magnetic field amplitude.


manauref (approved):
approving, but please check if rescale_ghost_jacf has some associated gkylcas/maxima code, and delete that too

manauref:
we also have to be mindful that GK depends on some code in `vlasov/` (and maybe moments too), so that to get proper coverage we need the unit tests to be clean for `vlasov/` too.

Antoinehoff:
> approving, but please check if rescale_ghost_jacf has some associated gkylcas/maxima code, and delete that too

I see that the kernels used are in `gyrokinetic/ker/inflate_surf_kernels.h` and these seem to be unused in the rest of the code. I will remove them.

Antoinehoff:
> we also have to be mindful that GK depends on some code in `vlasov/` (and maybe moments too), so that to get proper coverage we need the unit tests to be clean for `vlasov/` too.

The version of the code proposed by this PR is passing all unit tests on both CPU and GPU except:
```
==================== gkeyll unit test summary ====================
Total: 126   Passed: 119   Failed: 7
Failed tests:
  moments: ctest_wv_vacuum_einstein_conformal
  moments: ctest_wv_vacuum_einstein
  moments: ctest_wv_gr_mhd_tetrad
  moments: ctest_wv_gr_mhd
  gyrokinetic: ctest_bc_twistshift
```

JunoRavin (approved):

PR #1091, opened 2026-07-31
Position map consistency check

Maxwell-Rosen:
Claude found an inconsistency in how the non-uniform position map is constructed for the Delta B minimizing map. The cap is fixed to pivot on the region's own uniform-coordinate bounds instead of raw theta, eliminating the geometry-breaking discontinuity and making maximum_slope_at_max_B safe to enable.

The slope limiters work by asking if the map is above or below the line that passes through the inflection point of the B field, the point where the non-uniform map changes from concave down to concave up, is above or below this line. The line has a slope set by the input file, so a linear map here is the maximum/minimum allowable deformation while preserving the map structure.

image

In this image, the y-axis is the new non-uniform position and the x-axis is the uniform axis. Here, the left side would have very sparse points, so we draw a red line with a known slope and instead use the blue region, where point density is lowest. Similarly, the grid gets very compact on the right, so we draw a line and take the blue linear part to keep the grid from getting too compact.

### DIFF
--- a/gyrokinetic/zero/gkyl_position_map_priv.h
+++ b/gyrokinetic/zero/gkyl_position_map_priv.h
@@ -673,18 +673,6 @@ position_map_constB_z_numeric(double t, const double *xn, double *fout, void *ct
   if (enable_limits_min_B || enable_limits_max_B)
   {
     // Set a minimum cell size on the edges.
-    // Theta_left/Theta_right are the physical (non-uniform) locations of the
-    // extrema bracketing this dB-accumulation region. theta lives in a
-    // different (uniform) coordinate, and the region's own span in that
-    // coordinate, [theta_bound_lower, theta_bound_upper], is generally NOT
-    // the same width as [Theta_left, Theta_right] -- regions accumulate
-    // equal dB, not equal theta or equal Theta. Anchoring the straight-line
-    // caps directly on raw theta (assuming theta ~= Theta_left/Theta_right)
-    // is only valid right at the region boundary; pivot on theta_bound_lower/
-    // upper instead so the cap is evaluated in the same coordinate as theta
-    // everywhere in the region (this matters most when a caller, e.g. the
-    // Gaussian smoothing quadrature, evaluates theta values whose own region
-    // is much narrower or wider in Theta-space than in theta-space).
     double Theta_left  = interval_lower;
     double Theta_right = interval_upper;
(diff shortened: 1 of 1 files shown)


Antoinehoff:
Hey Max, I'm not very familiar with that part of the code, could you illustrate more where it is used and why this fix matters? Also are Mana and I assignees or reviewers?

Maxwell-Rosen:
I'm using these slope limiters to make sure the non-uniform mapping doesn't become too steep. This is a concern near the mirror throat, midplane, and near the simulation boundaries, where dB/dz is near zero. I'm using these position maps in my mirror simulations.

gyrokinetic/zero/gkyl_position_map_priv.h
```diff
@@ -672,11 +672,25 @@ position_map_constB_z_numeric(double t, const double *xn, double *fout, void *ct
...
+    // different (uniform) coordinate, and the region's own span in that
+    // coordinate, [theta_bound_lower, theta_bound_upper], is generally NOT
+    // the same width as [Theta_left, Theta_right] -- regions accumulate
+    // equal dB, not equal theta or equal Theta. Anchoring the straight-line
+    // caps directly on raw theta (assuming theta ~= Theta_left/Theta_right)
+    // is only valid right at the region boundary; pivot on theta_bound_lower/
+    // upper instead so the cap is evaluated in the same coordinate as theta
+    // everywhere in the region (this matters most when a caller, e.g. the
+    // Gaussian smoothing quadrature, evaluates theta values whose own region
+    // is much narrower or wider in Theta-space than in theta-space).
```
Antoinehoff: Can you shorten this comment and make sure it answers to the "what" but not to the "how"? 
Since it seems you are changing the method here, could you provide a detailed explanation in the PR description on what this algorithm does now?

Maxwell-Rosen: I removed the bulk of this comment, shortening it to just "Set a minimum cell size on the edges."

Antoinehoff: Thanks, could you also provide more explanation about how this is done in the PR description for documentation purposes?

Maxwell-Rosen: Added. It's hard to describe this without an image, so I put a crude MS Paint sketch to explain what this feature is doing.

Antoinehoff (approved):

Maxwell-Rosen:
Continued in #1133

PR #1092, opened 2026-07-31
Remove Lapack dependency in core, as well as python shim wrapper

Maxwell-Rosen:
This PR is associated with https://github.com/gkeyllorg/postgkyl/pull/229. The Lapack-free build options were made possible by @ammarhakim, while the shim was my own work. This is such that a Python library can wrap Gkeyll without requiring any dependencies, thereby exposing fundamental array operations.


ammarhakim (approved):
This branch is ready to merge. It allows for a fully LAPACK-free build and it required for the use in pgkyl 2.0.

configure
```diff
@@ -267,6 +273,13 @@ else
     USE_LUA=
 fi
 
+if [ "$USE_LAPACK_LITE" = "yes" ]
+then
+    USE_LAPACK_LITE=1
+else
+    USE_MPI=
```
manauref: should this be "USE_LAPACK_LITE=" ?

manauref (changes requested):
see comment below, and please: A) resolve conflicts in any, B) run CI

One other thing, maybe this is in there already: should there be a check to make sure the builder (user) doesn't specify both lapack lite and regular lapack (to build in gkylsoft or to use from the machine/cluster)?

configure
Maxwell-Rosen: To my understanding, no

PR #1094, opened 2026-08-04
Fix GK scaled neutrals restart

manauref:
Previously when restarting it interpreted the frame we restart from as the t=0 state. Instead we need to re-project the initial conditions and record the true t=0 state. Also, need to use proj_host for this so that calling the projection doesn't interfere with the state read in from the restart frame.

### DIFF
--- a/gyrokinetic/apps/gk_neut_species_fluid.c
+++ b/gyrokinetic/apps/gk_neut_species_fluid.c
@@ -308,7 +308,9 @@ gk_neut_species_fluid_init(struct gkyl_gk *gk, struct gkyl_gyrokinetic_app *app,
   }
 
   // Initialize projection routine for initial conditions.
-  gk_neut_species_projection_init(app, ns, ns->info.projection, &ns->proj_init);
+  if (ns->info.init_from_file.type == 0) {
+    gk_neut_species_projection_init(app, ns, ns->info.projection, &ns->proj_init);
+  }
 
   // Allocate objects for computing diagnostic moments.
   int ndm = ns->info.num_diag_moments;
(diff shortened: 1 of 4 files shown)


JunoRavin:
Apologies, I might be going cross-eyed, but at first glance it looks like proj_host is only defined if you are running on GPUs, so are we certain this runs on CPUs? I know running with neutrals is mostly for production but if I'm reading this right, this would seg fault on CPUs (unless I'm just missing where proj_host is correctly acquired when running on CPUs).

manauref:
> Apologies, I might be going cross-eyed, but at first glance it looks like proj_host is only defined if you are running on GPUs, so are we certain this runs on CPUs? I know running with neutrals is mostly for production but if I'm reading this right, this would seg fault on CPUs (unless I'm just missing where proj_host is correctly acquired when running on CPUs).

It works on CPUs, I debugged there (and also tested on GPUs). proj_host is allocated here
https://github.com/gkeyllorg/gkeyll/blob/3a3102e5e12d36d9a2c3c95294b22f681e543e5e/gyrokinetic/apps/gk_neut_species_projection.c#L255

JunoRavin (approved):

JunoRavin:
Ah I understand, I was looking here:
https://github.com/gkeyllorg/gkeyll/blob/3a3102e5e12d36d9a2c3c95294b22f681e543e5e/gyrokinetic/apps/gk_neut_species_projection.c#L93

maybe we should remove this guard for use_gpu since we let proj_host always be allocated with the other ways of initializing? I don't know how often the PROJ_FUNC type is being used for initial conditions. 

Edit: or at line 247. Same thing.

manauref:
Perhaps. I think PROJ_FUNC still works, though I don't remember if I tried.

Antoinehoff (approved):

PR #1098, opened 2026-08-27
Add fdot_abs_integrated_moms diagnostic

Maxwell-Rosen:
Greg and I have been interested in the diagnostic of integral |fdot| dv dz in the mirror simulations. This is helpful for the L2 norm of the change in the distribution function. Here is an example of it being used. 

image

The diagnostic is added to the `time_rate_diagnostics` flag, so it will show both fdot and fdot_abs integrated diagnostics.

A new sum abs function was added, with a unit test, as this is necessary for |fdot| to be computed.

I also had GPT sol refactor the fdot diagnostics to take an enum table, specifying which of the fdot diagnostics the user wants. This also provides the fdot, RHS vector directly, which is a phase space array.


core/zero/array_reduce.c
```diff
@@ -119,3 +119,28 @@ gkyl_array_reduce_range(double *res,
...
+  long n = arr->ncomp;
+  for (long i=0; i<n; ++i) res[i] = 0;
+
+  struct gkyl_range_iter iter;
+  gkyl_range_iter_init(&iter, range);
+  while (gkyl_range_iter_next(&iter)) {
+    long start = gkyl_range_idx(range, iter.idx);
+    const double *d = gkyl_array_cfetch(arr, start);
+    for (long i=0; i<n; ++i)
+      res[i] += fabs(d[i]);
```
Antoinehoff: I am a bit confused about how $\sum |f_{dot}|$ is computed component wise like
```math
 v^{k} = \sum_{i=0}^{N} |f_i^{(k)}|
```
What's the meaning of this quantity? when $k\neq 0$ ?

Maxwell-Rosen: I believe this sums over all the components, which is also what the reduce operation for the regular sum does as well. In your notation, does i represent "cell" and k is the component? It sums over all cells and components here.

Antoinehoff: Yeah sorry my question was not clear and I think I misunderstood something. We agree that the `start` is used to take only the 0th coefficient to sum over the absolute value of the cell average ?

PR #1099, opened 2026-08-27
Improve comment in LBO app so it's more accurate.

manauref:


Maxwell-Rosen:
Why are we removing the detail that delta_ss = 2? Is it not =2?

Antoinehoff (approved):

manauref:
> Why are we removing the detail that delta_ss = 2? Is it not =2?

delta_ss doesn't appear anymore. Please approve

PR #1101, opened 2026-08-31
Refactor phi wall boundary conditions into species table

Maxwell-Rosen:
This is from sol 5.6 high. The parameter for the shetath potential function should be a property of the species, not the field, says Mana. This is because the sheath BC is a species BC, not a field BC.

This is behind my PR #997 

Sol's notes
  - Sheath BCs now provide wall potential through .aux_profile and .aux_ctx.
  - Sheath and fdot loss-cone paths consume species-owned projected profiles.
  - Grounded walls default to zero; time-dependent profiles update each stage.
  - Removed obsolete field-owned wall machinery and cleaned multiblock input handling.
  - Added focused coverage in gyrokinetic/unit/ctest_gk_species_wall_potential.c.


manauref (changes requested):
This approach is not quite as modular as I expected. Specifically, rather than a `gk_species_wall_potential.c` file that calculated phi_wall lower AND upper, there should just be two calls to a `gk_species_phi_wall_advance` method in gk_species_apply_bc, right before the gk_species_sheath_advance, one for each of lower/upper.

The call to calculate phi_wall in the `calc_field` will also go away, since phi_wall is a BC property.

Additionally:
- I don't see why lines 356-364 in `gyrokinetic_multib.c` should be deleted.
- L913 in `gyrokinetic_multib.c` will disappear after changes requested above.
- The introduction of `struct gk_species_wall_potential` seems unnecessary at first, but let's wait (until after changes requested above) and see. It might be a good idea.

Maxwell-Rosen:
This has been refactored to be contained inside a `gk_species_phi_wall_advance` method. The methods from `gk_field_biased_wall` have been moved into this new `gk_species_phi_wall` updater, as the biases are now a property of the species. Full functionality is retained.

Redundant code was removed from `gyrokinetic_multib.c`, but the lines in question were restored.

gyrokinetic/apps/gk_species_phi_wall.c
```diff
@@ -0,0 +1,50 @@
...
+      1, bc->aux_profile, bc->aux_ctx);
+    gk_species_phi_wall_advance(app, wall, 0.0);
+  }
+}
+
+void
+gk_species_phi_wall_advance(gkyl_gyrokinetic_app *app,
+  const struct gk_species_wall_potential *wall, double tm)
+{
+  if (wall->projector) {
```
manauref: as with other modules, since this _advance method is called in the time loop, remove this if-statement by creating a function pointer in the `struct gk_species_wall_potential` that gets allocated to an empty function if `aux_profile` is not given.

gyrokinetic/apps/gk_species.c
```diff
@@ -1497,6 +1501,10 @@ gk_species_init(struct gkyl_gk *gk_app_inp, struct gkyl_gyrokinetic_app *app, st
       gks->upper_bc[d].type = GKYL_BC_GK_SPECIES_PERIODIC;
     }
   }
+
+  int wall_dir = app->cdim-1;
+  gk_species_phi_wall_init(app, &gks->lower_bc[wall_dir], &gks->phi_wall_lo);
+  gk_species_phi_wall_init(app, &gks->upper_bc[wall_dir], &gks->phi_wall_up);
```
manauref: I think these should go around lines 880 and 940 I think

manauref (approved):
see 2 more requested edits below before you merge.

Antoinehoff:
Wait why was this merged with only one approval??

Maxwell-Rosen:
It was merged with only 1 approval because it was merged into a branch, not into main

Antoinehoff:
oh my bad hahaha thanks!

Issue #1093, opened 2026-08-04
Refactor unit test to unify the naming of CPU vs. GPU tests and improve test suite


Antoinehoff:
This issue is meant to open a discussion.

After working on PR #1090  and #1089, we realize that it would be nice to have a way to know if a unit test is a CPU or GPU test. For example it would allow to provide more information about what fails when running the test suite. Also, the naming convention is very blurry with some tests being named `_ho`/`_dev` or -/`_gpu` or -/`_cu`. 

Maybe it would be worth unifying first the naming convention of all unit tests. Once we've done that, we could consider a way for the Makefile unit test suite to detect if a test failed on CPU or GPU following the naming convention.

manauref:
I think it should be
_ho
_cu
because in the future we may have some other unit tests for AMD GPUs (so _dev is not specific enough).

Also, it'd be good to make test names more specific. Like instead of
test_1x1v
it'd be good for the name to say which operator it is testing, like
test_proj_maxwellian_1x1v

Antoinehoff:
I agree mostly but I am not sure what we will do when porting to AMD GPUs. We will have to differentiate between `use_gpu` and "execute CUDA code". I don't know if we want to have three type of unit tests or if we would rather run the unit test with `use_gpu` and then the test would simply execute on the current architecture.

JunoRavin:
In the AMD port I was experimenting with, I was more generally using _dev to mean that it was device code/memory (and most of the GKYL_HAVE_CUDA -> GKYL_HAVE_GPU) so I think not using _cu is okay (since in general we would want arrays to just be delineated by whether they are device or host memory).

Antoinehoff:
I agree this would be cleaner. The only thing is that unit tests are meant to cover some parts of the codebase and if we call a test `_dev` and let the architecture decide if the test goes through the CUDA or the AMD compatible one, then the name of the test will not indicate which part of the code we tested anymore. One thing I realize I'm not clear about: the AMD port is meant to have specific AMD compatible code right? Or is there a way to just make the AMD GPU go through the CUDA code?

JunoRavin:
Because AMD copied the Nvidia API almost identically, every AMD call is routed through a CUDA call with a lightweight shim translation layer to convert cu -> hip in a bunch of places (NCCL doesn't even require us to change things to RCCL, AMD straight up accepts NCCL calls as valid symbols). 

[...]

PR #1103, opened 2026-09-03
Clean up unit test names

manauref:
Following the discussion in Issue #1093 we:
- Rename all unit tests to end in _ho or _dev*.
- Add a unique portion of the file name in the test name so that when all tests are run we can tell where a test came from.

*note @JunoRavin that some moment tests had _ho and _lo for high- and low-order flux, respectively. To avoid conflict with _ho (for host) I renamed these to _hof and _lof.

Partly did this to test agent tracking with Entire.


Antoinehoff:
Great, I think we can now improve the output of the `make check` and `make unit-run` commands so that it reports on if the test failed on CPU or GPU. I can do that if you want.

PR #1104, opened 2026-09-03
Jenkins for CI

manauref:
We add the Jenkins file and readme with instructions on how to set up.

Note some terminology:
One may add
- Controllers: a controller is a place CI is launched from. For example, I added a personal computer controller which only runs CI for PRs authored by myself (others can do this too).
- Nodes: this is hardware CI runs on. The hardware where the controller resides is typically a node, but one may also add a remote node (e.g. if I had access to a workstation via ssh, I could add it as a node in my controller).

Although one can technically set up multiple nodes for a single controller, I've set it up (that is tested and written READMEs) for with a single node per controller. There are two types of nodes/controllers at the moment:

1. A team workstation: this will periodically query GitHub for any PR into main, and build/test it.
2. Personal/cluster machines: one has to manually initiate a build/test. This was to avoid over-clocking our personal computer, and to simplify handling SSH into clusters we don't own.

Once a build/test initiates it:

1. Builds the target branch (e.g. main), creates accepted results for reg tests.
3. Checks out the candidate branch.
4. Runs unit tests.
5. Runs regression tests and checks.
6. Reports to GitHub if unit and reg tests passed.

If a PR genuinely needs to change the accepted results, one needs to list that test in `ci/jenkins/expected_regression_diffs.txt‎`.

See ci/jenkins/README.md for more details.

Lastly, see below and you'll see this PR passes CI run on my laptop. You may not be able to click it (not sure, try), but when I do it opens my local Jenkins page in which I see this
Screenshot 2026-09-05 at 10 35 42 PM


Antoinehoff:
Hm there is a commit "Co-Authored-By" Claude. I really want to avoid that, they don't deserve to be mentioned.

manauref:
> Hm there is a commit "Co-Authored-By" Claude. I really want to avoid that, they don't deserve to be mentioned.

If we are all tracking stuff with something like Entire, sure. Otherwise I'd prefer to keep it.

PR #1107, opened 2026-09-04
Formatting for C/C++/CUDA files

manauref:
Here we use clang-format to:
1. Standardize the format of all C/C++/CUDA files (except for those in minus/ and ker/).
2. Add agent and CI tools so these formats are automatically preserved.

This is a different take on (but influenced by) PR #939 .

Standard format
-----------------

We adopt a Linux kernel style (even though this isn't the industry standard for C++/CUDA, but we like it). We made two modifications to the style via the `.clang-format` config file:
- Use 2-space indentation.
- Pack function arguments on a line (rather than each argument in a separate line).

**If you have modifications to the standard you'd like to see, please voice it in the comments below**. Note: we tweak a change here or there, maybe, but we should avoid over-tweaking because it's best if we stay as close as possible to an international standard (in this case the Linux kernel style). And keep in mind that no one (myself included) will be happy with every aspect of the style, but this will be the last time we have to talk about it.

Formatting CI
--------------

One may apply the `clang-format` formatter manually, see bottom of README. But we add two things so that we rarely have to do this:
- A `.pre-commit-config.yaml` file so that pre-commit automatically formats things when you do `git commit`, see README.
- We add a `.github/workflows/format-check.yml` file so that GitHub CI checks for the correct format.
- An agent SKILL file so agents format things correctly automatically.

One ask:
Git blame will see the commit that reformatted a given line, blaming me. If you want to see the actual code change, you'll have to git blame twice. I'm fine with this, as:
1. we rarely use git blame
2. if you suspect this PR broke something, git blame will point you to this PR/commit.
That said, if you prefer that I remove these commits/this PR from git blame, let me know, I think I can.


core/apps/lw_priv.c
```diff
@@ -6,56 +6,42 @@
...
-  { "bcZeroFlux", GKYL_SPECIES_ZERO_FLUX },
-  { "bcRecycle", GKYL_SPECIES_RECYCLE },
-  { 0, 0 }
-};
+static const struct gkyl_str_int_pair species_bcs[] = {{"bcCopy", GKYL_SPECIES_COPY},
+  {"bcWall", GKYL_SPECIES_REFLECT}, {"bcReflect", GKYL_SPECIES_REFLECT},
+  {"bcAbsorb", GKYL_SPECIES_ABSORB}, {"bcNoSlip", GKYL_SPECIES_NO_SLIP},
+  {"bcWedge", GKYL_SPECIES_WEDGE}, {"bcFunc", GKYL_SPECIES_FUNC},
+  {"bcFixedFunc", GKYL_SPECIES_FIXED_FUNC}, {"bcZeroFlux", GKYL_SPECIES_ZERO_FLUX},
+  {"bcRecycle", GKYL_SPECIES_RECYCLE}, {0, 0}};
```
Antoinehoff: do we really like this formatting?

core/creg/rt_kann_bench.c
```diff
@@ -170,8 +166,11 @@ main(int argc, char *argv[])
         out->vals[i][0] = ufunc(inp->vals[i][0]);
       }
       struct gkyl_kann_train_params params = {
-        .learning_rate = 1e-3f, .mini_size = 64,
-        .max_epoch = 5, .max_drop_streak = 5, .frac_val = 0.1f,
+        .learning_rate = 1e-3f,
+        .mini_size = 64,
+        .max_epoch = 5,
+        .max_drop_streak = 5,
+        .frac_val = 0.1f,
```
Antoinehoff: This does not seem to follow the style of other struct (see my next comment)

core/creg/rt_kann_cmp_arch.c
```diff
@@ -39,50 +36,41 @@ struct train_inp {
...
-    .N = Nt
-  };
+  struct xrange tr = {.xleft = 0.0f, .xright = 3.0f, .N = Nt};
 
-  struct xrange xr = {
-    .xleft = 0.0f,
-    .xright = 1.0f,
-    .N = Nx
-  };
+  struct xrange xr = {.xleft = 0.0f, .xright = 1.0f, .N = Nx};
```
Antoinehoff: this struct is not in lines while in my comment above it is. I don't understand why they do not follow the same style.

gyrokinetic/creg/rt_gk_tcv_iwl_adapt_source_3x2v_p1.c
```diff
@@ -621,186 +670,232 @@ main(int argc, char **argv)
...
+          .dir = 2,
+          .edge = GKYL_LOWER_EDGE,
+          .type = GKYL_BC_GK_SPECIES_SHEATH,
+        },
+        {
+          .dir = 2,
+          .edge = GKYL_UPPER_EDGE,
+          .type = GKYL_BC_GK_SPECIES_SHEATH,
+        },
+      },
```
Antoinehoff: I don't know about this too...

core/apps/lw_priv.c
manauref: there are many things I don't "like" about the formatting, but I've come to terms with the idea that we need to adopt something, and no one will be fully happy with it. Yet by doing this we don't have to discuss it again.

This case particular I agree with you, it's a bit too packed. I think there's a config file option that will help (Cpp11BracedListStyle) but dunno what else it'll change. Let me try

core/creg/rt_kann_bench.c
manauref: maybe the formatter doesn't pack lines (in a struct specification) already split, or split lines already packed.

I think the line limit also comes into play here. I think it's set to 80 (Linux kernel style). It asks: should we pack these arguments or not? if they fit in 80 characters, it allows packing, but if not, it splits them. But we could increase it a bit.

gyrokinetic/creg/rt_gk_tcv_iwl_adapt_source_3x2v_p1.c
manauref: i don't love it either

actually, it seems that when there's a trailing comma like `.type = GKYL_BC_GK_SPECIES_ABSORB, }`, it tells clang-format to break.

I think we have 2 options (tell me which you prefer):
1. We leave it as is in this PR. Ultimately you code it any way you want, but the formatter will change it automatically.
2. The challenge is dev A may place a comma, and dev B may not. So this inconsistency always arises and the format will look different. I think the solution here is to use `cling-tidy`, and have it remove trailing commas. This can happen automatically with pre-commit and CI (as clang-format are set up in this PR).

I'm thinking 2 is the way to go. Let me try it.

Maxwell-Rosen:
I love this. A formatting scheme is very valuable. We can nitpick the details, but that's not a valuable discussion. We should get it to a form that it's minimally abrasive and just apply it everywhere. Everyone has their own style of what coding should look like, but it's not a fruitful discussion. It's more important that the codebase is uniform rather than tuned for every individual's preferences

core/apps/lw_priv.c
manauref: well, so I spent a while trying to improve this. I wasn't able to without making other things worse.

Formatters have their limitations, but it's a price i'm happy to pay to not think about format ever again.

gyrokinetic/creg/rt_gk_tcv_iwl_adapt_source_3x2v_p1.c
manauref: 2 is actually not a viable option. One would need the latest clang-tidy which is not easy to get, and one would need cmake or a compiled version of Gkeyll and its dependencies (which shouldn't be necessary just to format).

I experimented with removing trailing commas with a python script. But that starts to get messy, and the result was not entirely satisfactory anyway (see [agent_tools-clang_format-remove_trailing_comma](https://github.com/gkeyllorg/gkeyll/tree/agent_tools-clang_format-remove_trailing_comma)). Although verbose, I think what's currently in this PR is best. For reading the code i'm not terribly bothered, and for writing: you write the code whichever way you want, the formatters will adapt it accordingly.

core/apps/lw_priv.c
manauref: i tried many things. Spent too long trying to make this better. There's no ideal fix without unreasonable complexity or making other things worse.

I'm opting for the latest Linux kernel style (.clang-format file from Linus Torvald's github), with 3 mods:
- 100 character line limit.
- No tab characters to indent.
- Indents are 2 blank space.

we'll all be equally unhappy, but won't have to deal with style this ever again

manauref:
Ok, after looking at some files and looking into @Antoinehoff's feedback, I came to the conclusion that, like @Maxwell-Rosen  said, nit-picking almost anything is not particularly productive. I tried to improve some of the things Antoine pointed out, and wasn't able to without "breaking"* other things. Formatters just have their limitations if you are at the same time trying to adopt some or all of an accepted standard (in this case Linux kernel style).

So i've opted to:
1. Remove trailing commas in struct assignments because some users put them, others don't, and this triggers different behavior in the formatter.*
2. Use the latest Linux kernel style (.clang-format file from Linus Torvald's github), with 3 mods:
- 100 character line limit.
- No tab characters to indent.
- Indents are 2 blank space.

we'll all be equally unhappy, but won't have to deal with style this ever again.

*I also tried adding trailing commas in struct initializers. This is much harder, but I did it. However the problem is that it breaks apart every struct initialization into multiple lines, which we don't want.

Antoinehoff:
I don't think it's a good idea to rewrite everything like this but if everyone wants to pursue well I'll live with it. I'm just not seeing the benefit vs. the risks of messing up with our code base. I don't see why this is needed. I don't care if the writing style is not perfectly strict everywhere and I like the method of just following the guidelines with more recent PR. Style variation also gives us an intuition about how old some parts of the codes are. But I guess standardization and dehumanization are trendy nowadays 🤖

Maxwell-Rosen (approved):
I went over this with Mana and I love it. Looks great to me.

manauref:
@Antoinehoff i checked with CI that formatting didn't break a single unit or reg test

PR #1117, opened 2026-09-09
Cleaning up regression and unit tests plus valgrind

Maxwell-Rosen:
This PR is a collection of the changes I made when working at Preference Model to clean up the regression testing and unit testing inside Gkeyll. Several changes focus on reducing the resolution of regression tests to make them more manageable. 

Valgrind-sanitized unit tests.
Added several AI-generated unit tests to improve our coverage
Removing all writes from unit tests
Ensuring every regression test outputs only 1 frame
Loosened some tolerances so unit tests would pass

Improvements to runregression:
Sorting by cost

PR #1119, opened 2026-09-09
Add/clean core and gyrokinetic unit tests

manauref:
This was originally in PR #1117


Maxwell-Rosen:
It's too bad that this didn't preserve my git history or that I authored these lines. I've reviewed all these changes and they're fine

PR #1124, opened 2026-09-11
Fix simlink agentic discovery

Maxwell-Rosen:
Codex was having issues with the simlinks and permissions. It wasn't able to read any of the skills we are using. I had codex fix the issues with the agentic workflow in the repo.

  - Shared policy in AGENTS.md, with corrected task-specific skill links.
  - Removed configuration symlinks; preserved local Claude permissions.
  - Added regular-file entry points for Claude, Gemini, and Copilot.
  - Standardized skill names and corrected build/test guidance.
  - Documented maintenance in .agents/README.md.


.agents/skills/compiling/SKILL.md
```diff
@@ -89,12 +89,12 @@ in `gyrokinetic/creg/`, it can be compiled with
 make -j"$NPROC" build/gyrokinetic/creg/my_sim
 ```
 
-If the input file is outside of the repository, one needs to use the shared makefile in
-`<path_to_gkylsoft>/gkylsoft/gkeyll/share/`. For example, if `my_other_sim.c` is an input file in `$HOME/new_sim/`,
-compile it with: 
+If the input file is outside of the repository, use the shared makefile in
+`<gkeyll_install_prefix>/share/`. Run it from the directory containing the input
```
manauref, line 93: change this to:

Run it from the directory containing the input
file:

```sh
cd <input_file_directory>

manauref (changes requested):
1. Why do you need to replace _ with - everywhere? My claude was loading skills just fine (claude tells me).
2. I think there's too much complexity or duplication in the agent files. Specifically:
2a) I don't think we need .agents/README.md. I think that is sometimes used as "the" agents.md file. So it'd be cleaner if we put everything (including what's currently in AGENTS.md) into .agents/README.md, and make AGENTS.md, CLAUDE.md and GEMINI.md links to it.
2b) My understanding is that nowadays Copilot defaults to looking for AGEntS.md if it doesn't find copilot-instructions.md, so I would delete it.
3. Gemini.md needs to be made a link.

Maxwell-Rosen:
> * Why do you need to replace _ with - everywhere? My claude was loading skills just fine (claude tells me).

It might be compatable with Claude, but this is folloiwng the specification in https://agentskills.io/specification. I think this is more universal

> * I think there's too much complexity or duplication in the agent files. Specifically:
>   2a) I don't think we need .agents/README.md. I think that is sometimes used as "the" agents.md file. So it'd be cleaner if we put everything (including what's currently in AGENTS.md) into .agents/README.md, and make AGENTS.md, CLAUDE.md and GEMINI.md links to it.

Agree. Deteleting. Refactoring and condensing.

>   2b) My understanding is that nowadays Copilot defaults to looking for AGEntS.md if it doesn't find copilot-instructions.md, so I would delete it.

Yes. Deleted

> * Gemini.md needs to be made a link.

It already is a link. Gemini does default to GEMINI.md, so this is neccisary, although there is a way to change it in settings https://geminicli.com/docs/cli/gemini-md/#modularize-context-with-imports

PR #1127, opened 2026-09-15
Revamp KITP work to expand capabilities for new Hamiltonians, complete hybrid tensor basis work, and streamline build system

JunoRavin:
In the original DRs for the KITP branch, we laid out a vision for different Hamiltonian types being all that was needed to obtain the spectrum of Vlasov systems we support (https://github.com/gkeyllorg/gkeyll/issues/776) 

In the first pass through, we only divided our Hamiltonian types between a `hamil_vel` and `hamil_gen`. The former mixes the different types of Hamiltonians we support, like the non-relativistic Hamiltonian $H = v^2/2$ and the relativistic Hamiltonian $H = \sqrt{m^2 + p^2}$; the latter is a non-descript name in the triads era where a generic Hamiltonian no longer implies geometry by default. 

This branch is an aggressive refactor of the Vlasov system to split the Hamiltonian into three types: `hamil_vel_sparse`, `hamil_vel_dense`, and `hamil_phase`, with the idea that the latter is for canonical Poisson bracket simulations and can reproduce @johnson452's [Journal of Computational Physics paper ](https://www.sciencedirect.com/science/article/pii/S0021999126004857). Note that the difference between `hamil_vel_sparse` and `hamil_vel_dense` is **only** for the Vlasov equation and does not say specifically what the Hamiltonian is. The form of the Hamiltonian is still determined by the Hamiltonian initialization object, so `hamil_vel_dense` can equally be utilized for relativistic Vlasov, $H = \sqrt{m^2 + p^2}$, or photons $H = |k|$, where $|k|$ is the momentum of the photon. 

In this way, triads now interfaces with all three Hamiltonian types, meaning gyrokinetic neutrals can utilize the sparse non-relativistic Hamiltonian, with the breakthroughs in the triad formalism. And if you want to do general geometry relativistic Vlasov? Well you will be able to do that too once we complete the app-side refactor sorting into our six options: three Hamiltonians + yes/no to triads. 

Since we were regenerating all kernels, and the pure p=1 Serendipity performance on the sparse velocity Hamiltonian with triads was quite poor in initial gyrokinetic neutrals testing (with significant errors in the energy), I pushed ahead with the completion of the new hybrid formalism where we utilize a hybrid tensor basis that is a tensor p=2 velocity-space representation that is more accurate for our dense velocity (and phase-space) Hamiltonians in Vlasov, along with allowing the more accurate, if more computationally expensive, velocity map representation that is a C^1 cubic with piecewise quadratic, continuous Jacobians. 

All of this required significant cleanup: renaming kernels, getting rid of deprecated canonical_pb kernels, and a large refactor of the surface integral updates driven by Fable to utilize look-up tables that then cut the size of the kernels by orders of magnitude (instead of repeated 16 digit coefficients that were driving the size of the kernels up by a factor of up to 100, just use the header file with the single declaration of the modal to nodal and nodal to modal coefficients that are needed for surface node evaluation and then surface node to surface mode evaluation). The new nodal flux kernels now compile much faster, and even admit a new nodal parallelization strategy that is quite a bit faster on GPUs (with my workstation GPU seeing factors of ~70 speed up for this component of the update while hyper_dg might only be a factor of 20-30, and since this piece of the update is 2-3x of the total cost of a simulation it means we get closer to a factor of 50-60 speed up on my workstation GPU now; pretty cool). I also split the kernels for Vlasov into three pieces: flux, mom, and vlasov, with the former two now containing all the surface update quantities and moment calculation quantities respectively. This split has two advantages: one, the kernels for Vlasov directory is now not 1000+ files (and thus the Github UI doesn't crap out at this directory), and two, more importantly, after the exploration from this old issue https://github.com/gkeyllorg/gkeyll/issues/310, we can now specify kernel-family compiler flags to optimize certain kinds of updaters. Some preliminary testing suggest nodal fluxes benefit from 128 registers per thread while moments benefit from 64 registers per thread. We now have the granularity to specify this. 

Because some of the new kernels are quite large, since not everyone needs certain Vlasov kernels for their applications, I have added three build flags with a sweep of the kernel selection in the private headers to not look for certain object files with some help from Fable; in this way, unless you are doing things like high dimensional general relativistic Vlasov simulations, or using the Gkeyll Vlasov solver for neutrals, you can reduce the overall Vlasov compilation time significantly.


core/unit/ctest_basis.c
```diff
@@ -297,13 +298,132 @@ test_hyb_members(struct gkyl_basis basis)
...
   TEST_CHECK( -fin[1] == fout[1] );
-  TEST_CHECK( fin[2] == fout[2] );  
+  TEST_CHECK( fin[2] == fout[2] );
   TEST_CHECK( -fin[3] == fout[3] );
   TEST_CHECK( fin[4] == fout[4] );
   TEST_CHECK( -fin[5] == fout[5] );
 
 }
 
+// Deterministic pseudo-random number in [-1,1] (LCG).
```
johnson452, line 308: Why not just use rand() from the standard C library?

core/unit/ctest_basis.c
```diff
@@ -297,13 +298,132 @@ test_hyb_members(struct gkyl_basis basis)
...
+      for (int j=0; j<nb; ++j)
+        gram[i*nb+j] += w*b[i]*b[j];
+  }
+  for (int i=0; i<nb; ++i)
+    for (int j=0; j<nb; ++j)
+      TEST_CHECK( gkyl_compare(i==j ? 1.0 : 0.0, gram[i*nb+j], 1e-12) );
+  gkyl_free(gram);
+
+  // Random modal vector.
+  unsigned seed = 12345u + 100u*cdim + vdim;
```
johnson452: Same here, could use just use a simple number?

core/unit/ctest_basis.c
```diff
@@ -286,7 +287,7 @@ test_hyb_members(struct gkyl_basis basis)
   TEST_CHECK( gkyl_compare(0.0, b[2], 1e-15) );
   TEST_CHECK( gkyl_compare(0.0, b[3], 1e-15) );
   TEST_CHECK( gkyl_compare(-sqrt(5.0)/4, b[4], 1e-15) );
-  TEST_CHECK( gkyl_compare(0.0, b[6], 1e-15) );
+  TEST_CHECK( gkyl_compare(0.0, b[5], 1e-15) );
```
johnson452, line 290: Was this a bug?

johnson452 (commented):
I've reviewed core unit/zero files, vlasov apps/creg and the configure, makefile, readme. It appears that the vlasov/kernel changes are to large and it stopped rendering files after 3000, so the vlasov zero changes don't appear here.

I've check the code runs and compiles. The ctests for vlasov and core work as intended, and that select regression tests for VM+triads work as intended and work with the new hybrid basis.

The code has also been checked for these select tests to be sanitizer clean. I have a few comments on the ctest_basis.c that I've left.

core/unit/ctest_basis.c
JunoRavin: rand() sequences differ between glibc and the BSD/macOS libc, so a failure on one machine would not reproduce with the same numbers on another. rand() also shares one global state across the whole acutest executable, so the values a test sees depend on which other tests ran before it. A local 3-line LCG avoids all of that. The test does not depend on the specific values, so srand(seed); rand() would work too. It just makes failures harder to reproduce across platforms.

core/unit/ctest_basis.c
JunoRavin: Yea this was a bug (the size of the basis in 1x1v is 6, so b[6] was garbage and depending on the machine this test would fail)

core/unit/ctest_basis.c
JunoRavin: Here yes we can use a simpler number. Fixing

PR #1132, opened 2026-09-16
Agent tools, formatting & Jenkins CI

manauref:
(this goes with [PR #108 of gkylcas](https://github.com/gkeyllorg/gkylcas/pull/108))

In order to help standardize and secure the way agents interact with Gkeyll, we add three things in this PR:

1. Agent (SKILL) files.
2. Formatter for C/C++/CUDA files.
3. A CI system based using Jenkins.

More details below.

# Agent (SKILL) files

We add a set of agent files to provide some guidelines for agents on how to interact (explore, modify, expand) Gkeyll. These files are:
- AGENTS.md: A brief summary of the agent's profile, pointing to the existing skill files.
- .agents/skills/directory-structure: Explanation of directories.
- .agents/skills/software-design: Tips on Gkeyll design.
- .agents/skills/naming-conventions: Naming patterns used for files, functions, variables, etc.
- .agents/skills/compiling: Instructions to compile code.
- .agents/skills/testing-and-verification: Instructions for running tests and CI.
- .agents/skills/gyrokinetic-details: Tips to consider when working on gyrokinetics.

There are some other files specific to Claude, Gemini or Codes, which are simply softlinks to the AGENTS.md file and the .agents/ folder.

# Formatter for C/C++/CUDA files

This was explained in PR #1107 , summarized here.

- We adopt a standard format, basically the Linux kernel style but with 2 spaces, packed function arguments, and curly braces everywhere.
- We add formatting as a GitHub CI step, so that incorrectly formatted code fails a PR. We also provide a `.pre-commit-config.yaml` file so that pre-commit can automatically format code before commiting.
- We add the formatting instructions in a SKILL file.
- Formatting skips the `ker/` and `minus/`, as those contain Maxima-generated and external library files, respectively.

We checked with our Jenkins CI that formatting didn't break unit/regression tests.

# A CI system based using Jenkins.

Introduced in PR #1104, this work exploits/incorporates:
- Cleaning of unit and regression tests that @Maxwell-Rosen @Antoinehoff @tnbernard  did.
- The runregression tool @JunoRavin implemented.

Jenkins is a launcher, and we built Jenkins scripts (jenkinsfile) that launch both unit and regression tests on a variety of machines. In order to balance safety and simplicity we opted for the following mode of operation:
- Team workstations (we have none yet): will periodically check for PRs into main, and will build/test those.
- All other machines (personal/stellar/perlmutter): one has to manually tigger the build/test, and it'll automatically report the result to the corresponding GitHub PR.

The setup requires using the Jenkins UI at least once, see ci/jenkins/README.md. After that, one may interact with Jenkins CI via the gkeyll-ci.sh script, for example

```
./ci/jenkins/gkeyll-ci.sh run --pr 123
```
will initiate the build/test for PR 123.
```
./ci/jenkins/gkeyll-ci.sh active
```
will report the presently running build/test, if any, (it'll give the build number, PR, etc), while
```
./ci/jenkins/gkeyll-ci.sh recent
```
will list the 10 most recent builds/tests. If the active build/test is 6, then
```
./ci/jenkins/gkeyll-ci.sh follow --build 6
``` 
will displace the Jenkins console (exit with Ctrl+c). Finally,
```
./ci/jenkins/gkeyll-ci.sh abort --build 6
``` 
will cancel the build/test.

Current limitations:
- It's only set up to test on single CPU (personal, stellar_cpu) or single GPU (perlmutter_gpu). We need to extend CI to do multi CPU/GPU.
- Most GK regression tests are run, with the exception of 3x2v multiblock (which we don't have true support for).
- A full run takes 3 hours on my laptop. It'd be nice to cut that down some but it's hard given that some GK sims take a long time to create geo. There are ways to speed this up, @Maxwell-Rosen for example managed to group together tests according to their runtime.
- We should add a check to make sure a PR doesn't blow up CI runtime.

As more people use it we'll find the correct way to expand/improve the CI system.

# Additional perks & caveats

1. Although there is a setup for Perlmutter, perlmutter queues are a pain and can delay CI for a long time. I'm hoping we can use StellarAI more routinely for this, and I got preliminary access to stellarAI just for trying this (I think it'll go public in a couple of weeks).
2.  The current PR will not show that it is passing the tests (I think) because the Jenkins files need to be in main for that.
* This is a security thing. It checks candidate PRs with the Jenkins configuration in main, so that a PR doesn't sidestep CI simply by changing Jennkinsfiles.
* I got around this (for testing) by creating many PRs of branches with Jenkinsfiles and testing against each other, as you may have seen over the last week.
* So you may not be able to actual test this without extra steps until we merge this in main.
3. CI found that we had been failing the 1x multiblock tests for a long time. At some point we broke this due to the way Jacobian rescaling was being handled at MB boundaries, so we fixed this here introducing weak division/inversion for p=0, fixing the 1x multiblock gk_collisionless_flux kernels, and adding an array in gk_field_1x that multiblock needs.

Future work
------------

Some things are left for future work:
1. GPU vs CPU comparisons: in a sense this checks if the implementation is correct, which we ask developers to confirm. At the moment (this PR) the intention of CI is to catch if a PR breaks the code.
2. Multi-GPU: there's preliminary infra to do this on Perlmutter GPU, but queues make it hard. When a team workstation or StellarAI comes online we can make this more routine.
3. Right now there are separate bash scripts for the 4 machines supported: personal, stellar_cpu, perlmutter_gpu and team. It'd be good to merge those into 1 or 2 in the future.


Maxwell-Rosen:
Can you please add a check that the execution time of regression tests on the branch being tested does not increase by a significant margin (10-20%). We should not be increasing the resolution or cost of regression tests on successive iterations. I would think this should throw a warning or error. Our regression testing seems to constantly regress as people add more production-capable simulations into it, so we want to prevent this.

Also, please diagnose why CI is failing. We should not merge a branch with CI fixes that has failing CI. I cannot view the jenkins run on your PR because it's local to your computer. I think that's a problem. Everyone should be able to view the jenkins logs for CI

Antoinehoff:
Great work Mana thank you! I'm excited to have more robust and automatized testing.

For the CI topic, I don't see here or in https://github.com/gkeyllorg/gkeyll/pull/1104 any mention of the new unit tests. Can you indicate what are these new unit tests for and what they do? Here or maybe #1104 is a better place. There are also new /core/ker/basis kernels, a modification of gk_field_1x.c, and changes in gk_collisionless_flux kernels.

Could you please also add a link to each sub-PRs that were merged into that one so that we can have a better granularity for the reviewing? Also it would be good ensure that each sub-PRs explain what is not explained in this PR.

Maxwell-Rosen:
Does CI test that all committed code is valgrind / compute-sanitizer clean? Perhaps we should only use the --moat tests and unit tests for valgrind sanitation. This is a very important part of keeping our code clean.

manauref:
> Can you please add a check that the execution time of regression tests on the branch being tested does not increase by a significant margin (10-20%). We should not be increasing the resolution or cost of regression tests on successive iterations. I would think this should throw a warning or error. Our regression testing seems to constantly regress as people add more production-capable simulations into it, so we want to prevent this.
> 
> Also, please diagnose why CI is failing. We should not merge a branch with CI fixes that has failing CI. I cannot view the jenkins run on your PR because it's local to your computer. I think that's a problem. Everyone should be able to view the jenkins logs for CI

I'll look into the runtime check.

CI is failing because the Jenkinsfiles need to be in main (and the other CI is failing because it's from an old setup on my laptop, i'll delete it now)

manauref:
> Great work Mana thank you! I'm excited to have more robust and automatized testing.
> 
> For the CI topic, I don't see here or in #1104 any mention of the new unit tests. Can you indicate what are these new unit tests for and what they do? Here or maybe #1104 is a better place. There are also new /core/ker/basis kernels, a modification of gk_field_1x.c, and changes in gk_collisionless_flux kernels.
> 
> Could you please also add a link to each sub-PRs that were merged into that one so that we can have a better granularity for the reviewing? Also it would be good ensure that each sub-PRs explain what is not explained in this PR.

I added a new section at the end of the PR description with some of this.
1. New kernels, gk_field_1x and changes in gk_collisionless_flux were to fix 1x multiblock sims (found out by CI).
2. New unit tests check cases that were not previously checked, @Maxwell-Rosen can expand (please).
3. Sub-PRs are linked in the description.

manauref:
> Does CI test that all committed code is valgrind / compute-sanitizer clean? Perhaps we should only use the --moat tests and unit tests for valgrind sanitation. This is a very important part of keeping our code clean.

I decided not to do this because we (I) don't have a robust place to valgrind. Stellar doesn't work, mac doesn't work, Perlmutter works but the queues are long. If a (Linux) team workstation materializes (;) @JunoRavin ) we can have that controller do valgrind.

Antoinehoff:
> > Does CI test that all committed code is valgrind / compute-sanitizer clean? Perhaps we should only use the --moat tests and unit tests for valgrind sanitation. This is a very important part of keeping our code clean.
> 
> I decided not to do this because we (I) don't have a robust place to valgrind. Stellar doesn't work, mac doesn't work, Perlmutter works but the queues are long. If a (Linux) team workstation materializes (;) @JunoRavin ) we can have that controller do valgrind.

we could use `leaks` on MacOS though I've heard it's not as reliable than valgrind or compute-sanitizer.

Antoinehoff:
For the formatting part:
I see that the formatter does not decide if we write functions comments like
```C
  // This is my comment and it is visible in the IDE.
void foo(double a)
{ 
solve_fusion();
}
```
or like
```C
void foo(double a)
{ 
  // This is my comment and it is NOT visible in the IDE.
  solve_fusion();
}
```
I don't think the auto formatter can impose this but could we agree on writing the first option?

Maxwell-Rosen:
It depends on the comment. Comments before a function definition are comments on that function in its entirety. A comment inside a function comments on what individual lines of code are doing. The second comment is about the intent of what `solve_fusion()` does while the first is the docstring of `foo`

manauref:
> For the formatting part: I see that the formatter does not decide if we write functions comments like
> 
> ```c
>   // This is my comment and it is visible in the IDE.
> void foo(double a)
> { 
> solve_fusion();
> }
> ```
> 
> or like
> 
> ```c
> void foo(double a)
> { 
>   // This is my comment and it is NOT visible in the IDE.
>   solve_fusion();
> }
> ```
> 
> I don't think the auto formatter can impose this but could we agree on writing the first option?

I personally prefer comments inside the function because if the function moves, the comments are guaranteed to move with it. I've seen cases where the function moved but the comments didn't. But I know Gkeyll is all over the place in this regard. If you and @Maxwell-Rosen find some accepted international standard out there, point me to it.

Whatever we decide we can put it in the agent files if clang-format can't do it.

Maxwell-Rosen:
I haven't actually used the Jenkins files, but from what I see, it seems like all the CI tests are grouped into one large report. Can you please break this up into separate reports for formatting, cpu unit tests, valgrind unit tests, GPU unit tests, compute sanitizer unit tests, CPU regression tests, GPU regression tests, valgrind regression, compute sanitizer regression. The current implementation makes it difficult to study this without getting local access to the host machine. There are several pull requests which will fail the regression tests, but are improving the physics, so failing the regression testing is fine (because our tests are high tolerance self-verifying against main). Failing unit tests, valgrind, and formatting should be blocking. Perhaps this can be done with one build, but separated into several GitHub reports.

Antoinehoff:
> It depends on the comment. Comments before a function definition are comments on that function in its entirety. A comment inside a function comments on what individual lines of code are doing. The second comment is about the intent of what `solve_fusion()` does while the first is the docstring of `foo`

Yes a comment should not be two lines away from it's commenting line, this is not my point. I'm talking, obviously, about the comment that says what `foo` does. I know that @manauref likes having the comment inside and I see his point BUT the standard seems to be to put the comment at the top because VScode, which I assume is following standards, is displaying the comment ONLY if it's above. Here is an example, I hope this is clear:
image
image

After a bit of research in the https://www.kernel.org/doc/html/v4.10/process/coding-style.html there is no example of a function general comment that would be placed inside the corpus of the function. 

To be clear, I don't want the formatter to do that, I don't think it can. But I want us to agree on this standard and adapt the code gradually, as @manauref did a lot of time for other coding style in other parts of the code.

Antoinehoff (approved):
great work thanks!

manauref:
> I haven't actually used the Jenkins files, but from what I see, it seems like all the CI tests are grouped into one large report. Can you please break this up into separate reports for formatting, cpu unit tests, valgrind unit tests, GPU unit tests, compute sanitizer unit tests, CPU regression tests, GPU regression tests, valgrind regression, compute sanitizer regression. The current implementation makes it difficult to study this without getting local access to the host machine. There are several pull requests which will fail the regression tests, but are improving the physics, so failing the regression testing is fine (because our tests are high tolerance self-verifying against main). Failing unit tests, valgrind, and formatting should be blocking. Perhaps this can be done with one build, but separated into several GitHub reports.

you can use "gkeyll jeci artifact" to see a list of available artifacts. You can download it artifacts with "gkeyll jeci artifact --fetch". Some of these artifacts have timings. Some are databases output by runregression, which you can query with "gkeyll queryrdb" to see what failed.

The commit status that is sent to Github has to be extremely brief, so we can't display much or any of this info on GH.

Maxwell-Rosen:
I'd be hapy to merge everything in this branch EXCEPT the Jenkins CI components. The changes to the code (formatting and agent files) are fine, and those are complete. The Jenkins work is far from complete. 

I'm not convinced Jenkins is the best platform for our team for several reasons. I don’t think individual users and contributors should have to set up their own laptops to run CI. We shouldn't merge this until we have a working workstation build. CI failing currently can be meaningless because we may be aware that the PR changes regression tests in order to correct the physics. The reasons for failure are hidden behind a local network, so most people just see that it failed and not a detailed explanation of what failed so that they can fix it.

manauref:
> I'd be hapy to merge everything in this branch EXCEPT the Jenkins CI components. The changes to the code (formatting and agent files) are fine, and those are complete. The Jenkins work is far from complete.
> 
> I'm not convinced Jenkins is the best platform for our team for several reasons. I don’t think individual users and contributors should have to set up their own laptops to run CI. We shouldn't merge this until we have a working workstation build. CI failing currently can be meaningless because we may be aware that the PR changes regression tests in order to correct the physics. The reasons for failure are hidden behind a local network, so most people just see that it failed and not a detailed explanation of what failed so that they can fix it.

@Maxwell-Rosen if you want to donate a $70,000 GPU workstation to the Gkeyll group, please do. Then we can set up "a working workstation build". Until then, we'll do what's implemented here and rely on developers to run Jenkins CI themselves.

Maxwell-Rosen:
A GPU workstation is overkill for this anyway. This job could be accomplished via a small mini pc (~$300) sitting at PPPL or Mana's house (but electricity is cheaper at PPPL). My point is that this is not a complete piece of work, and I don't want to see the rest of the changes in this PR stall until we figure out the CI situation. It is not a reasonable ask to have people run CI from their laptops when it takes ~3 hours, and we must use our laptops for daily activities or travel around with them. It's not reasonable to ask people to leave their laptops spinning overnight on CI either.

Making a seperate branch of this PR wouldn't be too bad. You'd remove those files from this PR, then branch from this branch, then make a commit in that branch to restore the files. This preserves the git history and takes <1 minute.

manauref:
We don't even have a team CPU machine! so we must adopt a distributed model.

And as for not using your personal computer:
- one has to run CI. We should've been doing it on our own computers all along, but we were not responsible/the system was too difficult. Now it's a lot easier to run everything and report to github (once you set it up).
- you don't have to use your personal computer, you can use stellar, or copy the stellar setup and set up another cluster

Maxwell-Rosen (approved):

PR #1135, opened 2026-09-19
Workstation CI

Maxwell-Rosen:
I set up the workstation Jenkins build with codex. It suggested a few changes. It had difficulty with the MPI dependencies, probably because I'm in an immutable operating system.


manauref:
> I set up the workstation Jenkins build with codex. It suggested a few changes. It had difficulty with the MPI dependencies, probably because I'm in an immutable operating system.

Can you explain what you changed and why? or share the Entire ID to your conversation so I can see

Maxwell-Rosen:
There is no Entire ID because this was done on a computer that does not develop Gkeyll and I didn't set it up. Basically, I pointed it to the README and told it, "Follow the README and set this computer up as the workstation node. Make as few changes to the codebase as necessary"

There was an issue with the build dependencies because this machine doesn't have Gkeyll installed. It was looking for MPI in $HOME, but there was no gkylsoft. The dependencies were not linked correctly to the temporary build. The dependencies seem to be built each time, so we should use those dependencies when checking CI.

